PR #10, opened 2020-07-25
ImagingExtractors and more

alejoe91:
Restructuring/cleanup of the repo.

- ImagingExtractors --> Tiff, Hdf5, Bioformats, Nwb

- Use static typing (from experience, better to start from the beginning and it makes code muuuuch nicer!)

PR #16, opened 2020-08-16
Various fixes and suggested changes

alejoe91:
Discussed here: https://github.com/catalystneuro/roiextractors/pull/10

+ decorators for input args for get_frames and get_videos, relative imports, and others

### DIFF
--- a/roiextractors/extraction_tools.py
+++ b/roiextractors/extraction_tools.py
@@ -1,11 +1,15 @@
 import numpy as np
 from typing import Union
 from pathlib import Path
+from functools import wraps
+from spikeextractors.extraction_tools import cast_start_end_frame
 
 ArrayType = Union[list, np.array]
 PathType = Union[str, Path]
 NumpyArray = Union[np.array, np.memmap]
 DtypeType = [str, np.dtype]
+
+MAX_FRAMES = 10000
 
 
 def _pixel_mask_extractor(_raw_images_trans, _roi_idx):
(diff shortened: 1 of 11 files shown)


bendichter:
@alejoe91 Thanks for working on this. Could you try to make smaller more modular PRs?

roiextractors/extraction_tools.py
```diff
@@ -39,3 +43,92 @@ def get_video_shape(video):
         num_channels, num_frames, size_x, size_y = video.shape
 
     return num_channels, num_frames, size_x, size_y
+
+
+def check_get_frames_args(func):
+    @wraps(func)
```
bendichter, line 49: Let's talk about this. I don't understand why it is necessary.

roiextractors/extraction_tools.py
```diff
@@ -39,3 +43,92 @@ def get_video_shape(video):
...
+
+        if len(frame_idxs) == 1:
+            return get_frames_correct_arg[0]
+        else:
+            return get_frames_correct_arg
+    return corrected_args
+
+
+def check_get_videos_args(func):
+    @wraps(func)
```
bendichter, line 80: Same here. I do not understand why this is necessary. You could just have

```python
def get_video(self, start_frame=None, end_frame=None, channel=0):
    return self._video[channel, start_frame: end_frame]
```
Those `start_frame` and `end_frame` default will be parsed correctly and have the same effect as the current code. The error messages you'd get would be pretty interpretable too.

bendichter: Why even have the `get_video` method? Wouldn't `get_frames(range(start_frame, end_frame))` have the same effect?

bendichter:
I don't see why we need 

`get_frame`, `get_frames` and `get_video`. Couldn't this all be easily handled by `get_frames`?

`get_frame(1)` === `get_frames([1])`
`get_video(start_frame, end_frame)` === `get_frames(range(start_frame, end_frame))`

Is there a reason to have all 3? Performance maybe? I don't think any of the backend I am familiar with would perform differently.

roiextractors/extraction_tools.py
alejoe91: Hi Ben,
I think this is very handy for the following (avoiding to repeat code)

1. Check that all frames are < get_num_frames()
2. Handle single frame (previous `get_frame`) and multiple frames.
3. Make `frame_idxs` and array always (otherwise all functions would need to parse it independently

If in the future we need more checks, I think that is the right way to do it. Not necessary, but very handy

roiextractors/extraction_tools.py
alejoe91: I think the `get_video` is more user friendly than the `get_frames(range(start_frame, end_frame))`. They are somewhat overlapping, but the use case is different:

- `get_frames` --> used to grab specific frames 
(e.g. random frames to get a mean image)
- `get_video` --> retrieve continuous snippets of video

Regarding the decorator, again, not necessary but useful! We can check that end_frame is not less than start_frame, that the number of frames is not too large, and so on.

alejoe91:
> I don't see why we need
> 
> `get_frame`, `get_frames` and `get_video`. Couldn't this all be easily handled by `get_frames`?
> 
> `get_frame(1)` === `get_frames([1])`
> `get_video(start_frame, end_frame)` === `get_frames(range(start_frame, end_frame))`
> 
> Is there a reason to have all 3? Performance maybe? I don't think any of the backend I am familiar with would perform differently.

I removed `get_frame` (I agree that is not needed), but I would leave `get_video` as IMO it is more user friendly for some use cases.

alejoe91:
Split in:
- https://github.com/catalystneuro/roiextractors/pull/24 
- https://github.com/catalystneuro/roiextractors/pull/25
- https://github.com/catalystneuro/roiextractors/pull/26

PR #18, opened 2020-08-17
added default fields to base class, add channels, traces methods

Saksham20:
`self._sampling_frequency`: already present from ImagingExtractors, but changed the default value from None to NaN since RoiResponseSeries would require this in the absence of timestamps. Keeping None, it complains. 
s`elf._channel_names` = ['Optical_channel'], this is the default name I have chosen for channels for which there are no names. We can debate about this. Its a list of len 1 reflecting the default number of assumed channels. 
`self._raw_movie_file_location:` required by method: get_movie_location() and some classes dont have this info but all need it. Makes sense to define it in the base then. 
`self.no_planes`: default is 1 for all the classes, would be different for the MultiSortingExtractor that we develop later. this is currently required by nwbwriter, so I though of defining this explicitly. 
`get_traces_dict() `return a dictionary with all the found ROIResponseSEries for the pipeline, for example suite2p has 3: Flourescence, Neuropil and deconvolved. Its different from `get_traces()` which currently returns a specific trace

### DIFF
--- a/roiextractors/segmentationextractor.py
+++ b/roiextractors/segmentationextractor.py
@@ -209,7 +209,7 @@ class SegmentationExtractor(ABC, BaseExtractor):
         pass
 
     @abstractmethod
-    def get_traces(self) -> ArrayType:
+    def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name=None) -> ArrayType:
         """
         Return RoiResponseSeries
         Returns
--- a/roiextractors/segmentationextractor.py
+++ b/roiextractors/segmentationextractor.py
@@ -16,7 +16,10 @@ class SegmentationExtractor(ABC, BaseExtractor):
 
     def __init__(self):
         BaseExtractor.__init__(self)
-        self._sampling_frequency = None
+        self._sampling_frequency = np.float('NaN')
+        self._channel_names = ['OpticalChannel']
+        self._raw_movie_file_location = ''
+        self.no_planes = 1
 
     @property
     def image_size(self):
@@ -64,7 +67,7 @@ class SegmentationExtractor(ABC, BaseExtractor):
         Returns
         -------
         roi_locs: np.array
-            Array with the first column representing the x (width) and second representing
+            Array with the first row representing the x (width) and second representing
             the y (height) coordinates of the ROI.
         """
         return self.get_roi_locations()
@@ -205,6 +208,28 @@ class SegmentationExtractor(ABC, BaseExtractor):
         """
         pass
 
+    @abstractmethod
+    def get_traces(self) -> ArrayType:
+        """
+        Return RoiResponseSeries
+        Returns
+        -------
+        traces: array_like
+            2-D array (ROI x timepoints)
+        """
+        pass
+
+    def get_traces_dict(self):
+        """
+        Returns traces as a dictionary with key as the name of the ROiResponseSeries
+        Returns
+        -------
+        _roi_response_dict: dict
+            dictionary with key, values representing different types of RoiResponseSeries
+            Flourescence, Neuropil, Deconvolved, Background etc
+        """
+        return self._roi_response_dict
+
     def get_sampling_frequency(self):
         """This function returns the sampling frequency in units of Hz.
 
@@ -225,6 +250,7 @@ class SegmentationExtractor(ABC, BaseExtractor):
         """
         return len(self.get_roi_ids())
 
+    @abstractmethod
     def get_images(self):
         """
         Retrieve any relevant greyscale images from the pipeline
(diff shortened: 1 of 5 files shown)


roiextractors/segmentationextractor.py
```diff
@@ -16,7 +16,10 @@ class SegmentationExtractor(ABC, BaseExtractor):
 
     def __init__(self):
         BaseExtractor.__init__(self)
-        self._sampling_frequency = None
+        self._sampling_frequency = np.float('NaN')
+        self._channel_names = ['OpticalChannel']
+        self._raw_movie_file_location = ''
+        self.no_planes = 1
```
bendichter, line 22: A `SegmentationExtractor` can only have 1 plane. Why specify it?

roiextractors/segmentationextractor.py
```diff
@@ -16,7 +16,10 @@ class SegmentationExtractor(ABC, BaseExtractor):
 
     def __init__(self):
         BaseExtractor.__init__(self)
-        self._sampling_frequency = None
+        self._sampling_frequency = np.float('NaN')
+        self._channel_names = ['OpticalChannel']
+        self._raw_movie_file_location = ''
```
bendichter, line 21: Why does `SegmentationExtractor` need a path to the raw movie file location? That should be handled by `ImagingExtractor`.

roiextractors/segmentationextractor.py
```diff
@@ -205,6 +208,28 @@ def get_image_size(self) -> ArrayType:
         """
         pass
 
+    @abstractmethod
+    def get_traces(self) -> ArrayType:
```
bendichter, line 212: how would a user access traces for Neuropil or Background? What about specific indices or channels?

roiextractors/segmentationextractor.py
```diff
@@ -238,7 +264,35 @@ def get_images(self):
...
+        """
+        Number of channels in the pipeline
+        Returns
+        -------
+        no_of_channels: int
+        """
+        return len(self._channel_names)
+
+    def get_movie_location(self):
+        """
```
bendichter, line 289: Why does `SegmentationExtractor` need a path to the raw movie file location? That should be handled by `ImagingExtractor`.

roiextractors/segmentationextractor.py
Saksham20: its only for now as nwbwriter looks for this, but once the Multisegmentation is implemented, then this would go

roiextractors/segmentationextractor.py
Saksham20: yes, correct, 
I think this would be redundant when we are able to link the functionality of imagingextractor and segmentationextractor object: meaning if we can run the segmentation pipeline (eg. SIMA, suite2p) from an imagingextractor object and store the output. Like in spikeinterface they have run methods of spikesorters classes that take in a recording object, run the pipeline and output a sorting object; in this case both the sorting and recording object pairs would point to the same raw file, no?

roiextractors/segmentationextractor.py
Saksham20: corrected here e3a438f6b82bf8aa40e91dbfee1f1e4f709f689e

Saksham20:
@bendichter  do you have any further feedback for this branch? 
Else, can you merge this, so that I can work on the branches for `MultiSegmentation` class and `write_segmentation()`

roiextractors/segmentationextractor.py
bendichter: Only the recording extractor should point to the raw data file

roiextractors/segmentationextractor.py
```diff
@@ -205,6 +208,28 @@ def get_image_size(self) -> ArrayType:
...
+        Return RoiResponseSeries
+        Returns
+        -------
+        traces: array_like
+            2-D array (ROI x timepoints)
+        """
+        pass
+
+    def get_traces_dict(self):
```
bendichter, line 222: are we still storing this as a dict?

roiextractors/segmentationextractor.py
```diff
@@ -225,6 +250,7 @@ def get_num_rois(self):
         """
         return len(self.get_roi_ids())
 
+    @abstractmethod
     def get_images(self):
```
bendichter, line 254: this should be `raise NotImplementedError`, because there may be `SegmentationExtractors` that do not have summary images.

bendichter:
@Saksham20 I don't think this branch is ready yet. There are some comments that you have not responded to

roiextractors/segmentationextractor.py
Saksham20: most of them do have mean images, and for those that dont have, it will return None. 
In this branch it will return: {'Images': 'meanImg': None}}, 
again if the implementation (`self._images_mean`, `self._images_correlation`) in the caiman branch is good, then I can make those changes to all classes in this branch.

PR #20, opened 2020-08-17
nwbwriter updates: multiple planes, check for ImagingExtractor written fields

### LINKED ISSUE #6: unused variables in NwbSegExt
These variables are unused

https://github.com/catalystneuro/segmentationextractors/blob/ccbdd3146be1565b16974c668f1deb591296dcbe/segmentationextractors/nwbextractor/nwbsegmentationextractor.py#L78-L82

Saksham20:
fix #6 
Initially `write_segmentation()` for the `NwbSegmentationextractor `was setup to expect SegExt classes to spit out lists of data corresponding to the planes. Now since these are all handling 1 planes, all data spit out is ndarray or int etc. 
The changes in this writer is reflecting this. 
Also some changes are corresponding to making roundtrip tests for all the `SegmentationExtractor `classes: sima, cnmfe, extract, nwb, suite2p.


roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,124 +429,123 @@ def get_property_data(self, property_name):
         return ret_val
 
     @staticmethod
-    def write_segmentation(segext_obj, savepath, metadata_dict, **kwargs):
+    def write_segmentation(segext_obj, savepath, metadata_dict=None, **kwargs):
```
bendichter, line 432: A few problems here:

1. You have called this variable "metadata_dict", but you allow it to be a string. Why not call it "metadata"? 
2. If you allow this to be multiple types, you should use a docstring to indicate this.
3. You allow this to be `None`, which means it is optional, but there are several places in this function that would break if `metadata_dict` is not provided.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,124 +429,123 @@ def get_property_data(self, property_name):
...
+        if isinstance(metadata_dict['Ophys']['Device'], list):
+            for devices in metadata_dict['Ophys']['Device']:
                 nwbfile.create_device(**devices)
         else:
-            nwbfile.create_device(**metadata_dict['ophys']['Device'])
+            nwbfile.create_device(**metadata_dict['Ophys']['Device'])
 
         # Processing Module:
-        ophys_mod = nwbfile.create_processing_module('ophys',
+        ophys_mod = nwbfile.create_processing_module('Ophys',
```
bendichter, line 456: "ophys" with a lower case "o" is correct. See https://www.nwb.org/best-practices/. I know there are other files and conversion scripts out there that use a capital "O." Those are mistakes that need to be corrected.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,124 +429,123 @@ def get_property_data(self, property_name):
...
 
-        # OpticalChannel:
-        channel_names = segext_obj.get_channel_names()
-        input_args = [dict(name=i) for i in channel_names]
-        for j, i in enumerate(metadata_dict['ophys']['ImagingPlane']['optical_channel']):
-            input_args[j].update(**i)
-        optical_channels = [OpticalChannel(input_args[j]) for j, i in enumerate(channel_names)]
+        #OPtical Channel:
+        channel_names = [segext_obj.get_channel_names()]
+        input_args=[[dict(name=i) for i in channel_names[k]] for k in range(segext_obj.no_planes)]
```
bendichter, line 465: Shouldn't there me a check if this is a MultiSegmentationExtractor?

bendichter:
I don't see the changes to the round trip tests. I think those should be in a separate PR anyway

Saksham20:
> I don't see the changes to the round trip tests. I think those should be in a separate PR anyway

yes , i intend to create the changes to test files in a separate PR

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: I was thinking we make the metadata as an attribute of the class itself. THe user can fill in a yaml file and store the location in the variable: self.metadata = 'path-to-yaml' OR dict 
That way we also wouldn't need this as a required arg. 
What do you think?

bendichter: I'd like to keep this method static so that it maintains the same syntax as all of the other writers across SpikeInterface and RoiExtractors.

Saksham20:
@bendichter  do you have any further feedback for this branch? 
Else, can you merge this, so that I can work on the branches for `MultiSegmentation` class and `write_segmentation()`

I will make the changes in nwbwriter to check for `MultiSegmentation `class in that branch

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: it is static

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: yes, its fixed here 2ab62d4fa4dfaf0f7165f3b40358883f666485a6
The multisegmentation is implemented here #36

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,133 +416,143 @@ def get_property_data(self, property_name):
...
 
     @staticmethod
-    def write_segmentation(segext_obj, savepath, metadata_dict, **kwargs):
-        if isinstance(metadata_dict, str):
-            with open(metadata_dict, 'r') as f:
-                metadata_dict = yaml.safe_load(f)
-
+    def write_segmentation(segext_obj, savepath, **kwargs):
+        if not kwargs.get('metadata'):
+            raise Exception('provide metadata as argument')
```
bendichter, line 421: required args should be positional

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,133 +416,143 @@ def get_property_data(self, property_name):
...
+        metadata = kwargs.get('metadata')
+        if segext_obj.extractor_name == 'MultiSegmentationExtractor':
```
bendichter, line 423: use `isinstance`

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -429,133 +416,143 @@ def get_property_data(self, property_name):
...
+            segext_objs = segext_obj.segmentations
+            metadata_list = metadata
+        else:
+            metadata_list = [metadata]
+            segext_objs = [segext_obj]
+        source_path = segext_objs[0].filepath
+        print(f'writing nwb for {segext_objs[0].extractor_name}\n')
+        if isinstance(metadata_list[0], str):
```
bendichter, line 431: the list should be in ImagingPlane, and that list will contain information about individual planes. The session-level information should only be defined once.

Saksham20:
commit https://github.com/catalystneuro/roiextractors/pull/20/commits/1796c0a68e4ffbbce56c9ea3f25380a00578ed05
checks for pre-existing nwbfile while would have been written by an ImagingExtractor. It seems like a large diff since it incorporates changes from downstream testing branch which are not currently merged into master. 
One option is for me to undo the commit, merge this PR and then create another PR (after some other PRs are also merged into master) , else we can merge this itself into master after review

Saksham20 (commented):
@bendichter @alejoe91 
This is a fairly large change from the original; can be have a short chat regarding this? OR should this PR be moved from master to nwb_imaging branch? 
However, 
Ignore changes in `NwbImagingExtractor` since these are picked from branch: nwb_imaging. 
Ignore changes to `NwbSegmentationExtractor` class as they have been picked from master. 

Changes to note:

1.  `NwbSegmentationExtractor.write_segmentation()`,
2. `NwbSegmentationExtractor.get_nwb_metadata()`
3.  func `nwb_metadata_recursive_update()` 

Summary of changes: 

1. `NwbSegmentationExtractor.write_segmentation()`, this checks for existing nwb file and existing datatypes before writing (checks are useful when the nwbfile is an output from the `NwbImagingExtractor.write_imaging()`)
2. `NwbSegmentationExtractor.get_nwb_metadata()`: this is what we discussed: this method is moved from the base class to here, and does not depend on a yaml file, instead generates the dict in the code. 
3. `nwb_metadata_recursive_update()` : updates the metadata with the user input metadata.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -229,54 +492,46 @@ def __init__(self, filepath, optical_channel_name=None,
...
 
         # Extract samp_freq:
-        self._samp_freq = self.roi_response.rate
-        # Extract no_rois/ids:
+        self._sampling_frequency = mod['Fluorescence'].get_roi_response_series(roi_names[0]).rate
+        # Extract get_num_rois()/ids:
         self._roi_idx = np.array(ps.id.data)
 
         # Imaging plane:
         _optical_channel_exist = [i for i, e in enumerate(
```
alejoe91, line 520: Optical channels will be in `ophys/OpticalChannel`.

Why not just looping through those:
```
if `OpticalChannel` in nwbfile['ophys']:
    for oc in nwbfile['ophys']['OpticalChannel]:
        ...
```
 I think it's nicer to read!

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -164,50 +254,223 @@ def get_num_channels(self):
...
         self.image_masks = None
-        self.pixel_masks = None
         self._roi_locs = None
         self._accepted_list = None
-        io = NWBHDF5IO(filepath, mode='r+')
-        nwbfile = io.read()
+        self._io = NWBHDF5IO(file_path, mode='r+')
+        nwbfile = self._io.read()
         self.nwbfile = nwbfile
         _nwbchildren_type = [type(i).__name__ for i in nwbfile.all_children()]
```
alejoe91, line 475: Since we know the structure of the NWB file, I'd rather access the different fields that we know should be there. E.g. `nwbfile['ophys']['OpticalChannel']` and so on. It would improve readibility

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +542,229 @@ def __init__(self, filepath, optical_channel_name=None,
...
+    def get_nwb_metadata(sgmextractor):
+        """
+        Converts metadata from the segmentation into nwb specific metadata
+        Parameters
+        ----------
+        sgmextractor: SegmentationExtractor
+        """
+        metadata = {'NWBFile': {'session_start_time': datetime.now(),
+                                'identifier': str(uuid.uuid4()),
+                                'session_description': 'ses desc'},
```
alejoe91, line 584:
```suggestion
                                'session_description': 'no description'},
```

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +542,229 @@ def __init__(self, filepath, optical_channel_name=None,
...
-        # Subject:
-        if 'Subject' in metadata_dict:
-            nwbfile.subject = Subject(**metadata_dict['Subject'])
-
-        # Device:
-        if isinstance(metadata_dict['ophys']['Device'], list):
-            for devices in metadata_dict['ophys']['Device']:
-                nwbfile.create_device(**devices)
+    def write_segmentation(segext_obj, save_path, plane_num=0, metadata=None, file_overwrite=False):
+        if os.path.exists(save_path) and not file_overwrite:
```
alejoe91, line 630:
```suggestion
        save_path = Path(save_path)
        if save_path.is_file() and not file_overwrite:
            assert save_path.suffix == '.nwb', "'save_path' should be an .nwb file"
```

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +542,229 @@ def __init__(self, filepath, optical_channel_name=None,
...
-
-        # ROI add:
-        pixel_mask_exist = segext_obj.get_pixel_masks() is not None
-        for i, roiid in enumerate(segext_obj.roi_idx):
-            if pixel_mask_exist:
-                [ps_loop.add_roi(
-                    id=roiid,
-                    pixel_mask=segext_obj.get_pixel_masks(roi_ids=[roiid])[:, 0:-1])
-                 for ps_loop in ps]
+            if os.path.exists(save_path):
```
alejoe91, line 634: same here

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +542,229 @@ def __init__(self, filepath, optical_channel_name=None,
...
+                        if img_no is not None:
+                            images.add_image(GrayscaleImage(name=img_name, data=img_no))
+                    ophys.add(images)
+
+            # saving NWB file:
             io.write(nwbfile)
 
         # test read
-        with NWBHDF5IO(savepath, 'r') as io:
+        with NWBHDF5IO(save_path, 'r') as io:
```
alejoe91, line 769: I think this is not necessary. We'll have specific tests in to do it.

alejoe91:
@Saksham20 thanks for the changes and for merging the changes from the nwb_imaging branch. I just had a few comments on the use of `_nwbchildren_name` and `_nwbchildren_type`. 

Since we know the structure of the NWB file, I think it would be nicer and more readable to read and check the relevant fields. The use of `_nwbchildren_name` removes the hierarchy of the metadata.

Also I think we should start using pathlib everywhere 👍 

The rest looks good! I would merge this directly into master once finished.

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: thanks for this suggestion, I have made another implementation and created a small PR altogether in #46

roiextractors/extractors/nwbextractors/nwbextractors.py
bendichter: actually, I'd like to keep this in for now. I've had problems previously where a file is written in pynwb but cannot be read. I think we have fixed all the places where this might occur now, but I'd like to catch this as early as possible if it occurs again.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -19,6 +27,25 @@
     HAVE_NWB = False
 
 
+def nwb_metadata_recursive_update(metadata_base, metadata_input):
```
bendichter, line 30: ok but I would argue this function does not really do anything nwb-specific, so should not be called "nwb_"

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -19,6 +27,25 @@
...
+    for base_key, base_val in return_dict.items():
+        if metadata_input.get(base_key):
+            if isinstance(base_val,dict):
+                return_dict[base_key] = nwb_metadata_recursive_update(base_val, metadata_input[base_key])
+            elif isinstance(base_val,list):
+                if isinstance(metadata_input[base_key], list):
+                    list_len = min(len(base_val),len(metadata_input[base_key]))
+                    for base_val_num in range(list_len):
+                        return_dict[base_key][base_val_num] = \
+                            nwb_metadata_recursive_update(return_dict[base_key][base_val_num], metadata_input[base_key][base_val_num])
```
bendichter, line 41: I think there's a cleaner way do to this. try zipping the lists together

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -80,55 +116,112 @@ class NwbImagingExtractor(ImagingExtractor):
...
+            raise ValueError('No acquisitions found in the .nwb file.')
+        self._optical_series_name = a_names[0]
+
+        opts = self.nwbfile.acquisition[self._optical_series_name]
+        assert isinstance(opts, TwoPhotonSeries), "The optical series must be of type pynwb.TwoPhotonSeries"
+
+        #TODO if external file --> return another proper extractor (e.g. TiffImagingExtractor)
+        assert opts.external_file is None, "Only 'raw' format is currently supported"
+
+        if hasattr(opts, 'timestamps') and opts.timestamps:
```
bendichter, line 150:
```suggestion
        if hasattr(opts, 'timestamps') and opts.timestamps is not None:
```

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -80,55 +116,112 @@ class NwbImagingExtractor(ImagingExtractor):
...
+            df_epochs = self.nwbfile.epochs.to_dataframe()
+            # TODO implement add_epoch() method in base class
+            self._epochs = {row['tags'][0]: {
+                'start_frame': self.time_to_frame(row['start_time']),
+                'end_frame': self.time_to_frame(row['stop_time'])}
+                for _, row in df_epochs.iterrows()}
+
+            self._kwargs = {'file_path': str(Path(file_path).absolute()),
+                            'optical_series_name': optical_series_name}
+        self.make_nwb_metadata(nwbfile=self.nwbfile, opts=opts)
```
bendichter, line 177: should not be in init

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -80,55 +116,112 @@ class NwbImagingExtractor(ImagingExtractor):
...
+        # Update metadata with Device info
+        self.nwb_metadata['Ophys']['Device'] = []
+        for dev in nwbfile.devices:
+            self.nwb_metadata['Ophys']['Device'].append({'name': dev})
+
+        # Update metadata with ElectricalSeries info
+        self.nwb_metadata['Ophys']['TwoPhotonSeries'] = []
+        self.nwb_metadata['Ophys']['TwoPhotonSeries'].append({
+            'name': opts.name
+        })
```
bendichter, line 208: This would be cleaner as a literal with list comprehensions

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -80,55 +116,112 @@ class NwbImagingExtractor(ImagingExtractor):
...
+    #TODO use lazy_ops
+    @check_get_frames_args
+    def get_frames(self, frame_idxs, channel=0):
+        opts = self.nwbfile.acquisition[self._optical_series_name]
+        if frame_idxs.size > 1 and np.all(np.diff(frame_idxs) > 0):
+            return opts.data[frame_idxs]
+        else:
+            sorted_idxs = np.sort(frame_idxs)
+            argsorted_idxs = np.argsort(frame_idxs)
+            return opts.data[sorted_idxs][argsorted_idxs]
```
bendichter, line 219: channel does not do anything here

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -80,55 +116,112 @@ class NwbImagingExtractor(ImagingExtractor):
...
+
+    @check_get_videos_args
+    def get_video(self, start_frame=None, end_frame=None, channel=0):
+        opts = self.nwbfile.acquisition[self._optical_series_name]
+        video = opts.data[start_frame:end_frame]
         return video
```
bendichter, line 225: channel does not do anything here. Also, use DatasetView

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -164,50 +254,223 @@ def get_num_channels(self):
...
+            # retrieve device
+            device = nwbfile.devices[list(nwbfile.devices.keys())[0]]
+
+            # create optical channel
+            if 'OpticalChannel' not in metadata['Ophys']:
+                metadata['Ophys']['OpticalChannel'] = [{'name': 'OpticalChannel',
+                                                        'description': 'no description',
+                                                        'emission_lambda': 500.}]
+
+            optical_channel = OpticalChannel(**metadata['Ophys']['OpticalChannel'][0])
```
bendichter, line 295:
```suggestion
            optical_channel_args = dict(name='OpticalChannel', description='no description', emission_lambda=np.nan)
            
            if 'OpticalChannel' in metadata['Ophys']:
                optical_channel_args.update(metadata['Ophys']['OpticalChannel'][0])

            optical_channel = OpticalChannel(**optical_channel_args)
```

If we make up placeholder data, it needs to be clear that we are doing that. In this case that means using `np.nan`. Using `500.` will inevitably lead to us converting data incorrectly!

bendichter: also this approach allows for the args to OpticalChannel to be partially filled in

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -164,50 +254,223 @@ def get_num_channels(self):
...
+                                                      'indicator': 'Indicator',
+                                                      'location': 'Location',
+                                                      'grid_spacing': [.01, .01],
+                                                      'grid_spacing_unit': 'meters'}]
+            imaging_meta = {'optical_channel': optical_channel,
+                            'imaging_rate': rate,
+                            'device': device}
+            metadata['Ophys']['ImagingPlane'][0] = update_dict(metadata['Ophys']['ImagingPlane'][0], imaging_meta)
+
+            imaging_plane = nwbfile.create_imaging_plane(**metadata['Ophys']['ImagingPlane'][0])
```
bendichter, line 312: see optical channel

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -164,50 +254,223 @@ def get_num_channels(self):
...
+                    NwbImagingExtractor.add_devices(imaging=imaging,
+                                                    nwbfile=nwbfile,
+                                                    metadata=metadata)
+
+                    NwbImagingExtractor.add_two_photon_series(imaging=imaging,
+                                                              nwbfile=nwbfile,
+                                                              metadata=metadata)
+
+                    NwbImagingExtractor.add_epochs(imaging=imaging,
+                                                   nwbfile=nwbfile)
```
bendichter, line 430: do these get called in r+ mode?

bendichter:
this PR is way too big. What would NWB read and NWB write be in the same PR? Why are segmentation extractors and imaging extractors in the same PR? This should be at least 4 PRs. I am closing. Please open smaller PRs so we can incorporate them more efficiently. Thanks.

Saksham20:
> this PR is way too big. Why are changes to NWB read and NWB write be in the same PR? Why are segmentation extractors and imaging extractors in the same PR? This should be at least 4 PRs. I am closing. Please open smaller PRs so we can incorporate them more efficiently. When creating these new PRs, please make the changes suggested above. Thanks.

Had mistakenly manually merged this file with nwb_imaging, have reversed it now, there are only changes in the `write_Segmentation()` method and two other methods: `get_nwb_metadata() `and `nwb_metadata_recursive_update()`

Saksham20:
> this PR is way too big. Why are changes to NWB read and NWB write be in the same PR? Why are segmentation extractors and imaging extractors in the same PR? This should be at least 4 PRs. I am closing. Please open smaller PRs so we can incorporate them more efficiently. When creating these new PRs, please make the changes suggested above. Thanks.

Had mistakenly manually merged this file with nwb_imaging, have reverse it now, there are only changes in the `write_Segmentation()` method and two other methods: `get_nwb_metadata() `and `nwb_metadata_recursive_update()`

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: it take input nwb specific metadata though, since it is currently used only for that. 
I dont see it being useful any other places for now so IMO it would be better to have this name.

bendichter: this is a function for merging two dictionaries together.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +276,232 @@ def __init__(self, filepath, optical_channel_name=None,
...
-        return self._roi_idx
+    def __del__(self):
+        self._io.close()
 
-    @property
-    def accepted_list(self):
+    def get_accepted_list(self):
         if self._accepted_list is None:
-            return list(range(self.no_rois))
+            return list(range(self.get_num_rois()))
```
bendichter, line 288: we shouldn't assume ROIs are accepted

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +276,232 @@ def __init__(self, filepath, optical_channel_name=None,
...
         else:
             return np.where(self._accepted_list==1)[0].tolist()
```
bendichter, line 290:
```suggestion
            return np.where(self._accepted_list)[0].tolist()
```

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -287,280 +276,232 @@ def __init__(self, filepath, optical_channel_name=None,
...
-
-    def get_roi_locations(self, roi_ids=None):
-        if roi_ids is None:
-            return self.roi_locs
-        else:
-            roi_idx = [np.where(np.array(i) == self.roi_idx)[0] for i in roi_ids]
-            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
-            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
-            return self.roi_locs[:, roi_idx_]
+            return self._roi_locs.data[:].T
```
bendichter, line 300: you don't need to `return None`

bendichter (changes requested):

Saksham20:
@bendichter this branch is now ready to be merged. 
Added an additional change of renaming `nwb_metadata_recursive_update()` to `dict_recursive_update()`

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -22,6 +27,26 @@
     HAVE_NWB = False
 
 
+def dict_recursive_update(metadata_base, metadata_input):
+    return_dict = deepcopy(metadata_base)
+    for base_key, base_val in return_dict.items():
+        if metadata_input.get(base_key):
```
bendichter, line 33: I think `if metadata_input.get(base_key)` will always evaluate to `True` given the line above

Saksham20: metadata input may not have the same keys as the metadata_base since it is what the user inputs

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -22,6 +27,26 @@
...
+                    for base_val_num,(a,b) in enumerate(zip(base_val, metadata_input[base_key])):
+                        if isinstance(a, dict) and isinstance(b, dict):
+                            return_dict[base_key][base_val_num] = dict_recursive_update(a, b)
+                        elif isinstance(a, (int, float, str)) and isinstance(b, (int, float, str)):
+                            return_dict[base_key][base_val_num] = b
+                else:
+                    continue
+            else:
+                return_dict[base_key] = metadata_input[base_key]
+    return return_dict
```
bendichter, line 47:
```suggestion
def dict_recursive_update(base, input_):
    for key, val in input_.items():
        if isinstance(val, dict) and isinstance(base[key], dict):
            dict_recursive_update(base[key], val)
        elif isinstance(val, list) and isinstance(base[key], list):
            for i, input_list_item in enumerate(val):
                if len(base[key]) > i:
                    if isinstance(base[key][i], dict) and isinstance(input_list_item, dict):
                        dict_recursive_update(base[key][i], input_list_item)
                    else:
                        base[key][i] = input_list_item
                else:
                    base[key].append(input_list_item)
        else:
            base[key] = val
```

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: will create a separate PR to have these changes for all classes.

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: will create a separate PR to have these changes for all classes.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -268,138 +293,202 @@ def get_image_size(self):
...
+                              'Fluorescence': {'roi_response_series':[{'name': 'RoiResponseSeries',
+                                                                       'description': 'array of raw fluorescence traces'}]},
+                              'ImageSegmentation': {'plane_segmentations': [{'description': 'Segmented ROIs',
+                                                                            'name': 'PlaneSegmentation'}]},
+                              'ImagingPlane':[{'name': 'ImagingPlane',
+                                               'description': 'imaging plane description',
+                                               'excitation_lambda': 600.0,
+                                               'indicator': 'GCaMP',
+                                               'optical_channels': [{'name': 'OpticalChannel',
+                                                                     'emission_lambda': 500.0,
```
bendichter, line 316:
```suggestion
                                               'excitation_lambda': np.nan,
                                               'indicator': 'unknown',
                                               'optical_channels': [{'name': 'OpticalChannel',
                                                                     'emission_lambda': np.nan,
```

please don't make up data!!

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -268,138 +293,202 @@ def get_image_size(self):
...
+                              'TwoPhotonSeries': [{'name': 'TwoPhotonSeries'}]}}
+        # Optical Channel name:
+        for i in range(sgmextractor.get_num_channels()):
+            ch_name = sgmextractor.get_channel_names()[i]
+            if i == 0:
+                metadata['ophys']['ImagingPlane'][0]['optical_channels'][i]['name'] = ch_name
+            else:
+                metadata['ophys']['ImagingPlane'][0]['optical_channels'].append(dict(
+                    name=ch_name,
+                    emission_lambda=500.0,
```
bendichter, line 327:
```suggestion
                    emission_lambda=np.nan,
```
please don't make up data!

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: I've put my comments as python comments
```python
def dict_recursive_update(base, input_):
    for key, val in input_.items():
        if isinstance(val, dict) and isinstance(base[key], dict): #this would throw an error if base[key] is not found
            dict_recursive_update(base[key], val)
        elif isinstance(val, list) and isinstance(base[key], list):
# I was wanting to define the length of the list with the predefined base dict. If for example, base data expects 
# 1 roiresponse series (since only 1 was found in SegExt object, but the user inputs the names for two then, 
# it would add the second one to the base, the data for which is non existant
            for i, input_list_item in enumerate(val):
                if len(base[key]) < i: 
                    if isinstance(base[key][i], dict) and isinstance(input_list_item, dict):
                        dict_recursive_update(base[key][i], input_list_item)
                    else:
                        base[key][i] = input_list_item
                else:
                    base[key].append(input_list_item)#input_ would then be expected to have all correct nwbfile specific fields, it would throw an error if the user inputs a wrong field.
        else:
            base[key] = val # if isinstance(base[key][i], dict) is true but input  isinstance(input_list_item, str/int) then it would be a problem
```

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: yes, thanks for catching that, fixed

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: yes, thanks for catching that, fixed

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: fixed

roiextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: true, but it's in the signature of the base class method

roiextractors/extractors/nwbextractors/nwbextractors.py
bendichter: I didn't understand all of your comments but I did fix one error you caught. How about this?

```python
def dict_recursive_update(base, input_):
    for key, val in input_.items():
        if key in base and isinstance(val, dict) and isinstance(base[key], dict):
            dict_recursive_update(base[key], val)
        elif key in base and isinstance(val, list) and isinstance(base[key], list):
            for i, input_list_item in enumerate(val):
                if len(base[key]) < i:
                    if isinstance(base[key][i], dict) and isinstance(input_list_item, dict):
                        dict_recursive_update(base[key][i], input_list_item)
                    else:
                        base[key][i] = input_list_item
                else:
                    base[key].append(input_list_item)
        else:
            base[key] = val
```

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -268,138 +291,202 @@ def get_image_size(self):
...
-
-        # Subject:
-        if 'Subject' in metadata_dict:
-            nwbfile.subject = Subject(**metadata_dict['Subject'])
-
-        # Device:
-        if isinstance(metadata_dict['ophys']['Device'], list):
-            for devices in metadata_dict['ophys']['Device']:
-                nwbfile.create_device(**devices)
+    def get_nwb_metadata(sgmextractor):
```
bendichter, line 294: I don't understand the purpose of this method.

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -268,138 +291,202 @@ def get_image_size(self):
...
+
+                # ImagingPlane:
+                image_plane_name = 'ImagingPlane' if plane_no_loop == 0 else f'ImagePlane_{plane_no_loop}'
+                if image_plane_name not in nwbfile.imaging_planes.keys():
+                    input_kwargs = dict(
+                        name=image_plane_name,
+                        description='no description',
+                        device=nwbfile.get_device(metadata_base_common['ophys']['Device'][0]['name']),
+                        excitation_lambda=np.nan,
+                        imaging_rate=1.0,
```
bendichter, line 412: please do not make up data!!

roiextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: @bendichter there is a separate PR for the NWB write method (https://github.com/catalystneuro/roiextractors/pull/20)

Should I just port the changes here? so it includes all write segmentation methods

roiextractors/extractors/nwbextractors/nwbextractors.py
Saksham20: okay this works, i've picked this method in the new commit

roiextractors/extractors/nwbextractors/nwbextractors.py
bendichter:
```suggestion
                        imaging_rate=np.nan,
```

PR #21, opened 2020-08-17
updated and added yaml files for all classes

Saksham20:
Updating yaml files for each of the classes, 
If you remember, we discussed that this would be necessary for filling in metadata by the user before writing to nwb. Each class has minor differences. 
I have yet to implement the method to generate a schema out of these for `nwb_conversion_tools `as we discussed.


metadatafiles/nct_ophys_metafile_cnmfe.yaml
```diff
@@ -23,28 +23,35 @@ Subject:
   species: species
   weight: weight number
   date_of_birth: 2010-02-02 00:00:00
-ophys:
+Ophys:
   Device:
   - name: ophys_Device1
     description: ophys_device1_description
```
bendichter, line 29: default should be "no description"

PR #22, opened 2020-08-17
Property methods updates

Saksham20:
All the classes of SegmentationExtractor: 1) sima, 2) cnmfe 3)extract 4)suite2p, 5) nwb 6) numpy 
are changed here. 
There are a number of properties which were redundant since there are get methods for each. These were defined in the base class by Allessio so I have removed them from here. 

Semantically: 
1. Properties are removed here since they are in the base class `SegmentationExtractor`.
2. SOme methods and properties you will see are not removed by renamed: matching those renamed in the base class abstract methods in master branch. 

Note: 
There are a few minor additional changes in Suite2p which are there since now we have agreed to have only 1 plane enforced.

### DIFF
--- a/roiextractors/extractors/numpyextractors/numpyextractors.py
+++ b/roiextractors/extractors/numpyextractors/numpyextractors.py
@@ -269,8 +269,8 @@ class NumpySegmentationExtractor(SegmentationExtractor):
                 np.append(temp, self.pixel_masks[self.pixel_masks[:, 3] == roiid, :], axis=0)
         return temp[1::, :]
 
-    def get_movie_framesize(self):
-        return self.image_dims
+    def get_images(self):
+        return {'Images': {'meanImg': self._summary_image}}
 
     def get_image_size(self):
         return self._movie_dims
--- a/roiextractors/extractors/numpyextractors/numpyextractors.py
+++ b/roiextractors/extractors/numpyextractors/numpyextractors.py
@@ -1,8 +1,7 @@
 import numpy as np
 from pathlib import Path
-from ...segmentationextractor import SegmentationExtractor
-from ...imagingextractor import ImagingExtractor
-from ...extraction_tools import get_video_shape
+from roiextractors import SegmentationExtractor, ImagingExtractor
+from roiextractors.extraction_tools import get_video_shape
 
 
 # TODO this class should also be able to instantiate an in-memory object (useful for testing)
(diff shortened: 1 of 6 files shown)


bendichter:
@Saksham20 the proper way to link issues to PRs is to write "fix #25" in the description of the PR. Can you please change it?

roiextractors/extractors/numpyextractors/numpyextractors.py
```diff
@@ -1,8 +1,7 @@
 import numpy as np
 from pathlib import Path
-from ...segmentationextractor import SegmentationExtractor
-from ...imagingextractor import ImagingExtractor
-from ...extraction_tools import get_video_shape
+from roiextractors import SegmentationExtractor, ImagingExtractor
+from roiextractors.extraction_tools import get_video_shape
```
bendichter, line 4: use relative imports

roiextractors/extractors/numpyextractors/numpyextractors.py
```diff
@@ -222,47 +208,31 @@ def roi_locs(self):
...
             roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
-            return self.roi_locs[:, roi_idx_]
+            return self.roi_locations[:, roi_idx_]
 
     def get_num_frames(self):
-        return self.num_of_frames
-
-    def get_sampling_frequency(self):
-        return self.samp_freq
+        return self._roi_response.shape[1]
```
bendichter, line 235: Is this the sampling frequency?

roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
```diff
@@ -2,7 +2,7 @@
 import h5py
 from roiextractors import SegmentationExtractor
 from lazy_ops import DatasetView
-
+from roiextractors.extraction_tools import _pixel_mask_extractor
```
bendichter, line 5: use relative imports

roiextractors/extractors/numpyextractors/numpyextractors.py
Saksham20: yes, this is the samp rate of the traces.

bendichter: maybe this is the length of the response, i.e. the number of samples?

Saksham20:
```
def get_num_frames(self):
      return self._roi_response.shape[1]
```
this refers to the number of frames. 
`get_sampling_frequency()` method is in the parent class

bendichter: oh, my fault, I was not reading the line changes correctly

Saksham20: When I open the file mentioned here, it does not show what you're mentioning lol
This is what I see:
image

PR #23, opened 2020-08-17
fix#5: remove non-existant fields for all classes

Saksham20:
There were some fields that I had added initially to cnmfe, extract, sima, numpy. These I think only over-complicate things, better to insert then on an as needed basis. 

Removing redundant fields:
**background**-__image_masks_, _roi_response_series_, snr _roi_resp_series_, _total_time_, _Filetype_, _raw_images_ ). These are not found for any pipeline.

### DIFF
--- a/roiextractors/extractors/numpyextractors/numpyextractors.py
+++ b/roiextractors/extractors/numpyextractors/numpyextractors.py
@@ -94,91 +94,59 @@ class NumpySegmentationExtractor(SegmentationExtractor):
     all data must be entered manually as arguments.
     """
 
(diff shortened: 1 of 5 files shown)


bendichter:
@Saksham20 these changes look great!

I would probably have chosen to use individual variables instead of `self._roi_response_dict` that holds all three response types, because it would make it easier to standardizing a naming convention across the different processing systems. If a certain response type were no present, I would just make that `None`. Is there a reason you chose to do it this way?

Regardless, this is a big improvement in standardizing the internal representation across the different processing systems, so I'm going to merge and we can discuss the dict vs attr question later.

PR #25, opened 2020-08-17
remove get_frame and add decorators to check input

alejoe91:
* remove get_frame
* add input check validator decorators


roiextractors/extractors/bioformatsimagingextractors/bioformatsimagingbaseextractor.py
```diff
@@ -57,31 +58,20 @@ def _start_javabridge_vm(self):
...
         planes = np.zeros((len(frame_idxs), self._size_x, self._size_y))
         for i, frame_idx in enumerate(frame_idxs):
             plane = self._reader.read(t=frame_idx).T
             planes[i] = plane
         return planes
 
+    # TODO make this memmap if too large
+    @check_get_videos_args
     def get_video(self, start_frame=None, end_frame=None, channel=0):
-        if start_frame is None:
```
bendichter, line 75: Is there a reason to keep `get_video`? Is there a use-case that can't be handled by `get_frames`?

roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -51,26 +51,17 @@ def __init__(self, file_path, mov_field='mov', sampling_frequency=None,
...
     def get_frames(self, frame_idxs, channel=0):
-        frame_idxs = np.array(frame_idxs)
-        assert np.all(frame_idxs < self.get_num_frames())
-        sorted_frame_idxs, sorting_inverse = np.sort(frame_idxs, return_inverse=True)
-        return self._video[channel, sorted_frame_idxs][:, sorting_inverse]
+        if frame_idxs.size > 1 and np.all(np.diff(frame_idxs) > 0):
+            return self._video[channel, frame_idxs]
+        else:
+            sorted_frame_idxs, sorting_inverse = np.sort(frame_idxs, return_inverse=True)
+            return self._video[channel, sorted_frame_idxs][:, sorting_inverse]
```
bendichter, line 60: Do we want this to continue to be a memmap for HDF5? `lazy_ops` can do that with:

`lazy_ops.DatasetView(self._video).lazy_slice[channel, frame_idxs]`

and 

`lazy_ops.DatasetView(self._video).lazy_slice[channel, sorted_frame_idxs].lazy_slice[:, sorting_inverse]`

roiextractors/extraction_tools.py
```diff
@@ -1,12 +1,16 @@
...
 from pathlib import Path
+from functools import wraps
+from spikeextractors.extraction_tools import cast_start_end_frame
 
 ArrayType = Union[list, np.array]
 PathType = Union[str, Path]
 NumpyArray = Union[np.array, np.memmap]
 DtypeType = [str, np.dtype]
 
+MAX_FRAMES = 10000
```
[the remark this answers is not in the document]
alejoe91: I think I misunderstood this comment! https://github.com/catalystneuro/roiextractors/pull/10#discussion_r470180773

Removing it :)

roiextractors/extractors/bioformatsimagingextractors/bioformatsimagingbaseextractor.py
alejoe91: You can do everything with `get_frames`, but personally I'd like to keep both as `get_video` is more compact for continuous data. If you prefer to remove I can do it.

roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
alejoe91: Sure let's do that!

roiextractors/extractors/bioformatsimagingextractors/bioformatsimagingbaseextractor.py
bendichter: OK, I'm just thinking about trying to keep the API minimal and consistent. How about we define `get_video` in `ImagingExtractor` as `return self.get_frames(range(start_frame, end_frame), channel=channel)` and overwrite as necessary? In this case, I think they would do the same thing, so I don't think overwriting would be necessary.

alejoe91: That's a great idea :)

Performance-wise, slicing should be better, but we can use the decorator of the get_frames to make a slice in case frame_idxs are continuous. 

Are you ok in keeping the decorators for now, or should I just remove them and make the function?

bendichter: Yes, let's leverage a continuous slice where we can for performance, but as you said this can be internal to `get_frames`. I'm not sure if I'd want to do it with a decorator though.

alejoe91:
@bendichter this is ready to go IMO!

PR #32, opened 2020-08-19
pixel masks are nx3 from nx4

Saksham20:
this was a fix to an error which was missed in the previous branches. Pixel masks need to a list of length no_rois and each element be of shape no_pixels x 3,

### DIFF
--- a/roiextractors/extraction_tools.py
+++ b/roiextractors/extraction_tools.py
@@ -40,8 +40,8 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
     image_mask: np.ndarray
     """
     image_mask = np.zeros(list(image_shape)+[len(_roi_ids)])
-    for no, rois in enumerate(_roi_ids):
-        for x, y, wt in pixel_mask[rois]:
+    for no, roi in enumerate(_roi_ids):
+        for x, y, wt in pixel_mask[roi]:
             image_mask[int(x), int(y), no] = wt
     return image_mask
 
--- a/roiextractors/extraction_tools.py
+++ b/roiextractors/extraction_tools.py
@@ -33,16 +33,16 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
     pixel_mask: list
         list of pixel masks (no pixels X 3)
     _roi_ids: list
-    image_shape: list
+    image_shape: array_like
     
     Returns
     -------
     image_mask: np.ndarray
     """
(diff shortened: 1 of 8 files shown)


bendichter:
Internally, we should either store this data as pixel or voxel mask, not both.

Saksham20:
> Internally, we should either store this data as pixel or image mask, not both. It looks like NumpyExtractor is storing both

yes, good point, check 2978f85b4e881728b7a6c855f607d22b250e3b4a,
I moved the pixel masks to the base class. 
This is not stored anymore, but only retrieved using the get method

Saksham20:
@bendichter 
If you could also review and merge this branch after merging the other two pending branches #20 #18 It would be great

roiextractors/extraction_tools.py
```diff
@@ -33,16 +33,16 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
...
     _roi_ids: list
-    image_shape: list
+    image_shape: array_like
     
     Returns
     -------
     image_mask: np.ndarray
     """
-    image_mask = np.zeros(image_shape + [len(pixel_mask)])
-    for rois in range(image_mask.shape[2]):
```
Saksham20, line 43: @Saksham20 will change 'rois' to roi

roiextractors/extraction_tools.py
```diff
@@ -33,16 +33,16 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
...
+    image_mask = np.zeros(list(image_shape)+[len(_roi_ids)])
+    for no, rois in enumerate(_roi_ids):
```
alejoe91, line 43:
```suggestion
    for no, roi in enumerate(_roi_ids):
```

roiextractors/segmentationextractor.py
```diff
@@ -188,10 +187,10 @@ def get_roi_pixel_masks(self, roi_ids=None) -> np.array:
 
         Returns
         -------
-        pixel_masks: numpy.ndarray
-            3-D array with weight for each pixel of the rroi: image_height X image_width X length(roi_ids)
+        pixel_masks: list
+            list of length number of rois, each element is a 2-D array os shape (no-pixels, 2)
```
bendichter, line 191:
```suggestion
        pixel_masks: list of lists
            list of length number of rois, each element is a 2-D array os shape (no-pixels, 2)
```

PR #34, opened 2020-08-19
CaImAn segmentationextractor

### LINKED ISSUE #29: support CaImAn segmentation

Saksham20:
fix #29 

class to read from a caiman output hdf5 file.


roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,151 @@
...
+            roi_idx_ = self.roi_ids
+        else:
+            roi_idx = [np.where(np.array(i) == self.roi_ids)[0] for i in roi_ids]
+            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
+            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
+        self.pixel_masks = _pixel_mask_extractor(self.get_roi_image_masks(roi_idx_), range(len(roi_idx_)))
+        return self.pixel_masks
+
+    def get_images(self):
+        return {'Images': {'meanImg': self._summary_image_read()}}
```
bendichter, line 148: 1. `Cn` is the correlation image, not the mean image. It is computed by taking the correlation that each pixel has with each of its neighbors over time. The pixels that have a higher value are generally places where there is a neuron. Cn is useful to get a quick sense of how neurons are dispersed in your image, but doesn't give you segmentation information.
2. As we discussed yesterday, I prefer attributes instead of dicts unless there will be user-defined fields. I would rather have `meanImg` be an attribute of the class.
3. Why the outer `{'Images': ...}`? I understand the inner dict, in case there are other image types (though I would prefer attrs of the class), but I don't see why we have a dict with a single key: `'Images'`.

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,151 @@
...
+        self._roi_response_dict = {'Fluorescence': self._roi_response,
+                                   'Neuropil':self._trace_extractor_read('C'),
+                                   'Deconvolved':self._trace_extractor_read('S')}
+
+        self._raw_movie_file_location = self._dataset_file['params']['data']['fnames'][0].decode('utf-8')
+        self._sampling_frequency = self._dataset_file['params']['data']['fr'].value
+        # file close:
+        # self._file_close()
+
+    def _file_close(self):
```
bendichter, line 40: See use of `__del__` e.g. https://github.com/catalystneuro/roiextractors/blob/4e5117ecabcb2340ca72166e32db96d411add351/roiextractors/extractors/bioformatsimagingextractors/bioformatsimagingbaseextractor.py#L26-L28

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,151 @@
...
+        return f
+
+    def _image_mask_sparse_read(self):
+        roi_ids = self._dataset_file['estimates']['A']['indices']
+        masks = self._dataset_file['estimates']['A']['data']
+        ids = self._dataset_file['estimates']['A']['indptr']
+        return masks, roi_ids, ids
+
+    def _trace_extractor_read(self, field):
+        extracted_signals = self._dataset_file['estimates'][field] # lazy read dataset)
```
bendichter, line 54: it would be better to use lazy_ops to make this lazy

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,151 @@
...
+        Parameters
+        ----------
+        filepath: str
+            The location of the folder containing caiman *.hdmf output file.
+        """
+        SegmentationExtractor.__init__(self)
+        self.filepath = filepath
+        self._dataset_file = self._file_extractor_read()
+        self._roi_response = self._trace_extractor_read('F_dff')
+        self._masks, self._mask_roi_ids, self._mask_ids = self._image_mask_sparse_read()
```
bendichter, line 30: Why would `self_mask_roi_ids` be different from `self._roi_ids`?

Saksham20: these are different variables. 
1. `self._masks`  is a 1d array containing pixel weights of all pixels in an image_mask across all ROIs. There are two other arrays keeping track two indices: 
1. `self._mask_roi_ids` are ids for the pixel relative to each ROI (flattened). Since there is one location for each this array is of length `self._masks` 
2. `self._mask_ids` index the  `self._masks` and mark the change in ROI. So this is length no_rois. 

`self._image_mask_sparse_read()` just pulls out these 3 arrays.  `self.get_roi_image_masks` has the logic to build a 3d image_mask array out of that

bendichter: I'd rather we fetch these with the necessary code to massage them into the same internal representation as other SegmentationExtractors, rather than keeping the same form as CaImAn under the hood

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: `extracted_signals `is a hdf5 dataset class. I assume that is not loaded into memory unless sliced etc, is that correct? 
If true, does calling `lazy_ops.DatasetView `over that (`DatasetView(extracted_signals)`) have further benefits?

bendichter: oh, `field` is a string. OK, I misunderstood, this looks good.

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed 007150d079ad5a8c25664a4c186db9f14d7c6821

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,151 @@
...
+            return self.roi_locations[:, roi_idx_]
+
+    def get_num_frames(self):
+        return self._roi_response.shape[1]
+
+    def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name='Fluorescence'):
+        if start_frame is None:
+            start_frame = 0
+        if end_frame is None:
+            end_frame = self.get_num_frames() + 1
```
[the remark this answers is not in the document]
Saksham20: oh, good to know, fixed here: 6e135bd33609b8efd1f1608d4e605660a1cfed51

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed here 87f31df3c08548e722c9396defb3fe1a3587f41c
generating image_masks from the h5py file would be expensive since it requires a reshape operation, can this be done lazily?
Keeping this in mind, I have kept `self.image_masks = None ` in `__init__`

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: made some changes here: 8d14fecd7902d7d17e390c322dffe768ea4d5b8f
I have unpacked the images and roi_response_dict into independent attributes, defined as None in the base class. 
THey are called in the downstream get methods based on the field that a user desires

roiextractors/extractors/caiman/caimansegmentationextractor.py
bendichter: reshape cannot be done lazily currently

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,153 @@
...
+        return self._roi_response.shape[1]
+
+    def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name='Fluorescence'):
+        if roi_ids is None:
+            roi_idx_ = range(self.get_num_rois())
+        else:
+            roi_idx = [np.where(np.array(i) == self.roi_ids)[0] for i in roi_ids]
+            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
+            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
+        traces = [getattr(self, i) for i in self.__dict__.keys() if name.lower() in i]
```
bendichter, line 115: what's going on here? This seems pretty complex. You can just use `traces = getattr(self, name)`?

Saksham20: I'm just trying to map the desired trace ( like 'Fluorescence') to the corresponding internal variable that stores this trace
It just frees us to use internal names that could be more programmer friendly than user friendly: "`Fluorescence`" maps to `self._roi_response_fluorescence`, 'Neuropil' to `self._roi_response_neuropil` ..

Saksham20: I'm using __dict__ to get a list of all attribute names to search over. Is there a better way of doing this?

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -84,7 +84,38 @@ def roi_locations(self):
...
+            #create base groups:
+            estimates = f.create_group('estimates')
+            params = f.create_group('params')
+            #adding to estimates:
+            if segmentation_object._roi_response_neuropil:
+                estimates.create_dataset('C',data=segmentation_object._roi_response_neuropil)
+            estimates.create_dataset('F_dff', data=segmentation_object._roi_response_fluorescence)
+            if segmentation_object._roi_response_deconvolved:
+                estimates.create_dataset('S', data=segmentation_object._roi_response_deconvolved)
+            if segmentation_object._images_mean:
```
bendichter, line 100: Cn is not a mean image

roiextractors/extractors/caiman/caimansegmentationextractor.py
bendichter: Could you populate self.image_masks in the constructor?

roiextractors/extractors/caiman/caimansegmentationextractor.py
bendichter: oh.. so you are searching if `'Fluorescence'.lower()` is in `'_roi_response_fluorescence'`. OK, now I understand. I think this will likely cause problems if there is another class attribute that contains 'fluorescence' in the name for some reason. If you feel there is a need, you can create a dictionary that contains e.g. `{'Fluorescence': '_roi_response_fluorescence'}`

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed here https://github.com/catalystneuro/roiextractors/pull/34/commits/82f90e6fe7770dc07c6e3f9024e870986d8c123f

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: would that be necessary? 
image_masks is necessary for every subclass to re-define anyway since its an important attribute

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+        self.filepath = filepath
+        self._dataset_file = self._file_extractor_read()
+        self._roi_response = self._trace_extractor_read('F_dff')
+        self._roi_response_fluorescence = self._roi_response
+        self._roi_response_neuropil = self._trace_extractor_read('C')
+        self._roi_response_deconvolved = self._trace_extractor_read('S')
+        self._images_correlation = self._summary_image_read()
+        self._raw_movie_file_location = self._dataset_file['params']['data']['fnames'][0].decode('utf-8')
+        self._sampling_frequency = self._dataset_file['params']['data']['fr'][()]
+        self.image_masks = None
```
bendichter, line 37: can we set this in the init?

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+        ids = self._dataset_file['estimates']['A']['indptr']
+        return masks, roi_ids, ids
+
+    def _trace_extractor_read(self, field):
+        extracted_signals = self._dataset_file['estimates'][field] # lazy read dataset)
+        return extracted_signals
+
+    def _summary_image_read(self):
+        if self._dataset_file['estimates'].get('Cn'):
+            summary_images_ = self._dataset_file['estimates']['Cn']
```
bendichter, line 58: why `summary_images_`, plural?

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+            The location of the folder containing caiman *.hdmf output file.
+        """
+        SegmentationExtractor.__init__(self)
+        self.filepath = filepath
+        self._dataset_file = self._file_extractor_read()
+        self._roi_response = self._trace_extractor_read('F_dff')
+        self._roi_response_fluorescence = self._roi_response
+        self._roi_response_neuropil = self._trace_extractor_read('C')
+        self._roi_response_deconvolved = self._trace_extractor_read('S')
+        self._images_correlation = self._summary_image_read()
```
bendichter, line 34: why is `images` plural?

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+            return None
+
+    def get_accepted_list(self):
+        accepted = self._dataset_file['estimates']['idx_components']
+        if len(accepted.shape)==0:
+            accepted = list(range(self.no_rois))
+        return accepted
+
+    def get_rejected_list(self):
+        return [a for a in range(self.no_rois) if a not in set(self.get_accepted_list())]
```
bendichter, line 70: rejected is not the same as not accepted. There should be a separate rejected list.

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+            )]
+            roi_location[:, i] = [((max_mask_roi_id+1)%(self.image_size[0]+1))-1,#assuming order='F'
+                                  ((max_mask_roi_id+1)//(self.image_size[0]+1))]
+            if roi_location[0,i]<0:
+                roi_location[0,i]=0
+        return roi_location
+
+    @staticmethod
+    def write_segmentation(segmentation_object, savepath, **kwargs):
+        from scipy.sparse import csc_matrix
```
bendichter, line 88: strange placement for this import. Is there a reason this is here and not at the top?

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+        plane_no = kwargs.get('plane_no', 0)
```
bendichter, line 89: default is 0 planes? That doesn't name sense

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,170 @@
...
+        filename = os.path.basename(savepath)
+        savepath_folder = os.path.join(os.path.dirname(savepath), f'Plane_{plane_no}')
+        savepath = os.path.join(savepath_folder, filename)
```
bendichter, line 92: All planes should be stored in the same NWB file

roiextractors/segmentationextractor.py
```diff
@@ -17,6 +17,11 @@ class SegmentationExtractor(ABC, BaseExtractor):
     def __init__(self):
         BaseExtractor.__init__(self)
         self._sampling_frequency = None
+        self._roi_response_fluorescence = None
+        self._roi_response_neuropil = None
+        self._roi_response_deconvolved = None
+        self._images_correlation = None
+        self._images_mean = None
```
bendichter, line 24:
```suggestion
        self._image_mean = None
```

roiextractors/segmentationextractor.py
```diff
@@ -17,6 +17,11 @@ class SegmentationExtractor(ABC, BaseExtractor):
     def __init__(self):
         BaseExtractor.__init__(self)
         self._sampling_frequency = None
+        self._roi_response_fluorescence = None
+        self._roi_response_neuropil = None
+        self._roi_response_deconvolved = None
+        self._images_correlation = None
```
bendichter, line 23:
```suggestion
        self._image_correlation = None
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: In the caiman file, the `image_masks `are stored as a sparse matrix. Since there is no way to lazily reshape and create an array out of the sparse matrix would load the whole thing into memory. 
I'm trying to avoid this so I set image_masks as None. Would it then be better to have image_masks as an internal attribute for all classes then? 

I'm calling `self._image_mask_sparse_read()` within the `get_image_masks()` which only returns `image_masks` for the requested ROI ids, thus possibly loading a smaller value into memory.

bendichter: let's just not make it lazy for now

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: yes, thanks for pointing that out, caiman does have another field that explicitly stores the rejected values. I've made the changes here https://github.com/catalystneuro/roiextractors/pull/34/commits/58ee2abc76309e4e928f271c27930756cdb8a45b
Though for other classes like extract, cnmfe and sima, there is no such info so its just the setdiff with the accepted.

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: yeah, i had initially kept it since scipy is used nowhere else in the repo. But since its a common package, it makes sense to add that as a dependency overall and move it on top

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: yes your're right, its a bit misleading, i'll take a note and change this for all classes together in another PR

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: hmm, okay, fixed that https://github.com/catalystneuro/roiextractors/pull/34/commits/5e3fd1678ae152fb198b8958dc3f5d593e94dacd

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+        accepted = self._dataset_file['estimates']['idx_components']
+        if len(accepted.shape)==0:
+            accepted = list(range(self.no_rois))
+        return accepted
+
+    def get_rejected_list(self):
+        rejected = self._dataset_file['estimates']['idx_components_bad']
+        if len(rejected.shape) == 0:
+            rejected = [a for a in range(self.no_rois) if a not in set(self.get_accepted_list())]
+        return rejected
```
bendichter, line 77: should return None if idx_components_bad is missing

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+    def _file_extractor_read(self):
+        f = h5py.File(self.filepath, 'r')
+        return f
+
+    def _image_mask_sparse_read(self):
+        roi_ids = self._dataset_file['estimates']['A']['indices']
+        masks = self._dataset_file['estimates']['A']['data']
+        ids = self._dataset_file['estimates']['A']['indptr']
+        _image_mask = np.reshape(csc_matrix((masks, roi_ids, ids), shape=(np.prod(self.get_image_size()),self.no_rois)).toarray(),
+            [self.get_image_size()[0],self.get_image_size()[1],-1],order='F')
```
bendichter, line 52:
```python
image_mask_in = csc_matrix((masks, roi_ids, ids),
shape=(np.prod(self.get_image_size()), self.no_rois)).toarray()

self._image_mask = np.reshape(image_mask_in, (*self.get_image_size(), -1), order='F')
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+        return masks, roi_ids, ids, _image_mask
```
bendichter, line 53: just return image_mask

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+    def _trace_extractor_read(self, field):
+        if self._dataset_file['estimates'].get(field):
+            return self._dataset_file['estimates'][field] # lazy read dataset)
+        else:
+            return None
+
+    def _summary_image_read(self):
+        if self._dataset_file['estimates'].get('Cn'):
+            return np.array(self._dataset_file['estimates']['Cn']).T
+        else:
```
bendichter, line 64:
```suggestion
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+            return None
```
bendichter, line 65:
```suggestion
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+        ids = self._dataset_file['estimates']['A']['indptr']
+        _image_mask = np.reshape(csc_matrix((masks, roi_ids, ids), shape=(np.prod(self.get_image_size()),self.no_rois)).toarray(),
+            [self.get_image_size()[0],self.get_image_size()[1],-1],order='F')
+        return masks, roi_ids, ids, _image_mask
+
+    def _trace_extractor_read(self, field):
+        if self._dataset_file['estimates'].get(field):
+            return self._dataset_file['estimates'][field] # lazy read dataset)
+        else:
+            return None
```
bendichter, line 59: you dont need else return None. If the function ends and there is no return, it will return None

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _summary_image_read(self):
```
bendichter, line 61:
```suggestion
    def _summary_image_read(self, image_name):
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+            #create base groups:
+            estimates = f.create_group('estimates')
+            params = f.create_group('params')
+            #adding to estimates:
+            if segmentation_object._roi_response_neuropil is not None:
+                estimates.create_dataset('C',data=segmentation_object._roi_response_neuropil)
+            estimates.create_dataset('F_dff', data=segmentation_object._roi_response_fluorescence)
+            if segmentation_object._roi_response_deconvolved is not None:
+                estimates.create_dataset('S', data=segmentation_object._roi_response_deconvolved)
+            if segmentation_object._images_mean is not None:
```
bendichter, line 111: Cn is not mean image

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,159 @@
...
+
+class CaimanSegmentationExtractor(SegmentationExtractor):
+    """
+    This class inherits from the SegmentationExtractor class, having all
+    its funtionality specifically applied to the dataset output from
+    the \'CNMF-E\' ROI segmentation method.
+    """
+    extractor_name = 'CaimanSegmentation'
+    installed = True  # check at class level if installed or not
+    is_writable = False
```
bendichter, line 17:
```suggestion
    is_writable = True
```

Saksham20 (commented):
this change reflects #45

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed https://github.com/catalystneuro/roiextractors/pull/34/commits/72eb48defa694c7207d8b218e8f7d85b4164d0af

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed https://github.com/catalystneuro/roiextractors/pull/34/commits/2e5cf3a7ae13566cbef96279a50fece9129e48a7

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed https://github.com/catalystneuro/roiextractors/pull/34/commits/2e5cf3a7ae13566cbef96279a50fece9129e48a7

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: https://github.com/catalystneuro/roiextractors/pull/34/commits/eca3a750d28f3ffe32dcd9a34fa99fc1765d7acf

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: https://github.com/catalystneuro/roiextractors/pull/34/commits/eca3a750d28f3ffe32dcd9a34fa99fc1765d7acf

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: https://github.com/catalystneuro/roiextractors/pull/34/commits/eca3a750d28f3ffe32dcd9a34fa99fc1765d7acf

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: https://github.com/catalystneuro/roiextractors/pull/34/commits/61677f0c32af4de5e5eac2faf37e322607da3165

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: fixed

roiextractors/segmentationextractor.py
Saksham20: naming have been fixed already

roiextractors/segmentationextractor.py
Saksham20: naming have been fixed already

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: naming have been fixed already

roiextractors/extractors/caiman/caimansegmentationextractor.py
Saksham20: naming have been fixed already

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,122 @@
...
+        rejected = self._dataset_file['estimates']['idx_components_bad']
+        if len(rejected.shape) > 0:
+            return rejected
+
+    @property
+    def roi_locations(self):
+        num_ROIs = self.get_num_rois()
+        roi_locations = np.ndarray([2, num_ROIs], dtype='int')
+        for i in range(num_ROIs):
+            temp = np.where(self.image_masks[:, :, i] == np.amax(self.image_masks[:, :, i]))
```
bendichter, line 77: a bit odd to use the max here. I think a weighted average might work better

roiextractors/extractors/caiman/caimansegmentationextractor.py
bendichter: It still looks to me like you are calling Cn the image_mean

roiextractors/segmentationextractor.py
bendichter: then marge it in

roiextractors/segmentationextractor.py
bendichter: then marge it in

bendichter (changes requested):

roiextractors/extractors/caiman/caimansegmentationextractor.py
bendichter: something like `[np.mean(image * dim) for dim in np.mgrid[0:5,0:5]]`

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,122 @@
...
+            #create base groups:
+            estimates = f.create_group('estimates')
+            params = f.create_group('params')
+            #adding to estimates:
+            if segmentation_object._roi_response_neuropil is not None:
+                estimates.create_dataset('C',data=segmentation_object._roi_response_neuropil)
+            estimates.create_dataset('F_dff', data=segmentation_object._roi_response_fluorescence)
+            if segmentation_object._roi_response_deconvolved is not None:
+                estimates.create_dataset('S', data=segmentation_object._roi_response_deconvolved)
+            if segmentation_object._images_mean is not None:
```
bendichter, line 99:
```suggestion
            if segmentation_object._image_correlation is not None:
```

roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -0,0 +1,122 @@
...
+                estimates.create_dataset('Cn', data=segmentation_object._images_mean)
```
bendichter, line 100:
```suggestion
                estimates.create_dataset('Cn', data=segmentation_object._images_correlation)
```

PR #41, opened 2020-08-25
Generate a toy example for testing

alejoe91:
- generate toy example videos by convolving circular ROIs with spike trains (generated by spikeextractors) and a decay response

- change to the `NumpyImagingExtractor` to allow the creation of an imaging object with an in-memory video

- change to the `NumpySextractgmentationeor` to allow the creation of in-memory segmentation object


roiextractors/example_datasets/toy_example.py
```diff
@@ -1 +1,150 @@
...
+
+                if np.linalg.norm(p - mean) < roi_size:
+                    pixels.append(p)
+                    if mode == 'uniform':
+                        image[i, j] = 1
+                    elif mode == 'gaussian':
+                        image[i, j] = _gaussian(i, mean[0], roi_size) + _gaussian(j, mean[1], roi_size)
+                    else:
+                        raise Exception("'mode' can be 'uniform' or 'gaussian'")
+        roi_pixels.append(np.array(pixels))
```
bendichter, line 56: https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html might be cleaner

bendichter: or at least vectorizing so it's not on a for loop

roiextractors/example_datasets/toy_example.py
```diff
@@ -1 +1,150 @@
...
+
+    # convolve response with ROIs
+    traces = np.zeros((len(sort.get_unit_ids()), rec.get_num_frames()))
+    frames = rec.get_num_frames()
+    for u_i, unit in enumerate(sort.get_unit_ids()):
+        for s in sort.get_unit_spike_train(unit):
+            if s + len(resp) < frames:
+                traces[u_i, s:s + len(resp)] += resp
+            else:
+                traces[u_i, s:] = resp[:frames - s]
```
bendichter, line 122: https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html might be cleaner here

roiextractors/example_datasets/toy_example.py
bendichter: @alejoe91

roiextractors/example_datasets/toy_example.py
bendichter: @alejoe91

Saksham20:
@alejoe91 l I have added dff, neuropil and deconvolved traces as argument to `NumpySegmentationExtractor`, please review. 
We still have to generate the traces in **toy_example.py**, I think it would be okay to have a random numbers for the 3 traces for now, what do you think?

roiextractors/example_datasets/toy_example.py
alejoe91: @bendichter convolve requires two signals. In this case, since spike trains are delta functions, this is easier

roiextractors/example_datasets/toy_example.py
alejoe91: @bendichter the gaussian filter is different, as it convolves the entire image with a gaussian kernel. 
Since each gaussian have different mean, I think we can't avoid a loop here (it's very fast anyways :))

roiextractors/extractors/numpyextractors/numpyextractors.py
```diff
@@ -131,24 +138,46 @@ def __init__(self, image_masks, signal,
         if isinstance(image_masks, (str, Path)):
             image_masks = Path(image_masks)
             signal = Path(signal)
+            dff = Path(signal) if dff is not None else Path('none.npy')
```
alejoe91, line 141: this will fail when loading the `none.py` as it doesn't exist. I would check later:

if dff.is_file():
      load

alejoe91:
> @alejoe91 l I have added dff, neuropil and deconvolved traces as argument to `NumpySegmentationExtractor`, please review.
> We still have to generate the traces in **toy_example.py**, I think it would be okay to have a random numbers for the 3 traces for now, what do you think?

This is done with https://github.com/catalystneuro/roiextractors/pull/41/commits/06a66229429cccab810373715dc07e25f6e24f80 

- deconvolved signals are all zeros, and ones in correspondence of spikes
- neuropil is just random (for now)

alejoe91:
@bendichter merging this as well so we can get started with roundtrip tests. We can refine it later too :)

PR #44, opened 2020-09-01
PR1: moving methods to base class

Saksham20:
moving methods for get_traces, get_image,pixel_masks, num_of_frames/rois, get_roi_locations to base class

### DIFF
--- a/roiextractors/segmentationextractor.py
+++ b/roiextractors/segmentationextractor.py
@@ -3,6 +3,9 @@ from spikeextractors.baseextractor import BaseExtractor
 import numpy as np
 from .extraction_tools import ArrayType
 from .extraction_tools import _pixel_mask_extractor
+from copy import deepcopy
+import warnings
+
 
 class SegmentationExtractor(ABC, BaseExtractor):
     """
@@ -126,7 +129,7 @@ class SegmentationExtractor(ABC, BaseExtractor):
         num_of_frames: int
             Number of frames in the recording (duration of recording).
         """
-        pass
+        return self._roi_response.shape[1]
 
     @abstractmethod
     def get_roi_locations(self, roi_ids=None) -> np.array:
@@ -144,7 +147,13 @@ class SegmentationExtractor(ABC, BaseExtractor):
         roi_locs: numpy.ndarray
             2-D array: 2 X no_ROIs. The pixel ids (x,y) where the centroid of the ROI is.
         """
-        pass
+        if roi_ids is None:
+            return self.roi_locations
+        else:
+            roi_idx = [np.where(np.array(i) == self.roi_ids)[0] for i in roi_ids]
+            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
+            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
+            return self.roi_locations[:, roi_idx_]
 
     @abstractmethod
     def get_roi_ids(self) -> list:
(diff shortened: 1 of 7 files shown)


roiextractors/segmentationextractor.py
```diff
@@ -204,6 +221,27 @@ def get_image_size(self) -> ArrayType:
...
+        if roi_ids is None:
+            roi_idx_ = range(self.get_num_rois())
+        else:
+            roi_idx = [np.where(np.array(i) == self.get_roi_ids())[0] for i in roi_ids]
+            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
+            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
+        traces = self._roi_response_dict.get(name)
+        if len(traces.shape) == 0:
+            print(f'traces for {name} not found, enter one of {list(self._roi_response_dict.keys())}')
+            return None
```
bendichter, line 241: This should throw an error

roiextractors/segmentationextractor.py
```diff
@@ -3,6 +3,9 @@
 import numpy as np
 from .extraction_tools import ArrayType
 from .extraction_tools import _pixel_mask_extractor
+from copy import deepcopy
+import warnings
```
bendichter, line 7: these imports appear to be unused
```suggestion
```

Saksham20: yes, they are unused in this PR. Fixed https://github.com/catalystneuro/roiextractors/pull/44/commits/4562f963d1d9d9aabd073c2ea2acba9d61eb594b

roiextractors/segmentationextractor.py
Saksham20: fixed error where it calls shape on `None`

roiextractors/segmentationextractor.py
```diff
@@ -204,6 +215,27 @@ def get_image_size(self) -> ArrayType:
...
         pass
 
+    def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name='Fluorescence'):
+        """
+        Return RoiResponseSeries
+        Returns
+        -------
+        traces: array_like
+            2-D array (ROI x timepoints)
+        """
```
bendichter, line 225:
```suggestion
        """
        if name not in self._roi_response_dict:
            raise ValueError('traces for {name} not found , enter one of {list(self._roi_response_dict.keys())}')
```

roiextractors/segmentationextractor.py
```diff
@@ -204,6 +215,27 @@ def get_image_size(self) -> ArrayType:
...
+        else:
+            roi_idx = [np.where(np.array(i) == self.get_roi_ids())[0] for i in roi_ids]
+            ele = [i for i, j in enumerate(roi_idx) if j.size == 0]
+            roi_idx_ = [j[0] for i, j in enumerate(roi_idx) if i not in ele]
+        traces = self._roi_response_dict.get(name)
+        if traces is None:
+            print(f'traces for {name} not found, enter one of {list(self._roi_response_dict.keys())}')
+            return None
+        else:
+            return np.array([traces[int(i), start_frame:end_frame] for i in roi_idx_])
```
bendichter, line 237:
```suggestion
        traces = self._roi_response_dict.get(name)
        return np.array([traces[int(i), start_frame:end_frame] for i in roi_idx_])
```

roiextractors/segmentationextractor.py
```diff
@@ -222,7 +252,7 @@ def get_num_rois(self):
         no_rois: int
             integer number of ROIs extracted.
         """
-        return len(self.get_roi_ids())
+        return self._roi_response.shape[0]
```
alejoe91, line 255: @Saksham20 IMO `len(self.get_roi_ids())` is better. Imagine we make a tool to manually select some ROIs from an image: we could then instantiate the `NumpySegmentationExtractor`, but there would be no roi responses (yet). Instead `self._get_roi_ids()` should always be implemented.

PR #45, opened 2020-09-02
PR2: attributes rename

Saksham20:
There are a fair bit of change in this PR, will it be better to break it into 2 parts? 

Here are the changes: 
1. Images attributes: `self._image_raw`, `self._image_correlation`, these are subsequently packed into a dict using `self.get_images_dict()`
2. Traces attributes: `self._roi_response_raw`, `self._roi_response_dff`,  `self._roi_response_neuropil,` `self._roi_response_deconvolved`, these are then packed into dict using `self.get_traces_dict()`
3. all the property methods are removed from segmentationextractor base class. 
4. Renaming for more clarity: `filepath` > `file_path,` `no_of_channels` to `num_of_channels`

### DIFF
--- a/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -83,7 +83,7 @@ class NwbImagingExtractor(ImagingExtractor):
     mode = 'file'
     installation_mesg = "To use the Nwb Extractor run:\n\n pip install pynwb\n\n"  # error message when not installed
 
-    def __init__(self, filepath, optical_channel_name=None,
+    def __init__(self, file_path, optical_channel_name=None,
                  imaging_plane_name=None, image_series_name=None,
                  processing_module_name=None,
                  neuron_roi_response_series_name=None,
@@ -91,7 +91,7 @@ class NwbImagingExtractor(ImagingExtractor):
         """
         Parameters
         ----------
-        filepath: str
+        file_path: str
             The location of the folder containing dataset.nwb file.
         optical_channel_name: str(optional)
             optical channel to extract data from
@@ -163,26 +163,25 @@ class NwbSegmentationExtractor(SegmentationExtractor):
     mode = 'file'
     installation_mesg = ""  # error message when not installed
 
-    def __init__(self, filepath):
+    def __init__(self, file_path):
         """
         Creating NwbSegmentationExtractor object from nwb file
         Parameters
         ----------
-        filepath: str
+        file_path: str
             .nwb file location
         """
         check_nwb_install()
         SegmentationExtractor.__init__(self)
-        if not os.path.exists(filepath):
+        if not os.path.exists(file_path):
             raise Exception('file does not exist')
 
-        self.filepath = filepath
+        self.file_path = file_path
         self.image_masks = None
-        self.pixel_masks = None
         self._roi_locs = None
         self._accepted_list = None
-        io = NWBHDF5IO(filepath, mode='r+')
-        nwbfile = io.read()
+        self._io = NWBHDF5IO(file_path, mode='r+')
+        nwbfile = self._io.read()
         self.nwbfile = nwbfile
         _nwbchildren_type = [type(i).__name__ for i in nwbfile.all_children()]
         _nwbchildren_name = [i.name for i in nwbfile.all_children()]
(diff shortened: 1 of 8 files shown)


roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -252,16 +253,21 @@ def __init__(self, filepath):
             self.property_vals.append(np.array(ps[i].data))
 
         #Extracting stores images as GrayscaleImages:
-        self._greyscaleimages = [_nwbchildren_name[f] for f, u in enumerate(_nwbchildren_type) if u == 'GrayscaleImage']
+        _greyscaleimages = [i for i in nwbfile.all_children() if type(i).__name__ == 'GrayscaleImage']
+        self._image_correlation = [i.data[()] for i in _greyscaleimages if 'corr' in i.name.lower()][0]
+        self._image_mean = [i.data[()] for i in _greyscaleimages if 'mean' in i.name.lower()][0]
```
bendichter, line 258: I feel like there is probably a better way to do this, but we can address this later

roiextractors/extractors/schnitzerextractor/cnmfesegmentationextractor.py
```diff
@@ -55,23 +52,23 @@ def _tot_exptime_extractor_read(self):
         return self._dataset_file[self._group0[0]]['time']['totalTime'][0][0]
 
     def _summary_image_read(self):
-        summary_images_ = self._dataset_file[self._group0[0]]['Cn']
-        return np.array(summary_images_).T
+        summary_image = self._dataset_file[self._group0[0]]['Cn']
+        return np.array(summary_image).T
```
bendichter, line 56: shouldn't this be the correlation image?

roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -53,6 +51,17 @@ def get_accepted_list(self):
...
+    def _summary_image_read(self, bstr='meanImg'):
+        img = None
+        if bstr in self.ops:
+            if bstr == 'Vcorr' or bstr == 'max_proj':
+                img = np.zeros((self.ops['Ly'], self.ops['Lx']), np.float32)
+                img[self.ops['yrange'][0]:self.ops['yrange'][-1],
+                self.ops['xrange'][0]:self.ops['xrange'][-1]] = self.ops[bstr]
+            else:
+                img = self.ops[bstr]
+        return img
```
bendichter, line 63: I don't really understand what's going on here

roiextractors/segmentationextractor.py
```diff
@@ -215,25 +154,66 @@ def get_image_size(self) -> ArrayType:
...
+    def get_traces_dict(self):
+        """
+        Returns traces as a dictionary with key as the name of the ROiResponseSeries
+        Returns
+        -------
+        _roi_response_dict: dict
+            dictionary with key, values representing different types of RoiResponseSeries
+            Flourescence, Neuropil, Deconvolved, Background etc
+        """
+        return deepcopy(dict(raw=np.array(self._roi_response_raw),
```
bendichter, line 185: I don't think you need a `deepcopy` here

roiextractors/segmentationextractor.py
```diff
@@ -215,25 +154,66 @@ def get_image_size(self) -> ArrayType:
...
+    def get_images_dict(self):
+        """
+        Returns traces as a dictionary with key as the name of the ROiResponseSeries
+        Returns
+        -------
+        _roi_response_dict: dict
+            dictionary with key, values representing different types of Images used in segmentation:
+            Mean, Correlation image
+        """
+        return deepcopy(dict(mean=self._image_mean,
```
bendichter, line 199: I don't think you need a `deepcopy` here

roiextractors/segmentationextractor.py
```diff
@@ -215,25 +154,66 @@ def get_image_size(self) -> ArrayType:
...
+                             correlation=self._image_correlation))
+
+    def get_images(self, name='correlation'):
```
bendichter, line 202: shouldn't this be `get_image`?

roiextractors/extractors/schnitzerextractor/cnmfesegmentationextractor.py
Saksham20: i'm calling this summary image in all classes, but its assigned to `self._image_correlation` in the `__init__`

roiextractors/extractors/suite2p/suite2psegmentationextractor.py
bendichter: are you trying to change the data type? Or flip the image?

roiextractors/segmentationextractor.py
```diff
@@ -3,6 +3,8 @@
 import numpy as np
 from .extraction_tools import ArrayType
 from .extraction_tools import _pixel_mask_extractor
+from copy import deepcopy
+import warnings
```
bendichter, line 7:
```suggestion
```

roiextractors/extractors/suite2p/suite2psegmentationextractor.py
Saksham20: this is how its implemented in the suite2p nwbwrite method: they have selected a subframe within the self.ops['Vcorr'] image to assign to the correlation image. This subset is defined by the keys 'xrange' and 'yrange' in self.ops. 
This is their implementation: https://github.com/MouseLand/suite2p/blob/main/suite2p/io/nwb.py#L272

roiextractors/segmentationextractor.py
Saksham20: it does seem to make a difference: i tested it by manually changing the returned image_array (`get_images_dict()['mean']`) (without deepcopy) and it also changed the internal attribute (`self._image_mean`). With deepcopy this override does not alter the internal attribute.

roiextractors/segmentationextractor.py
Saksham20: https://github.com/catalystneuro/roiextractors/pull/45/commits/f2ae79b86c9060c772720451349a96ff9eb6b0ba

roiextractors/segmentationextractor.py
Saksham20: yes thanks for catching that, changed https://github.com/catalystneuro/roiextractors/pull/45/commits/f2ae79b86c9060c772720451349a96ff9eb6b0ba

roiextractors/extractors/suite2p/suite2psegmentationextractor.py
bendichter: oh, I see. OK, I think they were storing images from multiple planes side-by-side and then pulling the pixels corresponding to the plane of interest

PR #48, opened 2020-09-10
Tests

Saksham20:
tests files/classes structured like spikeextractors


roiextractors/tests/test_extractors.py
```diff
@@ -0,0 +1,46 @@
+import h5py
+import numpy as np
+import os
+import sys
+import unittest
+from ..example_datasets.toy_example import toy_example
+
+
+class TestSegmentationExtractors(unittest.TestCase):
```
alejoe91, line 9:
```suggestion
class TestExtractors(unittest.TestCase):
```

These should also include ImagingExtractors, so we generate the toy example only once for both

roiextractors/extractorlist.py
```diff
@@ -6,4 +6,28 @@
...
+    Hdf5ImagingExtractor,
+    TiffImagingExtractor,
+    NwbImagingExtractor,
+]
+
+segmentation_extractor_dict = {segmentation_class.extractor_name: segmentation_class for segmentation_class in
+                            segmentation_extractor_full_list}
+
+imaging_extractor_dict = {imaging_class.extractor_name: imaging_class for imaging_class in
+                            imaging_extractor_full_list}
```
bendichter, line 33: It makes sense to have this but I don't see why we need it in list form and in dict form

PR #51, opened 2020-09-18
Implemented write_imaging methods for ImagingExtractors

alejoe91:
- Implemented/improved writers for:
    - `Hdf5ImagingExtractor`
    - `NpyImagingExtractor`
    - `TiffImagingExtractor`

- Fixed some bugs in `MemmapImagingExtractor`
- Fix to toy example if a spike occurs at the last frame (a spikeextractor error apparently)

Issues: 
- @bendichter somehow the `lazy_slice` from `lazy_ops` is failing with a: 
`IndexError: Indices must be either integers, iterators of integers, slice objects, or numpy boolean arrays`
even if all indeces are int ot arrays of int.. any ideas? I also tried slices with the same outcome.

- The tiff writing is not memory efficient (using the `get_video()`). The `tifffile.imsave()` method supports only numpy array as input, so in case we want/need chunking we have to implemet it ourself (maybe using `tifffile` to write the header and dump the video in binary format in chunks at the right place

Both these issues could be addressed properly later!

### DIFF
--- a/roiextractors/extractors/numpyextractors/numpyextractors.py
+++ b/roiextractors/extractors/numpyextractors/numpyextractors.py
@@ -81,9 +81,15 @@ class NumpyImagingExtractor(ImagingExtractor):
         return self._num_channels
 
     @staticmethod
-    def write_imaging(imaging, save_path):
+    def write_imaging(imaging, save_path, overwrite: bool = False):
         save_path = Path(save_path)
         assert save_path.suffix == '.npy', "'save_path' should havve a .npy extension"
+
+        if save_path.is_file():
+            if not overwrite:
+                raise FileExistsError("The specified path exists! Use overwrite=True to overwrite it.")
+            else:
+                save_path.unlink()
 
         np.save(save_path, imaging.get_video())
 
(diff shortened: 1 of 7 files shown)


bendichter:
> @bendichter somehow the lazy_slice from lazy_ops is failing with a:
> IndexError: Indices must be either integers, iterators of integers, slice objects, or numpy boolean arrays
> even if all indeces are int ot arrays of int.. any ideas? I also tried slices with the same outcome.
@alejoe91 can you raise an issue in [lazy_ops](https://github.com/catalystneuro/lazy_ops)?

roiextractors/extractors/numpyextractors/numpyextractors.py
```diff
@@ -81,10 +81,16 @@ def get_num_channels(self):
         return self._num_channels
 
     @staticmethod
-    def write_imaging(imaging, save_path):
+    def write_imaging(imaging, save_path, overwrite: bool = False):
         save_path = Path(save_path)
         assert save_path.suffix == '.npy', "'save_path' should havve a .npy extension"
```
bendichter, line 86:
```suggestion
        assert save_path.suffix == '.npy', "'save_path' should have a .npy extension"
```

bendichter:
> The tiff writing is not memory efficient (using the get_video()). The tifffile.imsave() method supports only numpy array as input, so in case we want/need chunking we have to implemet it ourself (maybe using tifffile to write the header and dump the video in binary format in chunks at the right place

It looks like tiffile does support an `append` feature to write large datasets iteratively

https://github.com/cgohlke/tifffile/blob/a766675ddedd64afe294e5ebc36b3ce863669640/tifffile/tifffile.py#L752

but I'd be fine with making this a separate issue

alejoe91:
@bendichter the `append` option of `tifffile` doesn't work as expected. Basically a new series is added for each `imsave` call. I opened an issue and we can fix it later (https://github.com/cgohlke/tifffile/issues/28). I also reverted the lazy_ops calls until the integer issue is fixed.

Let's merge this so we can start writing tests. We'll fix those two things later :)

PR #64, opened 2020-12-16
Add repo tests

bendichter:


Saksham20 (changes requested):
assertion method for segmentation extractors

Saksham20 (changes requested):
@bendichter could you check: utils(contains the assert equal methods) and test_conversions(which now have the assertions in them)

bendichter:
@Saksham20 utils.py needs to be renamed to testing.py

PR #74, opened 2021-01-27
Use DataChunkIterator

Saksham20:
fixes #71 
using DataChunkIterator for buffered reading of large image_series and image_masks data onto memory before compression.

Currently there are errors while creating the nwbfile:

1. For writing ImagingExtractor to nwb, the the size of two photon series data changes from 3d to 4d(4th dim being the number of chunks) : (200, 60, 80) to ( 10, 20, 60, 80)- [here 10 are the number of chunks](https://github.com/catalystneuro/roiextractors/blob/dataset_chunking/roiextractors/extractors/nwbextractors/nwbextractors.py#L301). 
2. For **S[egmentationExctractor to nwb](https://github.com/catalystneuro/roiextractors/blob/dataset_chunking/roiextractors/extractors/nwbextractors/nwbextractors.py#L745)**:  I get this error when writing to nwb:
``` 
File "c:\users\saksham\documents\nwb\roiextractors\roiextractors\extractors\nwbextractors\nwbextractors.py", line 800, in write_segmentation
    io.write(nwbfile)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\utils.py", line 580, in func_call
    return func(args[0], **pargs)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\backends\hdf5\h5tools.py", line 329, in write
    call_docval_func(super().write, kwargs)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\utils.py", line 423, in call_docval_func
    return func(*fargs, **fkwargs)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\utils.py", line 580, in func_call
    return func(args[0], **pargs)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\backends\io.py", line 49, in write
    f_builder = self.__manager.build(container, source=self.__source, root=True)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\utils.py", line 580, in func_call
    return func(args[0], **pargs)
  File "C:\Users\Saksham\Documents\NWB\roiextractors\venv\lib\site-packages\hdmf\build\manager.py", line 173, in build
    result = self.__type_map.build(container, self, source=source, spec_ext=spec_ext, export=export)
[...]

PR #76, opened 2021-02-05
Conversiontools compatibility

Saksham20:
This PR makes changes to nwbwriter so that it can handle NWBFile as input argument. Previously, it required a nwbfile_path. This makes it compatible with nwbconvertiontools


Saksham20:
this should be merge after https://github.com/catalystneuro/roiextractors/pull/74

Saksham20:
I'll merge this PR, there is just code rearrangement in nwbsegmentation writer to additionally accept a open NWBFile object and if not None, manipulate it without returning anything (what NWBConverter requires). io.close() is also explicitly called in case NWBFile is none, so there is a removal of indent. 
All these make the diff look large

PR #78, opened 2021-02-28
images reverse dim order

Saksham20:
reverse the dim order for mean/corr images and roi_image_masks

### DIFF
--- a/roiextractors/extraction_tools.py
+++ b/roiextractors/extraction_tools.py
@@ -82,7 +82,7 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
     image_mask = np.zeros(list(image_shape) + [len(_roi_ids)])
     for no, rois in enumerate(_roi_ids):
         for x, y, wt in pixel_mask[rois]:
-            image_mask[int(x), int(y), no] = wt
+            image_mask[int(y), int(x), no] = wt
     return image_mask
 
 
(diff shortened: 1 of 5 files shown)


Saksham20:
@bendichter on a general note, in pynwb's `TwoPhotonSeries` doc ([here](https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.TwoPhotonSeries)), and `Image` ([here](https://pynwb.readthedocs.io/en/stable/pynwb.base.html#pynwb.base.Image)) each frame of image is required to be of the order `(x,y)` so for an example frame, if `frame_one.shape=(50,100)` it **must be transposed** before writing into nwb right?

bendichter:
I don't think it matters much what we can x and what we call y as long as we are consistent. What does a frame from this fix look like?

Saksham20:
> I don't think it matters much what we can x and what we call y as long as we are consistent. What does a frame from this fix look like?

there is some inconsistency in nwbwidgets and roiextractors due to which the nwb files have the different order in case of TwoPhotonSeries vs MeanImages/image_masks.
 the image masks widget plot, the assumption is image_mask.shape=(x,y) ([here](https://github.com/NeurodataWithoutBorders/nwb-jupyter-widgets/blob/master/nwbwidgets/ophys.py#L319)), while the TwoPhotonSeries is assuming (y,x). 
Now, in case of Roiextractors, we were storing each image as (y,x) so these errors cancelled out. To be consistent, it would make more sense to change: 
1) Roiextractors' `write_imaging` to store TwoPhotonSeries as (x,y) and ( i can push that change also to this PR) 
2) Nwbwidgets' TwoPhotonSeriesWidget to invert each image again. (create a new PR)

PR #82, opened 2021-04-13
EXTRACT-PUBLIC .mat file support

Saksham20:
this PR is to support the file type for the output of the [EXTRACT-PUBLIC](https://github.com/schnitzer-lab/EXTRACT-public#frequently-asked-questions) method of Schnitzer lab.

### DIFF
--- a/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
+++ b/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
@@ -149,3 +149,80 @@ class ExtractSegmentationExtractor(SegmentationExtractor):
 
     def get_image_size(self):
         return self._image_masks.shape[0:2]
+
+
+class ExtractPublicSegmentationExtractor(SegmentationExtractor):
+    """
+    This class inherits from the SegmentationExtractor class, having all
+    its funtionality specifically applied to the dataset output from
+    the \'EXTRACT\' ROI segmentation method.
+    """
+
+    extractor_name = "ExtractPublicSegmentationExtractor"
+    installed = HAVE_H5PY  # check at class level if installed or not
+    is_writable = False
+    mode = "file"
+    installation_mesg = "To use extract install h5py: \n\n pip install h5py \n\n"  # error message when not installed
+
+    def __init__(self, file_path: PathType):
+        """
+        Parameters
+        ----------
+        file_path: str
+            The location of the folder containing dataset.mat file.
+        """
+        SegmentationExtractor.__init__(self)
+        self.file_path = file_path
+        self._dataset_file = self._file_extractor_read()
+        self._image_masks = self._image_mask_extractor_read()
+        extraced_traces = self._trace_extractor_read()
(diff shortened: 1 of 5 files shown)


roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
```diff
@@ -149,3 +149,80 @@ def get_roi_ids(self):
...
+    def __del__(self):
+        self._dataset_file.close()
+
+    def _file_extractor_read(self):
+        return h5py.File(self.file_path, "r")
+
+    def _image_mask_extractor_read(self):
+        return DatasetView(
+            self._dataset_file["spatial_weights"]
+        )
```
bendichter, line 197: Do we need a DatasetView if we aren't transposing or slicing?

roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
```diff
@@ -149,3 +149,80 @@ def get_roi_ids(self):
...
+        return [a for a in range(self.get_num_rois()) if a not in ac_set]
+
+    @staticmethod
+    def write_segmentation(
+            segmentation_object: SegmentationExtractor, save_path, overwrite=True
+    ):
+        raise NotImplementedError
+
+    def get_roi_ids(self):
+        return list(range(self.get_num_rois()))
```
bendichter, line 225: Maybe this should be in the `SegmentationExtractor` class as a default implementation?

CodyCBakerPhD:
@bendichter What should be done about this PR?

There is no example file for it: https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/segmentation_datasets/extract

and no PR history for adding it back when this was WIP.

I would be hesitant to review/merge this without tests.

bendichter:
It looks like the file is the output of the tutorials [here](https://github.com/schnitzer-lab/EXTRACT-public/tree/master/Learning%20materials/Lecture%20Tutorials/Tutorial%201%20-%20Starting%20code)

CodyCBakerPhD:
> It looks like the file is the output of the tutorials [here](https://github.com/schnitzer-lab/EXTRACT-public/tree/master/Learning%20materials/Lecture%20Tutorials/Tutorial%201%20-%20Starting%20code)

Nice, I'll get it added, then.

CodyCBakerPhD:
Actually, I'm re-examining the example data.

The 'example' .mat file from that tutorial is actually just a couple of MATLAB variables that are used in running the tutorial itself, I do not believe they actually contain the structured format of the output of the call to the EXTRACT library functions. Instead, these objects are transformations of the output (see https://github.com/schnitzer-lab/EXTRACT-public/blob/master/Learning%20materials/Lecture%20Tutorials/Tutorial%201%20-%20Starting%20code/example_tutorial.m#L38-L39 for one such object)

However, by running the tutorial myself, I can just save the `output` variable from the EXTACT method run on that example data and use that instead.

CodyCBakerPhD:
Replaced by #170

PR #83, opened 2021-04-15
check file ext base class

Saksham20:
check for file_ext (of the file_path argument) in all `SegmentationExtractor `classes

### DIFF
--- a/roiextractors/segmentationextractor.py
+++ b/roiextractors/segmentationextractor.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-
+from pathlib import Path
 import numpy as np
 from spikeextractors.baseextractor import BaseExtractor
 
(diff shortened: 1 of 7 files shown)


CodyCBakerPhD:
@bendichter Was this feature still desired?

bendichter:
As this is currently implemented, a file extension is required, but I don't think it necessarily should be. I could imagine a segmentation interface where the file extension is flexible. If we want to keep this I think we should make that argument optional.

roiextractors/segmentationextractor.py
```diff
@@ -19,8 +19,11 @@ class SegmentationExtractor(ABC, BaseExtractor):
 
     installed = True
     installation_mesg = ""
+    file_ext = ""
 
-    def __init__(self):
+    def __init__(self, file_path):
```
bendichter, line 24:
```suggestion
    def __init__(self, file_path=None):
```

roiextractors/segmentationextractor.py
```diff
@@ -19,8 +19,11 @@ class SegmentationExtractor(ABC, BaseExtractor):
...
+        self.file_path = Path(file_path)
+        assert self.file_path.suffix == self.file_ext, f"file type should be of type: {self.file_ext}"
```
bendichter, line 26:
```suggestion
    def __init__(self, file_path=None):
        self.file_path = Path(file_path)
        if file_path is not None and not self.file_path.suffix == self.file_ext:
        warnings.warn(f"file type should be of type: {self.file_ext}")
```

CodyCBakerPhD:
Either needs to be rebased or just re-written from latest master (latter probably is just as easy given how small this PR is)

PR #85, opened 2021-12-03
cast ImagingExtractor.get_image_size output to list

### LINKED ISSUE #84: output from ImagingExtractor.get_image_size leads to failing metadata validation in nwb_conversion_tools
output is a numpy array. Validation expects a python list

cechava:
fixes #84

ImagingExtractor.get_image_size is cast to list


cechava:
@CodyCBakerPhD let me know if this would be the way fix such an issue. I think the issue is general enough that it should be addressed here rather than `nwb-conversion-tools`

CodyCBakerPhD:
@cechava Could very well be, and if it works locally then I'd trust it.

However, using the CI is usually the better standard so what we _should_ do is replace `git+https://github.com/catalystneuro/roiextractors@image_size_fix` in the requirements for the branch running the new `ophys` tests to be completely sure this does the trick.

I do see that this comes from your own copy of the repo, though - I'd typically recommend using the catalystneuro source directly so other people on the team can contribute. I went ahead and made this branch with that source and altered the base for this PR so it can be tested in this fashion.

PR #86, opened 2021-12-03
Image size fix

### LINKED ISSUE #84: output from ImagingExtractor.get_image_size leads to failing metadata validation in nwb_conversion_tools

CodyCBakerPhD:
fixes #84

Alias of PR #85 pointing to master, but referenceable by `nwb-conversion-tools`

ImagingExtractor.get_image_size is cast to list


CodyCBakerPhD:
@cechava See my comment on the other PR for adding `git+https://github.com/catalystneuro/roiextractors@image_size_fix` to the main testing branch to verify fix; although, as I think about it, if this is what fixes it I would ask that you try to find the place in `nwb-conversion-tools` where this problematic function is called, and cast the class type there instead of on `roiextractors`. This kind of class simplification is only required within the JSON schemas for the `nwb-conversion-tools` repo and may actually interfere with other intended behavior (like array-indices) if we changed it at the core level here.

cechava:
@CodyCBakerPhD those are very good points. I will make a fix on the nwb-conversion-tools side

Issue #52, opened 2020-09-18
Implement irregular sampling


alejoe91:
@bendichter @Saksham20 I have given it some though and I think we should support irregular sampling.

A `set_timestamps` base method in the `ImagingExtractor` and a modification of `frame_to_time` and `time_to_frame` (as suggested by Ben here - https://github.com/catalystneuro/roiextractors/pull/40#discussion_r476728127) should do the trick. 

Especially for calcium imaging, sometimes there are even a few milliseconds of difference between frames..

alejoe91:
Fixed with https://github.com/catalystneuro/roiextractors/pull/88

PR #88, opened 2021-12-29
Added set_times to imaging and nwb writer

alejoe91:
Added the possibility to set timestamps to the `ImagingExtractor` object, which are also converted to NWB as `timestamps` of the `TwoPhotonSeries`.

Implements request https://github.com/catalystneuro/roiextractors/issues/52

### DIFF
--- a/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -4,6 +4,7 @@ from collections import abc
 from datetime import datetime
 from pathlib import Path
 from warnings import warn
+from typing import Union
 
 import numpy as np
 from lazy_ops import DatasetView
@@ -28,6 +29,7 @@ except ImportError:
     HAVE_NWB = False
 
 from ...extraction_tools import (
+    NumpyArray,
     PathType,
     FloatType,
     IntType,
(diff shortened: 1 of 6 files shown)


roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -236,11 +240,18 @@ def __init__(
...
+        if self._times is None:
+            return int((times - self._imaging_start_time) * self.get_sampling_frequency())
+        else:
+            return super().time_to_frame(times)
+            
+    def frame_to_time(self, frames: Union[IntType, NumpyArray]):
+        if self._times is None:
+            return float(frames / self.get_sampling_frequency() + self._imaging_start_time)
+        else:
+            return super().frame_to_time(frames)
```
bendichter, line 253: Wouldn't it make more sense to move this to the base class?

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -354,6 +365,23 @@ def data_generator(imaging):
...
+            else:
+                two_p_series_kwargs.update(
+                    timestamps=H5DataIO(
+                        imaging.frame_to_time(
+                            np.arange(imaging.get_num_frames())),
+                        compression="gzip"
+                    )
+                )
+                if "rate" in two_p_series_kwargs:
+                    del two_p_series_kwargs["rate"]
```
bendichter, line 383: Why add "rate" just to delete it?

roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -465,8 +494,11 @@ def write_imaging(
             metadata info for constructing the nwb file (optional).
         overwrite: bool
             If True and save_path is existing, it is overwritten
-        num_chunks: int
+        buffer_size: int
             Number of chunks for writing data to file
+        use_times: bool (optional, defaults to False)
+            If True, the times are saved to the nwb file using imaging.frame_to_time(). If False (defualut),
```
bendichter, line 500:
```suggestion
            If True, the times are saved to the nwb file using imaging.frame_to_time(). If False (defualt),
```

roiextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: No because `imaging_start_time` currently is only supported by NWB

roiextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: `rate` is added to the dictionary automatically when retrieving nwb metadata https://github.com/catalystneuro/roiextractors/blob/timestamps/roiextractors/extractors/nwbextractors/nwbextractors.py#L457-L460)

So in case `use_times` is used it's easier to just remove it IMO. If you prefer, we can avoid adding it in the `get_nwb_metadata` and take care of `rate` and/or `timestamps` directly in the `add_two_photon_series` function. What do you think?

alejoe91:
> @alejoe91 For more thoroughly testing this through the conversion tool interfaces, would it be OK if I copy over this version of the `write` to NWB functions to `nwb-conversion-tools`?

@CodyCBakerPhD can you take care of it?

CodyCBakerPhD:
> @CodyCBakerPhD can you take care of it?

Got the ball rolling but needs a little more work: https://github.com/catalystneuro/nwb-conversion-tools/pull/397

CodyCBakerPhD:
Test results over there: https://github.com/catalystneuro/nwb-conversion-tools/pull/397

Suggested fixes: https://github.com/catalystneuro/nwb-conversion-tools/pull/397#issuecomment-1035711980

alejoe91:
@CodyCBakerPhD Done! can you try to retrigger tests using this branch fron nwb-conv-tools?

PR #89, opened 2021-12-29
Added citation.cff for "cite" widget

alejoe91:


CITATION.cff
```diff
@@ -0,0 +1,15 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+  - family-names: Saksham 
+    given-names: Sharda
+  - family-names: Alessio Paolo 
+    given-names: Buccino
+    orcid: https://orcid.org/0000-0003-3661-527X
+  - family-names: Ben 
+    given-names: Dichter
```
bendichter, line 10: "family name" means last name and "given name" means first name. Also, for these types of things use Benjamin as my first name

bendichter (changes requested):
"family name" means last name and "given name" means first name. Also, for these types of things use Benjamin as my first name

alejoe91:
> "family name" means last name and "given name" means first name. Also, for these types of things use Benjamin as my first name

Oooops! I know but I did it quickly and got it worng. Fixing it now

bendichter (approved):

PR #90, opened 2022-02-02
Relax dependency versions to fix installation issues

alejoe91:
Try to fix installation issues in tests https://github.com/catalystneuro/roiextractors/issues/91


alejoe91:
@bendichter @Saksham20 can we merge this? I relaxed the dependencies in the requirement files, otherwise that forced very old versions of several packages

PR #92, opened 2022-02-11
Fix NumpySegmentationExtractor

alejoe91:
- Fixes a bug in instantiating a `NumpySegmentationExtractor` with image_masks. Basically, the `roi_ids` were automatically inferred using the wrong dimension of the masks, resulting in a mismatch between `roi_ids` and the number of masks.
- Propagate properly device description from metadata (@CodyCBakerPhD this second change should be integrated in https://github.com/catalystneuro/nwb-conversion-tools/pull/397)

### DIFF
--- a/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -318,10 +318,15 @@ class NwbImagingExtractor(ImagingExtractor):
     def add_devices(imaging, nwbfile, metadata):
         # Devices
         metadata = dict_recursive_update(get_default_nwb_metadata(), metadata)
+        
         # Tests if devices exist in nwbfile, if not create them from metadata
         for dev in metadata["Ophys"]["Device"]:
             if dev["name"] not in nwbfile.devices:
-                nwbfile.create_device(name=dev["name"])
+                if "description" in dev:
+                    descr = dev["description"]
+                else:
+                    descr = None
+                nwbfile.create_device(name=dev["name"], description=descr)
 
         return nwbfile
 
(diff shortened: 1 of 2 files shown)


roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -320,10 +320,15 @@ def get_num_channels(self):
...
+
         # Tests if devices exist in nwbfile, if not create them from metadata
         for dev in metadata["Ophys"]["Device"]:
             if dev["name"] not in nwbfile.devices:
-                nwbfile.create_device(name=dev["name"])
+                if "description" in dev:
+                    descr = dev["description"]
+                else:
+                    descr = None
+                nwbfile.create_device(name=dev["name"], description=descr)
```
CodyCBakerPhD, line 331: Instead of manually adding this, I've just generalized it to `nwbfile.create_device(**device)`

CodyCBakerPhD:
@alejoe91 How was this bug caught? Is it detectable through any tests either on synthetically generated data or GIN-hosted data? Would prefer to add some tests somewhere that directly reflect this if so

roiextractors/extractors/nwbextractors/nwbextractors.py
alejoe91: No, we were trying to instantiate a `NumpySegmentationExtractor` object with manually cut ROIs and when converting to NWB it threw an error. I agree we should have testing for the `Numpy` objects too

CodyCBakerPhD: What specifically is the error when writing to NWB? Is it fixed in https://github.com/catalystneuro/nwb-conversion-tools/pull/397?

alejoe91: The error was a mismatch between the roi_ids and the number of available masks. I'll test if fixed on the `nwb-conv-tools` side

CodyCBakerPhD (approved):

CodyCBakerPhD:
@alejoe91 Merge at will

Add tests to NWBCT if possible for this specific issue

PR #93, opened 2022-02-21
Remove NWB write functions; throw error with instructions

CodyCBakerPhD:
Removal of local NWB write function logic, but maintaining the function itself which throws the following error with instructions

image

I can set a notification 3 months from now to actually remove the methods themselves, if that timeline works for you all.

Exact message instructions may depend on the upcoming tool/import re-organization, which is why I'm setting this to draft for the moment.


CodyCBakerPhD:
To-Do before next release

- [ ] fix tests

CodyCBakerPhD:
@alejoe91 Ready for review

bendichter:
Looks good overall, but test names e.g. `test_convert_img_interface_to_nwb` need to change

bendichter (approved):

CodyCBakerPhD:
@bendichter Note the tests occasionally do test roundtrip write within their own format - I restored the roundtrip check for this now. So test suite is 'mostly' read, but 'some' write where applicable.

bendichter:
ok, sounds good

CodyCBakerPhD:
@bendichter Oh my, it actually seems that roundtrip checking of available non-NWB write methods fails. I've commented it out for now and I'll raise an issue for it.

PR #105, opened 2022-05-18
[DO NOT MERGE] conflicting CI

CodyCBakerPhD:
The current CI is failing due to inability to download test data on the fly.

This revamps the CI on `roiextractors` to use our established practice of data caching as well as decoupling the act of downloading the data through datalad from the actual testing of that data, which makes it easier to parse errors like we're seeing now.


CodyCBakerPhD:
@h-mayorquin If you could check my work here - this PR updates the CI to use the NWBCT convention for handling caching (tests now run in 3-4 minutes, and seamless between remote and local copies). This required that other stuff you recently updated such as `/src/` nesting, the GIN config setup, and 'extra' requirements.

Tests should now be much more readable than they used to be, and also reflect the `param` usage from NWBCT.

CodyCBakerPhD:
Reopenened on split prior to .toml induced conflicts: https://github.com/catalystneuro/roiextractors/pull/107

PR #106, opened 2022-05-18
Standardize black options

CodyCBakerPhD:
When I setup the pre-commit config, I did not add a `.toml` specification for black rules. This updates them from the default to the CatalystNeuro standard.


CodyCBakerPhD:
Wow, our custom choice of rules is good. +146/-387 (net -241 lines) just based on splitting.

PR #107, opened 2022-05-18
[Bug] Fix CI

CodyCBakerPhD:
@h-mayorquin If you could check my work here - this PR updates the CI to use the NWBCT convention for handling caching (tests now run in 3-4 minutes, and seamless between remote and local copies). This required that other stuff you recently updated such as /src/ nesting, the GIN config setup, and 'extra' requirements.

Tests should now be much more readable than they used to be, and also reflect the param usage from NWBCT.


CodyCBakerPhD:
@h-mayorquin Never mind, still getting format conflicts - I know how to fix it, just be a sec.

PR #108, opened 2022-05-18
[Bug] Fix CI

CodyCBakerPhD:
OK... this one shouldn't create any conflicts

@h-mayorquin If you could check my work here - this PR updates the CI to use the NWBCT convention for handling caching (tests now run in 3-4 minutes, and seamless between remote and local copies). This required that other stuff you recently updated such as /src/ nesting, the GIN config setup, and 'extra' requirements.

Tests should now be much more readable than they used to be, and also reflect the param usage from NWBCT.


.github/workflows/testing.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Get ophys_testing_data current head hash
+        id: ophys
+        run: echo "::set-output name=HASH_OPHYS_DATASET::$(git ls-remote https://gin.g-node.org/CatalystNeuro/ophys_testing_data.git HEAD | cut -f1)"
+      - name: Cache ophys dataset - ${{ steps.ophys.outputs.HASH_OPHYS_DATASET }}
+        uses: actions/cache@v2
+        id: cache-ophys-datasets
+        with:
+          path: ./ophys_testing_data
+          key: ophys-datasets-051822-${{ matrix.os }}-${{ steps.ophys.outputs.HASH_OPHYS_DATASET }}
```
h-mayorquin, line 58: You know, when I saw the date here for the first time I could not tell it was a date because of the format : )

setup.py
```diff
@@ -1,31 +1,46 @@
...
+from shutil import copy as copy_file
 
-d = {}
-exec(open("roiextractors/version.py").read(), None, d)
-version = d["version"]
-pkg_name = "roiextractors"
-with open("README.md") as rd:
-    long_description = rd.read()
+
+path = os.path.abspath(os.path.dirname(__file__))
```
h-mayorquin, line 8: I have always wanted to do all of this with `pathlib` instead of `os`

setup.py
```diff
@@ -1,31 +1,46 @@
...
+    packages=find_packages(where="src"),
+    package_dir={"": "src"},
+    include_package_data=True,  # Includes files described in MANIFEST.in in the installation.
+    python_requires=">=3.7",
+    install_requires=install_requires,
+    extras_require=extras_require,
     classifiers=(
         "Programming Language :: Python :: 3",
-        "License :: OSI Approved :: Apache Software License",
+        "License :: OSI Approved :: BSD-3 Software License",
```
h-mayorquin, line 43: relevant now!

tests/test_io.py
```diff
@@ -1,140 +1,127 @@
...
+
+            roundtrip_kwargs = copy(extractor_kwargs)
+            roundtrip_kwargs.update(file_path=output_path)
+            roundtrip_extractor = extractor_class(**roundtrip_kwargs)
+            # TODO: this roundtrip test has been failing for some time now
+            # check_imaging_equal(img1=extractor, img2=roundtrip_extractor)
         except NotImplementedError:
             return
 
-    @parameterized.expand(
```
h-mayorquin, line 93: So good that we got rid of this.

h-mayorquin (approved):
The test are now passing. My comments are comments rather than suggestions that I wrote as I read. 

Thanks for getting the repo ready now that I am going to work on it : )

.github/workflows/testing.yml
CodyCBakerPhD: lol that's just my personal way of associating a manual hash to it - it doesn't *have* to be a date per se. In fact I used to just use a running integer count, but I figured a date would give more contextual detail

setup.py
CodyCBakerPhD: Shouldn't be hard, lemme mess with that rq

setup.py
CodyCBakerPhD: Yeah IDK where that Apache bit came from, must have been auto-populated or something

setup.py
CodyCBakerPhD: I'll do it in a follow-up, though. Need to get this through ASAP

PR #109, opened 2022-05-18
[Refactor] setup.py to use pathlib

CodyCBakerPhD:
@h-mayorquin

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,19 +1,18 @@
-import os
 from pathlib import Path
 from setuptools import setup, find_packages
 from copy import copy
 from shutil import copy as copy_file
 
 
-path = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Hmm.. odd, the cache is missing after merge of previous PR... might need to instantiate the cache directory on main. Ah well, nightly test run should take care of it

h-mayorquin (approved):
Amazing!
Maybe root is more informative than parent? 

Never seen the this patch bot before. Nice.

CodyCBakerPhD:
> Never seen the this patch bot before. Nice.

Ahh that is one of the default options of codecov reports - it's honestly not a great classification of what constitutes a patch vs. project so we generally disable it. Geez, I didn't realize the repo was also missing codecov specification. Adding that in follow-up.

Issue #111, opened 2022-05-19
[Refactor] MemmapImagingExtractor


CodyCBakerPhD:
Just noticed this extractor expose at the outer level: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/memmapextractors.py

(i) the `__init__` method seems to be structured more after the intention of the static `write_imaging` methods per class, not one that 'reads' data from a binary blob. Let's refactor this to be a proper ImagingExtractor so we can use it on the Seidemann project.

(ii) Also, let's add it to the main listing: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractorlist.py

(iii) Also, let's add some tests for it. Roundtrip preferably.

(iv) Propagate all this up to NWBCT DataInterfaces and tests

PR #112, opened 2022-05-19
Refactor memap

h-mayorquin:
Related to #111. 

This is the first attempt at refactoring `MemmapImagingExtractor`.

I added a new sets of tests. Let's discuss the scope and how to organize them.
There is some functionality that is not implemented yet (channel names for example).

Maybe we should discuss what is necessary for the current conversion and do the rest of things after we are done with it.


CodyCBakerPhD:
@h-mayorquin Can you move the import from it's current location to follow the structure of all the other extractors? I.e., `/extractors/memmapextractor/...`

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
 import numpy as np
 from tqdm import tqdm
 
 from .extraction_tools import PathType, check_get_frames_args
 from .imagingextractor import ImagingExtractor
 
+PathType = Union[str, Path]
+DtypeType = Union[str, np.dtype]
+OptionalDtypeType = Optional[DtypeType]
+DtypeType = Union[str, np.dtype]
```
CodyCBakerPhD, line 16: This last one supposed to be Optional? Otherwise duplicate of line 14

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
+
+    extractor_name = "MemmapImagingExtractor"
+
     def __init__(
         self,
-        imaging_extractor: ImagingExtractor,
-        save_path: PathType = None,
-        verbose: bool = False,
+        file_path: PathType,
+        frame_shape: tuple,
```
CodyCBakerPhD, line 26: For extra clarity, could do `Tuple[int, int]` which specifies both form of contents and size of tuple

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
+        dtype: DtypeType,
+        offset: int = 0,
+        sampling_frequency: float = 0,
```
CodyCBakerPhD, line 29: I'd make sampling frequency required, since there is no global proprietary header for this format

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
+            Note that n_channels is 1 for grayscale and 3 for color images.
+        dtype : DtypeType
+            The type of the data to be loaded (int, float, etc.)
+        offset : int, optional
+            The offset in bytes. Usually corresponds to the number of bytes occupied by the header. 0 by default.
+        sampling_frequency : float, optional
+            The sampling frequency.
+        image_structure_to_axis : dict, optional
+            A dictionary indicating what axis corresponds to what in the memmap. The default values are:
+            dict(frame_axis=0, num_channels=1, rows=2, columns=3)
```
CodyCBakerPhD, line 51: Are 'rows' and 'columns' the best verbal way to communicate the (x,y) dimensions of an image? Must think on this...

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
         self,
-        imaging_extractor: ImagingExtractor,
-        save_path: PathType = None,
-        verbose: bool = False,
+        file_path: PathType,
+        frame_shape: tuple,
+        dtype: DtypeType,
+        offset: int = 0,
+        sampling_frequency: float = 0,
+        image_structure_to_axis: dict = None,
```
CodyCBakerPhD, line 30: Similarly, can be `Dict[str, int]` to better indicate expected format of specification

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
-                self.imaging.get_image_size()[0],
-            ),
-            dtype=self.imaging.get_dtype(),
-            mode="r",
-        )
-
-    @check_get_frames_args
-    def get_frames(self, frame_idxs, channel=0):
-        return self._video[channel, frame_idxs]
+        """Class for reading binary data.
```
CodyCBakerPhD, line 32: Specifically, optical imaging data stored in a binary format

src/roiextractors/memmapextractors.py
h-mayorquin: Yes, some libraries have (x, y) but that's ambiguous as sometimes this intuition crashes with our picture of the Cartesian plane. Some areas have x means rows and others have y means rows:

```
 ^
 |
 |
 Y
 |
 |
0/0---X--->
````

File: src/roiextractors/memmapextractors.py (hunk 3 above)
Comment (h-mayorquin) [reply src/roiextractors/memmapextractors.py:29]: At this level I think we could support the ability to extract images even if the ability to map this to times is not available.

File: src/roiextractors/memmapextractors.py (hunk 2 above)
Comment (h-mayorquin) [reply src/roiextractors/memmapextractors.py:26]: Right!

File: src/roiextractors/memmapextractors.py (hunk 1 above)
Comment (h-mayorquin) [reply src/roiextractors/memmapextractors.py:16]: No, good catch!

File: src/roiextractors/memmapextractors.py (hunk 5 above)
Comment (h-mayorquin) [reply src/roiextractors/memmapextractors.py:30]: Thanks!

File: src/roiextractors/memmapextractors.py (hunk 6 above)
Comment (h-mayorquin) [reply src/roiextractors/memmapextractors.py:32]: Thanks!

File: src/roiextractors/memmapextractors.py [hunk 7]
```diff
@@ -2,109 +2,118 @@
...
+            The sampling frequency.
+        image_structure_to_axis : dict, optional
+            A dictionary indicating what axis corresponds to what in the memmap. The default values are:
+            dict(frame_axis=0, num_channels=1, rows=2, columns=3)
+            frame_axis=0 indicates that the first axis corresponds to the frames (usually time)
+            num_channels=1 here indicates that the first axis corresponds to the  n_channels.
+            rows=2 indicates that the rows in the image are in the second axis
+            columns=3 indicates that columns is the last axis or dimension in this structure.
+
+            Notice that this should correspond with frame_shape.
```
CodyCBakerPhD, line 57: This argument is an interesting proposition. But I wonder if there aren't better ways of combining this intention with the `frame_shape`, such as combininng the two in something like an `OrderedDictionary` where the order communicates the intended order of the axis?

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
-            save_path = Path(save_path)
-            if save_path.suffix != ".dat" and save_path.suffix != ".bin":
-                save_path = save_path.with_suffix(".dat")
-            if not save_path.parent.is_dir():
-                os.makedirs(save_path.parent)
-            self._is_tmp = False
-            self._tmp_file = save_path
-
-        self._save_memmap_video(verbose)
-        self._verbose = verbose
```
CodyCBakerPhD, line 37: This doesn't have to be outright removed - it can serve as the @staticmethod `write_imaging` for the class, which is what it seems to do as-is.

src/roiextractors/memmapextractors.py
```diff
@@ -2,109 +2,118 @@
...
-            save_path,
-            shape=(
-                self.imaging.get_num_channels(),
-                self.imaging.get_num_frames(),
-                self.imaging.get_image_size()[0],
-                self.imaging.get_image_size()[0],
-            ),
-            dtype=self.imaging.get_dtype(),
-            mode="r",
-        )
```
CodyCBakerPhD, line 94: However, I have no clue why this is here. Probably can be safely deleted.

src/roiextractors/memmapextractors.py
h-mayorquin: I don't think it works as it is.

src/roiextractors/memmapextractors.py
CodyCBakerPhD: Dangerous territory; frame rate had to be present to allow acquisition, and core methods such as `frame_to_time` and vice versa won't be possible without it. I thought SpikeInterface and the other extractors here treat `sampling_frequency` generally as a requirement (unless, of course, it is automatically set from a header)?

src/roiextractors/memmapextractors.py
CodyCBakerPhD: Wow, https://github.com/staticmethod is an actual GitHub user, lol

CodyCBakerPhD:
@h-mayorquin Something we could do to break this up/ make several simple PR's...

i) Could make this PR about moving the extractor class into the proper folder/import/listing setup

and

ii) move the `__init__` to the `write_imaging`, remove unused or irrelevant methods like `move_to`

iii) define the `__init__` of course, but just as a `pass` or `raise NotImplementedError`

iv) Now, strictly speaking, what is usually done for this type of interface is to ensure every other imaging extractor can be written using this `write_imaging` method. I can see two ways of doing this; either make that `read_binary` help function an extraction tool like it is on spikeextractors/SpikeInterface, or just route everything through the `MemmapImagingExtractor.write_imaging` static method, your call.

until in a follow-up PR,

a) define the init method as you have

and

b) add a data read test (probably don't need GIN data for this one, though)

src/roiextractors/memmapextractors.py
CodyCBakerPhD: > I don't think it works as it is.

The `_save_memmap_video` doesn't work when `save_path` isn't `None`?

src/roiextractors/memmapextractors.py
h-mayorquin: All of those methods themselves will fail for lack of a sampling frequency. What is the danger that you see?

src/roiextractors/memmapextractors.py
h-mayorquin: It could be. I thought about like an hour. Then decided that implementation as I wanted a working prototype. I am hoping something better comes to my head. My naive implementation in that direction got to a point where I needed the string version of the variable name and I always stop when I go there. 

Open to suggestions.

src/roiextractors/memmapextractors.py
h-mayorquin: x D

h-mayorquin:
@CodyCBakerPhD 
I am confused about ii)
I don't see that structure in the `TiffImagingExtractor`, is this somewhere else so I can see it?

For iv) I am going to look how and where this is done in `spikeinterface` (pointers welcome!) I am concerned that doing this requires information about all the binaries of all the other formats (such as orientation and format), do you find that concern reasonable?

src/roiextractors/memmapextractors.py
CodyCBakerPhD: > All of those methods themselves will fail for lack of a sampling frequency. What is the danger that you see?

Why accept a value, or default a value for that matter, that will fail further down the line? Also an NWB requirement.


Elsewhere in the ecosystem...

https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/baserecording.py#L25

Required.

https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/basesorting.py#L13

Required.


Implication being that
https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L24 should also be required.

src/roiextractors/memmapextractors.py
CodyCBakerPhD: How about a single argument that is a `Union[Tuple[int, int, int], Tuple[int, int, int]]` that is used to map the axes from the memmap to an roiextractors-enforced standard (which would make array shaping easier for things like `get_frames` and cross-extractor usage) of (frames, rows, columns, channels).

Also, I kind of like `channels` to be on the outer level here so reflect the common ordering of such axes for movie data (which is another extractor to be added soon).

h-mayorquin: `get_frames` can be done in one line with the function `np.take` as long as we know which one is the frame_axis. 

That said, I am sympathetic to the idea to have a common inner standard in which the data should be forced for internal operations. 

Can you elaborate on how are you thinking though?
Let's say the example from the test:
```python
        num_frames = 50
        num_channels = 3
        rows = 10
        columns = 10
        memmap_shape = (num_channels, rows, columns, num_frames)
```

How are your tuples in this case?
Let's say that the reference is `(num_frames, num_channels, rows, columns)`.

CodyCBakerPhD:
> I am confused about ii)
I don't see that structure in the TiffImagingExtractor, is this somewhere else so I can see it?

Defined in base class: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L135-L149

And nor will it be for any extractor where it's too difficult to do so, or even impossible because their code doesn't allow it (more an issue for certain RecordingExtractors that have closed-source code).

But for something like this extractor, it should be allowed since `np.memmap` is so basic.


> For iv) I am going to look how and where this is done in spikeinterface (pointers welcome!) I am concerned that doing this requires information about all the binaries of all the other formats (such as orientation and format),

The use of such a static method, I believe, was more of a thing for the legacy spikeextractors project, so might help more to look there. This project started off following that structure, but as time goes on we are incorporating more changes to mimic the newer SpikeInterface, so this could be a decision point between which structure we want to follow.

In SpikeInterface, I think they use `tool`-like helper functions like `read_binary` and `write_binary` to do all that stuff.


> I am concerned that doing this requires information about all the binaries of all the other formats (such as orientation and format), do you find that concern reasonable?

This is the entire goal of creating a unified API 😉 You shouldn't need to know *anything* about the underlying storage/access of each other interface, simply assume they have a defined method for the API call `get_frame(...)`and use that to fill the block we want to allocate on disk.

src/roiextractors/memmapextractors.py
h-mayorquin: The use case that i mentioned is that you want to extract the images even if you don't know the times associated to each of the frames.  As long as they are using it to check the images, perform operations on them and do anything that does not require associating frames with times, they will be fine. Most of the imaging processing that I know does not require knowing the times.

But in our ecosystem you know the use cases better than me. For `nwb-conversion-tools` for example we need this for synchronization with behavior and other measures so over there it should be required.

If you think that we should prohibit the image processing use case at this level I am fine with that.

h-mayorquin:
> > I am confused about ii)
> > I don't see that structure in the TiffImagingExtractor, is this somewhere else so I can see it?
> 
> Defined in base class: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L135-L149

> But for something like this extractor, it should be allowed since `np.memmap` is so basic.
> 
I meant the thing about moving the `__init__` to `write_imaging`. I am aware that classes are supposed to have a `write_method`.

> > I am concerned that doing this requires information about all the binaries of all the other formats (such as orientation and format), do you find that concern reasonable?

> This is the entire goal of creating a unified API wink You shouldn't need to know _anything_ about the underlying storage/access of each other interface, simply assume they have a defined method for the API call `get_frame(...)`and use that to fill the block we want to allocate on disk.

Here I misunderstood you above. I was thinking about `reading`. Writing, specially for this format, should be easy as you say and makes my concerns a moot point.

src/roiextractors/memmapextractors.py
CodyCBakerPhD: >  memmap_shape = (num_channels, rows, columns, num_frames)

The tuple I'm imaging, in this case, would be `(3, 1, 2, 0)` -> `(frames, rows, columns, channels)` since the rows/columns axes are already in the right places, but the channels/frames are reversed.

Another option to allow textual specification without the change for key errors that a standard dictionary would have for this kind of thing, could be either a `namedtuple` or `dataclass` (`dataclass` being the more modern one, I'd say). Check out both those standard (but more exotic) Python data types if you haven't seen them before.

Specifically, the `dataclass `(here, named 'AxisOrder' for example) approach would look, as an input, like `AxisOrder(frames=3, rows=1, columns=2, channels=0)`, which is close to the spirit of the `dict` version you proposed but has the critical difference that each key is forced to be one of those 4 choices (with, perhaps, `channels` being optional) whereas a normal `dict` can contain any string (well, actually, any hashable Python data type). This, and the `Enum` data type, are the approach we took for this type of stuff in the NWBInspector and it really does help prevent silly things like typos from going any further than they need to. Also they aren't mutable, so that's a plus over the `dict` approach, too.

src/roiextractors/memmapextractors.py
CodyCBakerPhD: > The use case that i mentioned is that you want to extract the images even if you don't know the times associated to each of the frames. As long as they are using it to check the images, perform operations on them and do anything that does not require associating frames with times, they will be fine. Most of the imaging processing that I know does not require knowing the times.

Then for what you want here, the best case would be to have the `read_binary` method be a common helper function in `extractor_tools.py` or something like that. That way, if people want to do that, they can use the function to read the binary blob and do what they want.

But anything at the extractors level of operations (and that goes for SpikeInterface, too), requires a notion of temporal scale in order to visualize and package the data in a way that makes sense for analysis or further processing like segmentation.

CodyCBakerPhD:
> I meant the thing about moving the __init__ to write_imaging. I am aware that classes are supposed to have a write_method.

Yeah, because this newly rediscovered extractor clearly had some big differences from the way it ought to be to fit in with the rest of the package. Probably the strangest-designed extractor I've ever seen (even compared to Axona, lol)

src/roiextractors/memmapextractors.py
h-mayorquin: > > memmap_shape = (num_channels, rows, columns, num_frames)
> 
> The tuple I'm imaging, in this case, would be `(3, 1, 2, 0)` -> `(frames, rows, columns, channels)` since the rows/columns axes are already in the right places, but the channels/frames are reversed: 

So you have:
`Union[Tuple[int, int, int], Tuple[int, int, int]]`
There is a tuple on the left and one in the right.
Is the one in the left `(3, 1, 2, 0)` and the one in the right `(frames, rows, columns, channels)`?

> Another option to allow textual specification without the change for key errors that a standard dictionary would have for this kind of thing, could be either a `namedtuple` or `dataclass` (`dataclass` being the more modern one, I'd say). Check out both those standard (but more exotic) Python data types if you haven't seen them before.
> 
> Specifically, the `dataclass `(here, named 'AxisOrder' for example) approach would look, as an input, like `AxisOrder(frames=3, rows=1, columns=2, channels=0)`, which is close to the spirit of the `dict` version you proposed but has the critical difference that each key is forced to be one of those 4 choices (with, perhaps, `channels` being optional) whereas a normal `dict` can contain any string (well, actually, any hashable Python data type). This, and the `Enum` data type, are the approach we took for this type of stuff in the NWBInspector and it really does help prevent silly things like typos from going any further than they need to. Also they aren't mutable, so that's a plus over the `dict` approach, too.

Yes, I was explicitly thinking on using a data-class earlier today but my feeling is that it delegates too much definitions to the users. Tuples and dicts are easier to define and most people (specially scientist who are not well versed in OOD) understand them.  A data class simplifies our input and allows us to have checks on the spot but is a step further in complexity. I am on the fence about that trade-off.

The current solution is very simple as long as the format is on the "standard" (having the frames on the first axis) as it just demands structuring their data in a tuple. If their data is not in that format then and only then asks them to use a little bit more of complexity with a dict.

This is an interesting design problem.

h-mayorquin:
@CodyCBakerPhD 
Can we do what you suggested in the opposite order?
First we merge this as it is and then we add another PR were we add a proper `write_function`?
This as it is is what is needed for the conversion and I already did the work of re-organizing the class.

Also, by `data_read_test` you mean something different to what I have on the tests here?

src/roiextractors/memmapextractors.py
h-mayorquin: That's fine.

bendichter:
@h-mayorquin you can also use [lazy_ops transpose](https://github.com/catalystneuro/lazy_ops)

@CodyCBakerPhD and @h-mayorquin  Oliver mentioned that memmaps tend to cache accessed data into RAM, which may explain why we see an increase in ram as we iterate through binary data

src/roiextractors/extractors/memmapextractor/memmapextractors.py
```diff
@@ -0,0 +1,136 @@
...
+    PathType,
+    DtypeType,
+)
+
+
+class MemmapImagingExtractor(ImagingExtractor):
+
+    extractor_name = "MemmapImagingExtractor"
+
+    def __init__(
```
bendichter, line 23: why not just include the memmap here instead of the filename? Then this would work for HDF5 too

src/roiextractors/memmapextractors.py
h-mayorquin: I re-wrote the `write_imaging` method now.

src/roiextractors/extractors/memmapextractor/memmapextractors.py
CodyCBakerPhD: Basing design off of the SpikeInterface analog: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/binaryrecordingextractor.py

Also, part of the work here being done by the `__init__` is figuring out 'how' to read said memmap, which might be a tad much to ask first-time users.

Also also, we already have HDF5 Imaging handled: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py, unless you meant refactoring using a kind of base binary-like class for both use cases?

src/roiextractors/memmapextractors.py
CodyCBakerPhD: > but my feeling is that it delegates too much definitions to the users

If it helps, remember the user wouldn't define anything about the class - syntactically it would be just like specifying a dictionary, except it would immediately notify them if they forgot/failed to specify one of the keys, or if they mispelled a key. Basically, it's a dictionary with a hard-coded selection of keys they can associate values too.

src/roiextractors/extractors/memmapextractor/memmapextractors.py
CodyCBakerPhD: Actually, that's not a bad idea...

@h-mayorquin What do you think about this object hiearchy? `BaseImaging` -> `MemmapImaging` (which can also then handle HDF5 datasets, and any general sliceable memory map, not just numpy ones) which can then splinter into both (1) -> `HDF5Imaging` and (2) -> `BinaryImaging`.

h-mayorquin: I don't know enough `Hdf5` to have an idea of what is to be unified. What will be the common core that is shared in `MemmapImaging`. Looking at the `Hdf5ImagingExtractor` it seems that it does file loading assuming a very specific structure of the hdf5 file. Is the reading that will be unified? the writing? the structure that the videos have inside the class?

src/roiextractors/memmapextractors.py
h-mayorquin: Yes, I was thinking that they will need to import it if we are talking about a data class. I am happy to prototype something though and see how that look / feels.

src/roiextractors/extractors/memmapextractor/memmapextractors.py
CodyCBakerPhD: The `_video` object at this line: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py#L79

is a sliceable memmap-like array whose operations largely (up to transpose, which is why you need `lazy_ops`) match `np.memmap`

The idea is, the `MemmapImaing(memmap=my_memmap_object)` would define the `get_frames()` method, and anything else that makes sense given what you know about the `memmap` object, like it's shape.

The bottom class `__init__` methods would handle all of the format-specific I/O stuff and then hand off their internal objects to the more general extractor.

Look into the structure of the old NeuroscopeRecordingExtractor with respect to the BinDatRecordingExtractor to see what I mean about this kind of structure.

h-mayorquin:
@bendichter @CodyCBakerPhD 
Ok, so we have many ideas here. I am going to list them so maybe we can develop a plan:

- [ ] For the on-going conversion we need a raw format reader which we have in the current implementation.
- [ ] Better data structure (ordered dict, data class or something else) for passing the input (which is the expected image structure) to the current reader.
- [ ] Maybe working in an overall internal standard for video shape throughout `roiextractors` along the lines of: all the files inside `roiextractors` should be of the shape:  `(num_frames, num_channels, rows, columns)`. 
- [ ] Adding a `write_image_method` to this class (done) and write test where all the other formats are read with their respective classes and then written with this format (not-done).
- [ ] Unify the `Hdf5ImagingExtractor` and `MemmapImagingExtractor` in a common ancestor class that abstract some of the properties to be discussed.
- [ ] Something else?

In what order and priority should these be done.

CodyCBakerPhD:
> Maybe working in an overall internal standard for video shape throughout roiextractors along the lines of: all the files inside roiextractors should be of the shape: (num_frames, num_channels, rows, columns).

I guess this one wouldn't be necessary if we implement

>Better data structure (ordered dict, data class or something else) for passing the input (which is the expected image structure) to the current reader.

really well.

src/roiextractors/extractors/memmapextractor/memmapextractors.py
h-mayorquin: Thanks, @CodyCBakerPhD . That does clarify, the handling of the video inside both classes is the thing to be shared. I will check the example you mention.

h-mayorquin:
@CodyCBakerPhD maybe we can marge this one now that the meat of this in #113

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin Yup, this is a nice simple fix to redefine the weird `__init__` for this extractor more properly as a static `write_imaging` method, as it seemed to be intended.

Thanks!

bendichter:
Guys, can we talk about this class? It's confusing me in a few ways:

1. This class both inherits from ImagingExtractor and the constructor takes an ImagingExtractor? It's possible to do both, but my hunch is this was the result of some confusion.
2. I don't see how this class would even operate on a memmap. You can't input a file path or a memmap like object in the constructor. Just an ImagingExtractor? Is that supposed to somehow contain a memmap? I'm totally confused here.
3. This class is referencing internal attributes like self._num_frames, which could have been defined using the shape of a memmap, but are not.

bendichter:
> def write_imaging(imaging_extractor: ImagingExtractor, save_path: PathType = None, verbose: bool = False):
        """
        Static method to write imaging.
        Parameters
        ----------
        imaging: ImagingExtractor object
            The EXTRACT segmentation object from which an EXTRACT native format
            file has to be generated.
            
...what?

CodyCBakerPhD:
> > Guys, can we talk about this class? It's confusing me in a few ways:

@bendichter See this convo for how we're planning to proceed over several modular PRs: https://github.com/catalystneuro/roiextractors/pull/112#discussion_r877603517

> It's confusing me in a few ways:

> This class both inherits from ImagingExtractor and the constructor takes an ImagingExtractor? It's possible to do both, but my hunch is this was the result of some confusion.

Indeed the original form of this newly rediscovered out-of-place interface was very confusing to us all: https://github.com/catalystneuro/roiextractors/blob/f6c37d3e08fe20992e1a095ed8c0479b816f700b/src/roiextractors/memmapextractors.py

Not just the arguments and docstrings, but also the structure of the `__init__`, which is defined in the fashion that a `write_imaging` method is. This PR fixed the location and import of the class in the repo as well as properly defining the `write_imaging`, but with the `__init__` only a placeholder until the next PR which will define and test it in ways compatible with both `np.memmap` and `hdf5.dataset`

> I don't see how this class would even operate on a memmap. You can't input a file path or a memmap like object in the constructor. Just an ImagingExtractor? Is that supposed to somehow contain a memmap? I'm totally confused here.

> This class is referencing internal attributes like self._num_frames, which could have been defined using the shape of a memmap, but are not.

These will both be in the next PR. This PR is just a couple of modular fixes to the general structure of the class.



> The EXTRACT segmentation object from which an EXTRACT native format

Yeah, docstring also needs to be corrected there

PR #113, opened 2022-05-24
Add `NumpyMemmapImagingExtractor`

h-mayorquin:
As discussed in today meeting with @CodyCBakerPhD. 

The idea is to have two ImageExtractors one from `numpy` and another one from `hdf5` which determine the reading logic in their __init__   but also inherit a common and general `MemMapImageExtractor`  logic for the processing methods (e.g. `get_frame`).

This PR should address the reading art for `np.memmap` which is required in the current conversion. #112 is to be transformed in a purely cleaning PR that should pave the way to build the common class in the future.


src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
```diff
@@ -0,0 +1,178 @@
...
+
+        self.file_path = Path(file_path)
+        self._sampling_frequency = sampling_frequency
+        self.offset = offset
+        self.dtype = dtype
+
+        # Get the structure, apply default if not available
+        self.frame_shape = frame_shape
+        image_structure_to_axis = dict() if image_structure_to_axis is None else image_structure_to_axis
+        self.image_structure_to_axis = dict(frame_axis=0, num_channels=1, rows=2, columns=3)
```
CodyCBakerPhD, line 71: I don't know if it's a good idea to have a default for this argument - it's so important, I'd recommend making it required. 

I'd also recommend, if you're sure you want to keep it as a simple dictionary, we should check the keys for possible typos/missing values by asserting frame+rows+col keys are all present (channels could be optional since some binary blobs may represent grayscale and hence have no extra axis for that).

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
```diff
@@ -0,0 +1,178 @@
...
+        # Extract video
+        self._video = self.read_binary_video()
+
+        # Get the image structure as attributes
+        self._rows = self._video.shape[self.image_structure_to_axis["rows"]]
+        self._columns = self._video.shape[self.image_structure_to_axis["columns"]]
+        self._num_channels = self._video.shape[self.image_structure_to_axis["num_channels"]]
+        self._num_frames = self._video.shape[self.image_structure_to_axis["frame_axis"]]
+
+    def read_binary_video(self):
```
CodyCBakerPhD, line 84: Wasn't this going to be a generic `extraction_tool` so that people could use it outside of the extractor API?

CodyCBakerPhD:
@h-mayorquin Looks good! Only minor comments on argument structure

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
h-mayorquin: I am still on the fence about whether this dict structure is good or not. What where the other suggestions? a tuple of two tuples (the first with the values and the second with the mapping and a data class).

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
h-mayorquin: I don't remember this. What we would get of supporting this? my feeling is that I rather support this well tested inside of our class as this pertains to `roieextractors` functionality. Supporting stand-alone use of the function seems to be outside of the scope of the repo. 

`spikeitnterface` does it this way but I don't know why. I am not very convince that we should follow them by default. Maybe if a second use of this function inside the library appears we can move it out of here?

CodyCBakerPhD: Original discussion: https://github.com/catalystneuro/roiextractors/pull/112#discussion_r877590302

>spikeinterface does it this way but I don't know why. I am not very convince that we should follow them by default. Maybe if a second use of this function inside the library appears we can move it out of here?

It's more for working directly with the file contents hands-on, which is a common use case. I personally used this helper function when we were initially exploring these files, so it can be quite useful generally.

As is, it's not a static method so can't be called unless `__init__` succeeded, but maybe you don't entirely know all of the arguments to the currently designed `__init__`, so having a helper to explore the file structure can help.

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
CodyCBakerPhD: Not a tuple-of-tuples, definitely. That's one of those data structures that is less Pythonic and more reminiscent of MATLAB.

What we want is definitely something dictionary-like, but with constraints (limited possibilities) and convenience (fast, informative error message for typos).

Can always start off with a `dict` + `assert` and change it later if we can think of a better way. It is, I suppose, the simplest thing to ask of a user if not the most sophisticated from a development standpoint.

bendichter:
You can't inherit from MemmapImagingExtractor?

CodyCBakerPhD:
@bendichter The next PR will involve defining the `MemmapImaging.__init__` and other key functions shared between this and the `HDF5Imaging` so that both can then be refactored to depend on the it. Just taking it one step at a time

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
h-mayorquin: Yes, after I thought more about it and re-read your arguments in #112 I have changed my mind. Check #115 for an implementation of a data-class for handling this.

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
h-mayorquin: OK, let's move this outside. Check #115 for a stand-alone implementation of the functionality.

tests/test_numpymemmapextractor.py
```diff
@@ -0,0 +1,150 @@
...
+from roiextractors import NumpyMemmapImagingExtractor
+
+from .setup_paths import OUTPUT_PATH
+
+
+def custom_name_func(testcase_func, param_num, param):
+    return f"{testcase_func.__name__}_{param_num}_" f"_{param.kwargs.get('case_name', '')}"
+
+
+class TestMemmapExtractor(unittest.TestCase):
```
CodyCBakerPhD, line 19:
```suggestion
class TestNumpyMemmapExtractor(unittest.TestCase):
```

src/roiextractors/extractors/memmapextractors/numpymemampextractor.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @staticmethod
+    def write_imaging(imaging_extractor: ImagingExtractor, save_path: PathType = None, verbose: bool = False):
+        """
+        Static method to write imaging.
+
+        Parameters
+        ----------
+        imaging: ImagingExtractor object
+            The EXTRACT segmentation object from which an EXTRACT native format
```
CodyCBakerPhD, line 146: Gonna want to remove this bit

h-mayorquin: Yes!

h-mayorquin: Done.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin OK, next step is to generalize the memmap access for the two interfaces into the `MemmapImaging.__init__` and then we're good. Should see some decent code reduction there.

Again, I also think we can just have the `NumpyImaging.write_imaging` inherit the `MemmapImaging.write_imaging` since they ought to be doing almost the same exact thing (the binary blob has no format specificity to Numpy).

PR #114, opened 2022-05-25
re-enable check_comparisons in `test_io`

h-mayorquin:
I re-enabled the complete checks in the testing. It is working for the imaging extractors bur right now most of the segmentation extractors are missing. I fix one of them and intend to fix the next ones in following PRs.


src/roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -81,12 +81,17 @@ def get_accepted_list(self):
         accepted = self._dataset_file["estimates"]["idx_components"]
         if len(accepted.shape) == 0:
             accepted = list(range(self.get_num_rois()))
+        else:
+            accepted = list(accepted[:])
```
CodyCBakerPhD, line 85: Hmm.. What is this doing? What data type is the `accepted` object (or `self._dataset_file` for that matter) and what are it's contents being sliced out?

Just curious - assuming it's a lazy access data value of some kind, `[:]` is the call to retrieve all values and the output have to be casted as a `list` to be consistent with the other return type.

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -81,12 +81,17 @@ def get_accepted_list(self):
...
 
     def get_rejected_list(self):
         rejected = self._dataset_file["estimates"]["idx_components_bad"]
-        if len(rejected.shape) > 0:
-            return rejected
+        if len(rejected.shape) == 0:
+            rejected = list()
+        else:
+            rejected = list(rejected[:])
+        return rejected
```
CodyCBakerPhD, line 94: So this returning `None` was the source of the original error?

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin Cool, thanks for looking into this!

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
h-mayorquin: Yes, this and the one above. This one was a typing check if I remember well.

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
h-mayorquin: This is a file read with `h5.py`. (`class 'h5py._hl.files.File'>`).

So what was happening is the following:
The tests are composed of a 1) reading the format, 2) using the write function, 3) read again from what was written, 4) compare.
The first time this is read (1) accepted had shape 0 (`HDF5 dataset "idx_components": shape (), type "|O">`), so the line in 83 would kick in. That means that accepted was a list of integers. However, using the write function (2) this gets written to the proper place and when is written again now accepted has data but is an `h5py` object so it was returned in that format and the comparison in 4 failed.

The added line makes the type consistent.

PR #115, opened 2022-05-25
Add read binary numpy video as stand alone

h-mayorquin:
In the light of the discussion in #113 I am doing two things in this PR:

1) Created a data structure to model the structure of the expected video. This is implemented with a dataclass that allows to do structural checks at initialization and ensures name consistency 

2) Defined a binary reader fro numpy memmap in `extraction_tools` that operates independently and uses `VideoStructure` class as an input.  

I added unit-tests for the functionality in a new directory following the organization of `nwb-conversion-tools`.


src/roiextractors/extraction_tools.py
```diff
@@ -26,6 +29,88 @@
 FloatType = float
 
 
+@dataclass()
```
CodyCBakerPhD, line 32: Do you need the `()` here? Usually @dataclass alone works, I thought

src/roiextractors/extraction_tools.py
```diff
@@ -26,6 +29,88 @@
...
+        axis_values_are_not_unique = len(axis_values) != 4
+        if axis_values_are_not_unique:
+            raise ValueError(exception_message)
+
+        values_out_of_range = any([axis < 0 or axis > 4 for axis in axis_values])
+        if values_out_of_range:
+            raise ValueError(exception_message)
+
+    def build_video_shape(self, n_frames: int) -> Tuple[int, int, int, int]:
+        video_shape = [None, None, None, None]
```
CodyCBakerPhD, line 73:
```suggestion
        video_shape = [None] * 4
```

Simpler way of defining a list of homogenous items of specified length

src/roiextractors/extraction_tools.py
```diff
@@ -26,6 +29,88 @@
...
+        video_shape[self.frame_axis] = n_frames
+        video_shape[self.rows_axis] = self.rows
+        video_shape[self.columns_axis] = self.columns
+        video_shape[self.num_channels_axis] = self.num_channels
```
CodyCBakerPhD, line 77: This kind of setup is where I was seriously wondering if it might be beneficial to just enforce a common standard of axis ordering that we can map to. *I.e.*, `video_shape[0]` is always the frames, and you can use the `frame_axis` to map from the source `memmap` to that.

src/roiextractors/extraction_tools.py
```diff
@@ -26,6 +29,88 @@
...
+
+        return tuple(video_shape)
+
+    def transform_video_to_canonical_form(self, video: np.array) -> np.array:
+        canonical_frame_axis = 0
+        canonical_rows_axis = 1
+        canonical_columns_axis = 2
+        canonical_num_channels_axis = 3
```
CodyCBakerPhD, line 85: But I do suppose that is the kind of thing that would have more of a place in the Extractors; the tool as mentioned would only handle direct file access given what we know about the basic layout. OK

CodyCBakerPhD:
@h-mayorquin Very nice - the basic tool function is precisely what we wanted there, and the tests look great.

The dataclass is a bit complicated, but has a nice sophistication in resolving the mapping details. Time will tell if we prefer that way of doing things or if it can perhaps be reduced to a simple dictionary.

src/roiextractors/extraction_tools.py
h-mayorquin: No, this is not needed. I was thinking on making the dataclass [frozen](https://docs.python.org/3/library/dataclasses.html#frozen-instances) but I realize that it does not play that well with the `__post_init__`

src/roiextractors/extraction_tools.py
```diff
@@ -30,6 +33,88 @@
 FloatType = float
 
 
+@dataclass()
```
h-mayorquin, line 36:
```suggestion
@dataclass
```

h-mayorquin:
@CodyCBakerPhD 
Git-annex error:
```python
CommandError: '"git" "-c" "diff.ignoreSubmodules=none" "cat-file" "blob" "git-annex:remote.log"' failed with exitcode 128 under D:\a\roiextractors\roiextractors\ophys_testing_data
fatal: path 'remote.log' does not exist in 'git-annex'
Error: Process completed with exit code 1.
```

CodyCBakerPhD:
> @CodyCBakerPhD
Git-annex error:

@h-mayorquin Yup, I've been tracking it. Trying to fix now

PR #116, opened 2022-05-25
Modified import  of `mat_struct` for deprecation warning

h-mayorquin:
Small PR that brings the number of warnings in this repo to 0!
: D

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Union
 
 import numpy as np
+import scipy
 from spikeextractors.extraction_tools import cast_start_end_frame
 from tqdm import tqdm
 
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Ok, so the problem is Python 3.7, I fixed with a conditional import that fetches the right attribute when it exists and gets it from the location in the scipy version that gets installed with Python 3.7 when not.

src/roiextractors/extraction_tools.py
```diff
@@ -13,7 +14,10 @@
 except ImportError:
     HAVE_H5 = False
 try:
-    import scipy.io as spio
+    if hasattr(scipy.io.matlab, "mat_struct"):
```
CodyCBakerPhD, line 17: I, for one, very much appreciate that Python modules treat submodules as attributes.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin Very nice, thanks! Would this be necessary as well on NWBCT (like Neuroscope/CellExplorer Interfaces?)

h-mayorquin:
> @h-mayorquin Very nice, thanks! Would this be necessary as well on NWBCT (like Neuroscope/CellExplorer Interfaces?)

I will take a loot into it.

PR #118, opened 2022-05-25
[Testing] non-memmap TIFF

### LINKED ISSUE #100: [Testing]: Add example of non-memmap TIFF file
Ben may have run into one during the Froemke conversion, could add it to the GIN ~~behavior~~ ophys repo for the CI to use.

CodyCBakerPhD:
fix #100

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -24,7 +24,7 @@ jobs:
 
       - if: ${{ matrix.os == 'ubuntu-latest' }}
         name: Get datalad - Linux
-        run: conda install -c conda-forge datalad==0.16.1
+        run: conda install -c conda-forge datalad==0.16.3
       - if: ${{ matrix.os == 'windows-latest' }}
         name: Get git-annex - Windows
         uses: crazy-max/ghaction-chocolatey@v1.6.0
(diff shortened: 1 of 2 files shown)


.github/workflows/testing.yml
```diff
@@ -24,15 +24,15 @@ jobs:
 
       - if: ${{ matrix.os == 'ubuntu-latest' }}
         name: Get datalad - Linux
-        run: conda install -c conda-forge datalad==0.16.1
+        run: conda install -c conda-forge datalad==0.16.3
```
h-mayorquin, line 27: Should we do the same update on nwb-conversion-tools?

h-mayorquin (commented):
This will always through a warning, right?

https://github.com/catalystneuro/roiextractors/blob/40c1ad88bbf8eb8f031254a78bfff57d66b48742/src/roiextractors/extractors/tiffimagingextractor/tiffimagingextractor.py#L48-L54

If we are going to have that warning there should't we have something in the testing that catches it and, for this round trip test, suppresses it?

CodyCBakerPhD:
> If we are going to have that warning there should't we have something in the testing that catches it and, for this round trip test, suppresses it?

Oh, sure good point

CodyCBakerPhD:
> If we are going to have that warning there should't we have something in the testing that catches it and, for this round trip test, suppresses it?

@h-mayorquin Added now

h-mayorquin (approved):

PR #122, opened 2022-05-26
Memmap ImagingExtractor classes unification I

h-mayorquin:
This just moves the `write_imaging` method from `NumpyMemmapImagingExtractor` to `MemmapImagingExtractor` Subsequent PRs should also migrate the reading functionality to the `MemmapImagingExtractor` and unify this with the `Hdf5ImagingExtractor`.

### DIFF
--- a/src/roiextractors/extractors/memmapextractors/memmapextractors.py
+++ b/src/roiextractors/extractors/memmapextractors/memmapextractors.py
@@ -5,7 +5,7 @@ import psutil
 from tqdm import tqdm
 
 from ...imagingextractor import ImagingExtractor
-from typing import Tuple, Dict
+from typing import Tuple, Dict, Optional
 
 from ...extraction_tools import (
     PathType,
@@ -66,7 +66,7 @@ class MemmapImagingExtractor(ImagingExtractor):
         imaging_extractor: ImagingExtractor,
         save_path: PathType,
         verbose: bool = False,
-        buffer_gb: float = 0,
+        buffer_size_in_gb: Optional[float] = None,
     ):
         """
         Static method to write imaging.
@@ -78,15 +78,20 @@ class MemmapImagingExtractor(ImagingExtractor):
             path to save the native format to.
         verbose: bool
             Displays a progress bar.
-        buffer_gb: int
-            The size of the buffer in Gigabytes. Default of 0 forces chunks that cover only one image.
+        buffer_size_in_gb: float
+            The size of the buffer in Gigabytes. The default of None results in buffering over one frame at a time.
         """
+
+        # The base and default case is to load one image at a time.
+        if buffer_size_in_gb is None:
+            buffer_size_in_gb = 0
+
         imaging = imaging_extractor
         file_size_in_bytes = Path(imaging.file_path).stat().st_size
         available_memory_in_bytes = psutil.virtual_memory().available
-        buffer_size_in_bytes = buffer_gb * 1e9
+        buffer_size_in_bytes = int(buffer_size_in_gb * 1e9)
         if available_memory_in_bytes < buffer_size_in_bytes:
-            raise f"Not enough memory available memory {available_memory_in_bytes* 1e9} for buffer size {buffer_gb}"
+            raise f"Not enough memory available, {available_memory_in_bytes* 1e9} for buffer size {buffer_size_in_gb}"
 
         num_frames = imaging.get_num_frames()
         memmap_shape = imaging.video_structure.build_video_shape(n_frames=num_frames)
@@ -105,27 +110,27 @@ class MemmapImagingExtractor(ImagingExtractor):
             video_memmap[:] = video_data_to_save
 
         else:
-            chunk_size_in_bytes = int(buffer_size_in_bytes)
+            buffer_size_in_bytes = int(buffer_size_in_bytes)
             type_size = np.dtype(dtype).itemsize
 
             n_channels = imaging.get_num_channels()
             pixels_per_frame = n_channels * np.product(imaging.get_image_size())
             bytes_per_frame = type_size * pixels_per_frame
-            frames_per_chunk = chunk_size_in_bytes // bytes_per_frame
+            frames_in_buffer = buffer_size_in_bytes // bytes_per_frame
 
-            # If the chunk size is smaller than one image, the iterator goes over one image only.
-            frames_per_chunk = max(frames_per_chunk, 1)
-            iterator = range(0, num_frames, frames_per_chunk)
+            # If the buffer size is smaller than the size of one image, the iterator goes over one image only.
+            frames_in_buffer = max(frames_in_buffer, 1)
+            iterator = range(0, num_frames, frames_in_buffer)
             if verbose:
                 iterator = tqdm(iterator, ascii=True, desc="Writing to .dat file")
 
             for frame in iterator:
-                end_frame = min(frame + frames_per_chunk, num_frames)
+                end_frame = min(frame + frames_in_buffer, num_frames)
 
-                # Get the video chunk
-                video_chunk = imaging.get_video(start_frame=frame, end_frame=end_frame)
+                # Get the video in buffer
+                video_in_buffer = imaging.get_video(start_frame=frame, end_frame=end_frame)
 
-                # Fit the video chunk in the memmap array
+                # Fit the video buffer in the memmap array
                 indices = np.arange(start=frame, stop=end_frame)
                 axis_to_expand = (
                     imaging.video_structure.rows_axis,
@@ -134,7 +139,7 @@ class MemmapImagingExtractor(ImagingExtractor):
                 )
                 indices = np.expand_dims(indices, axis=axis_to_expand)
                 frame_axis = imaging.video_structure.frame_axis
-                np.put_along_axis(arr=video_memmap, indices=indices, values=video_chunk, axis=frame_axis)
+                np.put_along_axis(arr=video_memmap, indices=indices, values=video_in_buffer, axis=frame_axis)
 
         # Flush the video and delete it
         video_memmap.flush()
--- a/src/roiextractors/extractors/memmapextractors/memmapextractors.py
+++ b/src/roiextractors/extractors/memmapextractors/memmapextractors.py
@@ -64,9 +64,9 @@ class MemmapImagingExtractor(ImagingExtractor):
     @staticmethod
     def write_imaging(
         imaging_extractor: ImagingExtractor,
-        save_path: PathType = None,
+        save_path: PathType,
         verbose: bool = False,
-        buffer_data: bool = False,
+        buffer_gb: int = 0,
     ):
         """
         Static method to write imaging.
@@ -78,31 +78,34 @@ class MemmapImagingExtractor(ImagingExtractor):
             path to save the native format to.
         verbose: bool
             Displays a progress bar.
-        buffer_data: bool
-            Forces chunk to occur even if memmory is available
+        buffer_gb: int
+            The size of the buffer in Gigabytes. Default of 0 forces chunks that cover only one image.
         """
         imaging = imaging_extractor
-        memory_safety_margin = 0.80  # Accept a file smaller than 80 per cent of available memory
         file_size_in_bytes = Path(imaging.file_path).stat().st_size
         available_memory_in_bytes = psutil.virtual_memory().available
+        buffer_size_in_bytes = buffer_gb * 10**9
+        if available_memory_in_bytes < buffer_size_in_bytes:
+            raise f"Not enough memory available memory {available_memory_in_bytes* 10**9} for buffer size {buffer_gb}"
 
-        memory_limit = available_memory_in_bytes * memory_safety_margin
-        if file_size_in_bytes < memory_limit and not buffer_data:
+        num_frames = imaging.get_num_frames()
+        memmap_shape = imaging.video_structure.build_video_shape(n_frames=num_frames)
+        dtype = imaging.get_dtype()
+
+        # Load the memmap
+        video_memmap = np.memmap(
+            save_path,
+            shape=memmap_shape,
+            dtype=dtype,
+            mode="w+",
+        )
+
+        if file_size_in_bytes < buffer_size_in_bytes:
             video_data_to_save = imaging.get_frames()
-            memmap_shape = video_data_to_save.shape
-            video_memmap = np.memmap(
-                save_path,
-                shape=memmap_shape,
-                dtype=imaging.get_dtype(),
-                mode="w+",
-            )
-
             video_memmap[:] = video_data_to_save
-            video_memmap.flush()
 
         else:
-            chunk_size_in_bytes = int(available_memory_in_bytes * memory_safety_margin)
-            dtype = imaging.get_dtype()
+            chunk_size_in_bytes = int(buffer_size_in_bytes)
             type_size = np.dtype(dtype).itemsize
 
             n_channels = imaging.get_num_channels()
@@ -110,9 +113,8 @@ class MemmapImagingExtractor(ImagingExtractor):
             bytes_per_frame = type_size * pixels_per_frame
             frames_per_chunk = chunk_size_in_bytes // bytes_per_frame
 
-            num_frames = imaging.get_num_frames()
-            memmap_shape = imaging.video_structure.build_video_shape(n_frames=num_frames)
-
+            # If the chunk size is smaller than one image, the iterator goes over one image only.
+            frames_per_chunk = max(frames_per_chunk, 1)
             iterator = range(0, num_frames, frames_per_chunk)
             if verbose:
                 iterator = tqdm(iterator, ascii=True, desc="Writing to .dat file")
@@ -122,14 +124,6 @@ class MemmapImagingExtractor(ImagingExtractor):
 
                 # Get the video chunk
                 video_chunk = imaging.get_video(start_frame=frame, end_frame=end_frame)
-
-                # Load the memmap
-                video_memmap = np.memmap(
-                    save_path,
-                    shape=memmap_shape,
-                    dtype=dtype,
-                    mode="w+",
-                )
 
                 # Fit the video chunk in the memmap array
                 indices = np.arange(start=frame, stop=end_frame)
@@ -142,6 +136,6 @@ class MemmapImagingExtractor(ImagingExtractor):
                 frame_axis = imaging.video_structure.frame_axis
                 np.put_along_axis(arr=video_memmap, indices=indices, values=video_chunk, axis=frame_axis)
 
-                # Flush to liberate memory for next iteration
-                video_memmap.flush()
-                del video_memmap
+        # Flush the video and delete it
+        video_memmap.flush()
+        del video_memmap
--- a/src/roiextractors/extractors/memmapextractors/memmapextractors.py
+++ b/src/roiextractors/extractors/memmapextractors/memmapextractors.py
@@ -66,7 +66,7 @@ class MemmapImagingExtractor(ImagingExtractor):
         imaging_extractor: ImagingExtractor,
         save_path: PathType = None,
         verbose: bool = False,
-        force_chunk: bool = False,
+        chunk_data: bool = False,
     ):
         """
         Static method to write imaging.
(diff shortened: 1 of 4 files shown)


src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -71,36 +69,20 @@ def write_imaging(imaging_extractor: ImagingExtractor, save_path: PathType = Non
...
-            The EXTRACT segmentation object from which an EXTRACT native format
-            file has to be generated.
         save_path: str
             path to save the native format.
         overwrite: bool
             If True and save_path is existing, it is overwritten
         """
         imaging = imaging_extractor
-        video_to_save = np.memmap(
+        video_data_to_save = imaging.get_frames()[:]
```
CodyCBakerPhD, line 78: May want to use `psutil` to ensure there is actually enough memory available to perform this operation, otherwise the users computer will crash.

CodyCBakerPhD: Also, `tqdm` is always nice, I'd restore that functionality here.

h-mayorquin: Good idea.
I am pretty sure I saw that use of `psutil` somewhere over here but can't find it, am I imaging things? 

I guess you are envisioning something like that is available on the write method of `TiffImagingExtractor`?

CodyCBakerPhD: > I am pretty sure I saw that use of psutil somewhere over here but can't find it, am I imaging things?

https://github.com/catalystneuro/nwb-conversion-tools/blob/a0e5c3b564ff013668f01ceb91bdfc285b2efbda/src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py#L227-L233


> I guess you are envisioning something like that is available on the write method of TiffImagingExtractor?

The red parts of this changelog show how tqdm was used here already. More sophisticated approaches might use a `hdmf.data_utils.GenericDataChunkIterator` which comes along with `tqdm` already, but that might be a bit much to do here.

h-mayorquin: What is that 70 doing there in:
`uncompressed_estimate = Path(file).stat().st_size * 70`

Is it assuming a compression ratio of .7 for the file or something like that?

CodyCBakerPhD: Ohh uhh you can probably ignore that for these cases. That was a manual scalar inflation value I got from experimenting on with loading big movie files from OpenCV. But for all the imaging extractors here, the data is already more or less uncompressed

h-mayorquin: All right, check out the new changes.

h-mayorquin:
Tests are passing in Ubuntu. Git-annex problem in windows.

CodyCBakerPhD:
> Git-annex problem in windows.

It's almost like all these package distributors don't want to host the recent versions of git-annex, lol.

CodyCBakerPhD:
I'm hoping it was just a curl/url issue fixed by rerunning but if not we can just disable it for lesser versions of python. Mac/Windows were always a massive pain to setup git-annex and it wasn't until recently that a solution was discovered, but it appears to be a rather temporary one.

h-mayorquin:
> I'm hoping it was just a curl/url issue fixed by rerunning but if not we can just disable it for lesser versions of python. Mac/Windows were always a massive pain to setup git-annex and it wasn't until recently that a solution was discovered, but it appears to be a rather temporary one.

Yes, it seems that it was that indeed.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +59,87 @@ def get_num_channels(self):
...
+            bytes_per_frame = type_size * pixels_per_frame
+            frames_per_chunk = chunk_size_in_bytes // bytes_per_frame
+            # number_of_chunks = file_size_in_bytes / chunk_size_in_bytes
+
+            num_frames = imaging.get_num_frames()
+            memmap_shape = imaging.video_structure.build_video_shape(n_frames=num_frames)
+
+            iterator = range(0, num_frames, frames_per_chunk)
+            if verbose:
+                iterator = tqdm(iterator, ascii=True, desc="Writing to .tiff file")
```
h-mayorquin, line 116:
```suggestion
                iterator = tqdm(iterator, ascii=True, desc="Writing to .dat file")
```

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,86 @@ def get_num_channels(self):
         return self._num_channels
 
     @staticmethod
-    def write_imaging(imaging_extractor: ImagingExtractor, save_path: PathType = None, verbose: bool = False):
+    def write_imaging(
+        imaging_extractor: ImagingExtractor,
+        save_path: PathType = None,
+        verbose: bool = False,
+        force_chunk: bool = False,
```
CodyCBakerPhD, line 69: For consistency, it might just be nice to follow either the MovieInterface or RecordingInterfaces as much as possible.

The RecordingInterfaces use `iterator_opts` which is a dictionary passed directly as kwargs, including the `buffer_gb`, which can be set to larger than the file size, which has the equivalent effect of `force_chunk=False` here.

The MovieInterface is more similar, a boolean option called `chunk_data`: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/datainterfaces/behavior/movie/moviedatainterface.py#L110, though it is recommended to be `True` by default (in general, just a good practice I've found)

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,86 @@ def get_num_channels(self):
...
-                print(f"Saving channel {ch}")
-                for i in tqdm(range(imaging.get_num_frames())):
-                    plane = imaging.get_frames(i, channel=ch)
-                    video_to_save[ch, i] = plane
-        else:
-            for ch in range(imaging.get_num_channels()):
-                for i in range(imaging.get_num_frames()):
-                    plane = imaging.get_frames(i, channel=ch)
-                    video_to_save[ch, i] = plane
+        safety_margin = 0.80  # Accept a file smaller than 80 per cent of available memory
```
CodyCBakerPhD, line 85: A more informative name perhaps `memory_safety_margin`?

h-mayorquin: Yes, that's better. Thanks for the suggestion.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,86 @@ def get_num_channels(self):
...
+                # Get the video chunk
+                video_chunk = imaging.get_video(start_frame=frame, end_frame=end_frame)
+
+                # Load the memmap
+                video_memmap = np.memmap(
+                    save_path,
+                    shape=memmap_shape,
+                    dtype=dtype,
+                    mode="w+",
+                )
```
CodyCBakerPhD, line 132: Why can't you just keep the `video_memmap` open the entire time and pull a different frame slice at each iteration?

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Good idea. I will follow `MovieInterface`.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Would that work? I am remember the thing that @bendichter mentioned that `memmaps` tend to keep accessed data in the cache.

CodyCBakerPhD: The writing stages here seem a little odd - I'd refer to https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/core_tools.py#L181 (minus the parallelization) for more a typical way of doing it

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Done.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Done.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: Also, it should really probably be called 'buffer_data', even in the MovieInterface. 'Chunk' gets conflated when talking about binary vs. HDF5, so 'buffer' is more accurate to what we're doing.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: > Would that work? I am remember the thing that @bendichter mentioned that memmaps tend to keep accessed data in the cache.

That was something Ben mentioned that Oliver mentioned but I've not seen it produce that behavior w.r.t. buffering processes in practice.

h-mayorquin: Can you be more specific? At a first glance I see a mechanism of 
1) getting the traces.
2) Put them in the memmap.

Save the re-instantiation and deletion of the memmap at each loop it seems to the same to me. What I am missing?

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: OK, chunking is confusing to me I have to say.

The thing that we are doing is something like:
"Loading the data in a small chunks by means (by means of a buffer?) so we don't overload memory" 

I wonder if there is a better way of denoting that.

h-mayorquin: Done.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: The line I linked: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/core_tools.py#L181 - slice a frame region of the writing memmap, set equal to array extraction of data read from the imaging extractor. That's all you need to iterate, and I'd also be surprised if you need to flush each time.

The part I'm say you can ignore is the executor type setup: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/core_tools.py#L240-L242

Also, I just remembered Ben did a good job with this recent tool: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/tools/hdmf.py#L7-L23, I'd recommend giving it a try (we'll propagate it up to `hdmf` at some point).

h-mayorquin: > The line I linked: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/core_tools.py#L181 - slice a frame region of the writing memmap, set equal to array extraction of data read from the imaging extractor. That's all you need to iterate, and I'd also be surprised if you need to flush each time.

But the code that I have is the same save flushing and loading the memmap. Do you want those to be out of the loop? is that what are you saying?

CodyCBakerPhD: I'm saying your 13 lines: https://github.com/catalystneuro/roiextractors/pull/122/files#diff-f905195c77c6ce028089c3110e56083aedaf38e2592b2367f1af856d40a38148R134-R147

should be able to be done in one line, as in: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/core_tools.py#L181 and https://github.com/catalystneuro/nwb-conversion-tools/blob/main/src/nwb_conversion_tools/tools/hdmf.py#L23

h-mayorquin: They can't be done like that if you don't know which axis the frames are. 
All the machinery over there is to solve that problem.

h-mayorquin: @CodyCBakerPhD 
The requested refactor is implemented in #125 
Let me know if there is something else left to merge this.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
         return self._num_channels
 
     @staticmethod
-    def write_imaging(imaging_extractor: ImagingExtractor, save_path: PathType = None, verbose: bool = False):
+    def write_imaging(
+        imaging_extractor: ImagingExtractor,
+        save_path: PathType,
+        verbose: bool = False,
+        buffer_gb: int = 0,
```
CodyCBakerPhD, line 69: @h-mayorquin Minor detail - annotation type should be a `float`. It's also a little odd to default it to `0` (which I assume means no buffering). I'd think if `buffer_gb` is set to any value, then that should specify using iterative write. To disable iterative write, either set an artificial `buffer_gb > actual_data_size` or don't make use of iterative write code at all. The former would probably take less code, but in either case I think it makes more sense to set the default value here to `None` to indicate no iterative write.

```
        buffer_gb: Optional[float] = None,
```

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
-            file has to be generated.
+        imaging: An ImagingExtractor object that inherited from MemmapImagingExtractor
         save_path: str
-            path to save the native format.
-        overwrite: bool
-            If True and save_path is existing, it is overwritten
+            path to save the native format to.
+        verbose: bool
+            Displays a progress bar.
+        buffer_gb: int
```
CodyCBakerPhD, line 81:
```suggestion
        buffer_gb: float, optional
```
Here as well

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
+            The size of the buffer in Gigabytes. Default of 0 forces chunks that cover only one image.
```
CodyCBakerPhD, line 82:
```
            The size of the buffer in Gigabytes. The default of None results in buffering over one frame at a time.
```

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
         """
         imaging = imaging_extractor
-        video_to_save = np.memmap(
+        file_size_in_bytes = Path(imaging.file_path).stat().st_size
+        available_memory_in_bytes = psutil.virtual_memory().available
+        buffer_size_in_bytes = buffer_gb * 10**9
```
CodyCBakerPhD, line 87: In line with setting this argument appropriately as a `float`, this should use `1e9` instead of `10**9`

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
 
-        video_to_save.flush()
+        else:
+            chunk_size_in_bytes = int(buffer_size_in_bytes)
+            type_size = np.dtype(dtype).itemsize
+
+            n_channels = imaging.get_num_channels()
+            pixels_per_frame = n_channels * np.product(imaging.get_image_size())
+            bytes_per_frame = type_size * pixels_per_frame
+            frames_per_chunk = chunk_size_in_bytes // bytes_per_frame
```
CodyCBakerPhD, line 114: If we are indeed changing from 'chunk' to 'buffer', let's swap these internal variable names too to avoid future dev confusion as well

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
+            for frame in iterator:
+                end_frame = min(frame + frames_per_chunk, num_frames)
+
+                # Get the video chunk
+                video_chunk = imaging.get_video(start_frame=frame, end_frame=end_frame)
+
+                # Fit the video chunk in the memmap array
+                indices = np.arange(start=frame, stop=end_frame)
+                axis_to_expand = (
+                    imaging.video_structure.rows_axis,
```
CodyCBakerPhD, line 131: Other `ImagingExtractors` don't have `video_structure` as an attribute besides the new `NumpyMemmapExtractor`, right? These static methods ought to be designed to apply to any other extractor, especially if they're caching a binary blob.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -64,43 +62,80 @@ def get_num_channels(self):
...
+                    imaging.video_structure.columns_axis,
+                    imaging.video_structure.num_channels_axis,
+                )
+                indices = np.expand_dims(indices, axis=axis_to_expand)
+                frame_axis = imaging.video_structure.frame_axis
+                np.put_along_axis(arr=video_memmap, indices=indices, values=video_chunk, axis=frame_axis)
```
CodyCBakerPhD, line 137: Confused why you still need all this axis management here; if we enforce the `(frames, rows, cols, channels)` structure, and enforce that calls to `get_video` return the same data in that similar convention, then you should just be able to set the `video_memmap` values directly as

```
video_memmap[n_frames_in_buffer, ...] = video_buffer  # 'video_chunk' -> 'video_buffer' as well
```

h-mayorquin: This is done in the #125.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: This is done in the https://github.com/catalystneuro/roiextractors/pull/125.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Please remind me what is the logic again?
You are using buffer to mean "the size of the thing that we load" and "chunk" to mean the minimal size that could be loaded. Is that correct?

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Makes sense.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: No, you are wrong. 0 means that you will write frame by frame. I think that this makes as much sense as using `None` to denote this. 

The iterative write is disabled with when the buffer is large enough as your specifying.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Done in the last commit.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: OK

CodyCBakerPhD (dismissed):

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: 'Chunk' is a specific reference to a thing in a HDF5 dataset. We should probably avoid using the term in relation to binary blobs to avoid confusion between what we're talking about. The 'buffer' is the total amount of data being read into memory. Here, that means, really, the number of frames that you read and write at a single operation.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
CodyCBakerPhD: > No, you are wrong. 0 means that you will write frame by frame.

Ah, I did not realize that writing frame-by-frame implies that the size of the buffer is not only less than the size of a frame, but in fact zero!

`None` communicates this 'special' behavior more clearly, since it is equivalent to performing an automatic calulation of `buffer_gb=dtype.itemsize*n_rows*n_cols*n_channels/1e9`, which is clearly not zero.

h-mayorquin: All right, I am on the fence on this but you have more experience dealing with this object. The desired changes are in the last commit.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Thanks!

PR #126, opened 2022-05-31
Function `get_image_size` should return tuples

h-mayorquin:
The `get_image_size` function of the `SbxImagingExtractor` was returning an array which was causing problems in `nwb-conversion-tools`. 

Moreover, I think that all the `get_image_size` functions should return a tuple to mark the image size properly as an immutable. This PR fixes the problem and makes the output for all the imaging extractors a tuple.

### DIFF
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -9,6 +9,7 @@ from ...extraction_tools import (
     write_to_h5_dataset_format,
 )
 from ...imagingextractor import ImagingExtractor
+from typing import Tuple
 
 try:
     import h5py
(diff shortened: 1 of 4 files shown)


src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -108,7 +108,7 @@ def get_frames(self, frame_idxs, channel=0):
             # return lazy_ops.DatasetView(self._video).lazy_slice[channel, sorted_idxs].lazy_slice[:, argsorted_idxs]
 
     def get_image_size(self):
```
CodyCBakerPhD, line 110: Maybe add annotation typing returns to super clarify this, also all the way up to the abstract level? 😁

```python
def get_image_size(self) -> Tuple[int, int]:
```

CodyCBakerPhD: Otherwise yeah, totally supportive of this. Thanks for the small quick PR!

h-mayorquin: Should be done now.

CodyCBakerPhD (approved):

PR #127, opened 2022-06-01
Add MultiImagingExtractor

### LINKED ISSUE #119: Add MultiImagingExtractor
We have a MultiSegmentationExtractor, but no MultiImagingExtractor. @weiglszonja needs this for the Fee lab project in order to splice together a series of separate ImagingExtractors into one continuous ImagingExtractor to be written to the same TwoPhotonSeries

weiglszonja:
This PR introduces a new extractor object that can be used to combine multiple ImagingExtractors into to one continuous ImagingExtractor to be written to the same TwoPhotonSeries object.

Resolve #119

### DIFF
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -1,6 +1,6 @@
 from abc import ABC
 from array import ArrayType
-from typing import Tuple, List
+from typing import Tuple, List, Iterable
 
 import numpy as np
 
@@ -56,25 +56,32 @@ class MultiImagingExtractor(ImagingExtractor, ABC):
             get_channel_names="The name of the channels",
         )
         for method, property_message in properties_to_check.items():
-            values = [getattr(extractor, method, None) for extractor in self._imaging_extractors]
-            unique_values = list(set(values))
+            values = [getattr(extractor, method, None)() for extractor in self._imaging_extractors]
+            unique_values = set(tuple(v) if isinstance(v, Iterable) else v for v in values)
             assert (
                 len(unique_values) == 1
             ), f"{property_message} is not consistent over the files (found {unique_values})."
 
     def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> NumpyArray:
+        assert max(frame_idxs) < self._num_frames, "'frame_idxs' range beyond number of available frames!"
         extractor_indices = np.searchsorted(self._end_frames, frame_idxs, side="right")
+        # Match frame_idxs to imaging extractors
+        extractors_dict = {}
+        for extractor_index, frame_index in zip(extractor_indices, frame_idxs):
+            extractors_dict.setdefault(extractor_index, []).append(frame_index)
 
         frames_to_concatenate = []
         # Extract frames for each extractor and concatenate
-        for extractor_index in extractor_indices:
+        for extractor_index, frame_indices in extractors_dict.items():
             frames_for_each_extractor = self._get_frames_from_an_imaging_extractor(
                 extractor_index=extractor_index,
-                frame_idxs=frame_idxs,
+                frame_idxs=frame_indices,
             )
-            frames_to_concatenate.append(frames_for_each_extractor[np.newaxis, ...])
+            if len(frame_indices) == 1:
+                frames_for_each_extractor = frames_for_each_extractor[np.newaxis, ...]
+            frames_to_concatenate.append(frames_for_each_extractor)
 
-        frames = np.concatenate(frames_to_concatenate, axis=0)
+        frames = np.concatenate(frames_to_concatenate, axis=0).squeeze()
         return frames
 
     def _get_frames_from_an_imaging_extractor(self, extractor_index: int, frame_idxs: ArrayType):
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -56,9 +56,10 @@ class MultiImagingExtractor(ImagingExtractor):
     @check_get_frames_args
     def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> NumpyArray:
         extractor_indices = np.searchsorted(self._end_frames, frame_idxs, side="right")
+        relative_frame_indices = frame_idxs - np.array(self._start_frames)[extractor_indices]
         # Match frame_idxs to imaging extractors
         extractors_dict = defaultdict(list)
-        for extractor_index, frame_index in zip(extractor_indices, frame_idxs):
+        for extractor_index, frame_index in zip(extractor_indices, relative_frame_indices):
             extractors_dict[extractor_index].append(frame_index)
 
         frames_to_concatenate = []
@@ -76,10 +77,8 @@ class MultiImagingExtractor(ImagingExtractor):
         return frames
 
     def _get_frames_from_an_imaging_extractor(self, extractor_index: int, frame_idxs: ArrayType) -> NumpyArray:
-        relative_frame_indices = (np.array(frame_idxs) - self._start_frames[extractor_index]).astype(int)
         imaging_extractor = self._imaging_extractors[extractor_index]
-
-        frames = imaging_extractor.get_frames(frame_idxs=relative_frame_indices)
+        frames = imaging_extractor.get_frames(frame_idxs=frame_idxs)
         return frames
 
     def get_image_size(self) -> Tuple:
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -56,10 +56,7 @@ class MultiImagingExtractor(ImagingExtractor, ABC):
             get_channel_names="The name of the channels",
         )
         for method, property_message in properties_to_check.items():
-            values = [
-                getattr(extractor, method)() if hasattr(extractor, method) else None
-                for extractor in self._imaging_extractors
-            ]
+            values = [getattr(extractor, method, None) for extractor in self._imaging_extractors]
             unique_values = list(set(values))
             assert len(unique_values) == 1, f"{property_message} is not consistent over the files (found {unique_values})."
 
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -33,16 +33,14 @@ class MultiImagingExtractor(ImagingExtractor):
 
         self._start_frames, self._end_frames = [], []
         num_frames = 0
-        times = []
         for imaging_extractor in self._imaging_extractors:
             self._start_frames.append(num_frames)
             num_frames = num_frames + imaging_extractor.get_num_frames()
             self._end_frames.append(num_frames)
-            if getattr(imaging_extractor, "_times") is not None:
-                times.extend(imaging_extractor._times)
         self._num_frames = num_frames
 
-        if times:
+        if any((getattr(imaging_extractor, "_times") is not None for imaging_extractor in self._imaging_extractors)):
+            times = self._get_times()
             self.set_times(times=times)
 
     def _check_consistency_between_imaging_extractors(self):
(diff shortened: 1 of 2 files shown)


src/roiextractors/multiimagingextractor.py
```diff
@@ -56,18 +56,30 @@ def __init__(self, imaging_extractors: list):
...
+        frames_to_concatenate = []
+        # Extract frames for each extractor and concatenate
+        for extractor_index in extractor_indices:
+            frames_for_each_extractor = self._get_frames_from_an_imaging_extractor(
+                extractor_index=extractor_index,
+                frame_idxs=frame_idxs,
+            )
+            frames_to_concatenate.append(frames_for_each_extractor[np.newaxis, ...])
+
+        frames = np.concatenate(frames_to_concatenate, axis=0)
```
weiglszonja, line 70: this is for handling cases when the requested frames belong to different imaging extractors, but I doubt this is the final solution, I'll come back to it tomorrow!

h-mayorquin (commented):
Looks good!

 Curios on how the  `get_frames` method will look like!

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,93 @@
...
+        assert all(isinstance(IX, ImagingExtractor) for IX in imaging_extractors)
+        self._imaging_extractors = imaging_extractors
+
+        # Num channels and sampling frequency based off the initial extractor
+        self._first_imaging_extractor = self._imaging_extractors[0]
+        self._num_channels = self._first_imaging_extractor.get_num_channels()
+        self._channel_names = self._first_imaging_extractor.get_channel_names()
+        self._sampling_frequency = self._first_imaging_extractor.get_sampling_frequency()
+        self._image_size = self._first_imaging_extractor.get_image_size()
+
```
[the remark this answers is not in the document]
weiglszonja: Thanks, good suggestion! Let me know what you think about the implementation
https://github.com/catalystneuro/roiextractors/blob/a400e0c404ae01868eb566453bed9bc56284cfed/src/roiextractors/multiimagingextractor.py#L51-L67

weiglszonja:
Converting this to a draft until unittests are added

src/roiextractors/multiimagingextractor.py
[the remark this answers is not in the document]
CodyCBakerPhD:
```python
getattr(extractor, method)() if hasattr(extractor, method) else None
````

can be reduced to

```python
getattr(extractor, method, None)
```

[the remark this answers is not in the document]
CodyCBakerPhD: Also, to avoid confusion with metadata descriptions of properties may be better to rename the `property_desc` to something like `property_message`, since that's the role it plays in the format string below.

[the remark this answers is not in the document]
CodyCBakerPhD: Also, the `get_num_frames` need not be homogenous over the list of passed ImagingExtractors. Differing segments can be shorter or longer.

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+            get_num_channels="The number of channels",
+            get_channel_names="The name of the channels",
+        )
+        for method, property_desc in properties_to_check.items():
+            values = [
+                getattr(extractor, method)() if hasattr(extractor, method) else None
+                for extractor in self._imaging_extractors
+            ]
+            unique_values = list(set(values))
+            assert len(unique_values) == 1, f"{property_desc} is not consistent over the files {unique_values}"
```
CodyCBakerPhD, line 65:
```suggestion
            assert len(unique_values) == 1, f"{property_desc} is not consistent over the files (found {unique_values})."
```

CodyCBakerPhD: Also when you add the unit tests, make sure to have one that checks the assertion using `hdmf.testing.TestCase; self.assertRaisesWith`.

CodyCBakerPhD:
Overall looking good, will check back when all the unit tests are in.

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,107 @@
...
+from array import ArrayType
+from typing import Tuple, List, Iterable
+
+import numpy as np
+
+from .extraction_tools import NumpyArray
+from .imagingextractor import ImagingExtractor
+
+
+class MultiImagingExtractor(ImagingExtractor, ABC):
```
CodyCBakerPhD, line 11: Class is no longer abstract, can remove `ABC`

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # Checks that properties are consistent between extractors
+        self._check_consistency_between_imaging_extractors()
+
+        # Set properties based off the initial extractor
+        self._first_imaging_extractor = self._imaging_extractors[0]
+        self._num_channels = self._first_imaging_extractor.get_num_channels()
+        self._channel_names = self._first_imaging_extractor.get_channel_names()
+        self._sampling_frequency = self._first_imaging_extractor.get_sampling_frequency()
+        self._image_size = self._first_imaging_extractor.get_image_size()
```
CodyCBakerPhD, line 40: I don't mind much either way, but my general approach to this kind of thing is to reduce the amount of internal allocation in the way of class attributes and instead farm out the API calls respectively.

What I mean by that in this context is, instead of pre-extracting a private attribute such as `self._num_channels` and then having `def get_num_channels` simply return that private attribute (*i.e.*, basically just setting these as [class properties](https://realpython.com/python-property/)), I would personally not set any properties and instead define such methods like

```python
def get_num_channels:
    return self._imaging_extractors[0].get_num_channels()
```

Note this is perfectly safe to do since any instance of this class has passed `self._check_consistency_between_imaging_extractors()`.

This has some advantages such as; if scanning the attributes of an instance of this class then there's fewer items to search through, and it reduces the lines of code needed as well. It also makes the size of an instance of this class smaller, though that's less relevant since these are all near-scalar quantities.

Again, it's not a big deal either way so if you prefer the attribute approach feel free to keep it.

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self._start_frames, self._end_frames = [], []
+        num_frames = 0.0
+        for imaging_extractor in self._imaging_extractors:
+            self._start_frames.append(num_frames)
+            num_frames = num_frames + imaging_extractor.get_num_frames()
+            self._end_frames.append(num_frames)
+
+        self._num_frames = int(num_frames)
```
CodyCBakerPhD, line 49: Just curious, what was the reasoning behind starting the `num_frames` increment at `0.0` and then casting it back at the end? Can't you start it off as `int` and it will stay that way? (if it's overflow you're worried about, it would take an extremely large number of frames to have any effect on an `int32`, the Python default)

src/roiextractors/multiimagingextractor.py
```diff
@@ -0,0 +1,107 @@
...
+                len(unique_values) == 1
+            ), f"{property_message} is not consistent over the files (found {unique_values})."
+
+    def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> NumpyArray:
+        assert max(frame_idxs) < self._num_frames, "'frame_idxs' range beyond number of available frames!"
+        extractor_indices = np.searchsorted(self._end_frames, frame_idxs, side="right")
+        # Match frame_idxs to imaging extractors
+        extractors_dict = {}
+        for extractor_index, frame_index in zip(extractor_indices, frame_idxs):
+            extractors_dict.setdefault(extractor_index, []).append(frame_index)
```
CodyCBakerPhD, line 71: Instead of setting a default manually iteratively, would it be possible to use a `defaultdict(list)` from `collections`?

tests/test_internals/test_multiimagingextractor.py
```diff
@@ -0,0 +1,79 @@
...
+                self.extractors[1].get_frames(frame_idxs=[0, 2, 5]),
+                self.extractors[2].get_frames(frame_idxs=[0, 9]),
+            ),
+            axis=0,
+        )
+        assert_array_equal(self.multi_imaging_extractor.get_frames(frame_idxs=[8, 10, 12, 15, 20, 29]), expected_frames)
+
+    @parameterized.expand(
+        [
+            (3, 4, 15.0, 1, "The sampling frequency is not consistent over the files (found {20.0, 15.0})."),
```
CodyCBakerPhD, line 55: I love the use of `parameterized` for code reduction! However, for enhanced readability, would you mind using a `param`  to specify keyword arguments?

Example: https://github.com/catalystneuro/nwb-conversion-tools/blob/main/tests/test_on_data/test_gin_ecephys.py#L96-L99

CodyCBakerPhD:
Something I had forgotten to do initially with the sub-Imaging stuff, until Heberto reminded me, was to also make sure timestamps work alongside this (when you set times for the individual imaging extractors and then want to use it from the multi extractor). Can you ensure that is working and and add tests for it?

src/roiextractors/multiimagingextractor.py
weiglszonja: I see your point and agree, I'll make the changes. Thank you for all the suggestions!

src/roiextractors/multiimagingextractor.py
```diff
@@ -53,8 +53,8 @@ def _check_consistency_between_imaging_extractors(self):
                 len(unique_values) == 1
             ), f"{property_message} is not consistent over the files (found {unique_values})."
 
-    def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> NumpyArray:
-        assert max(frame_idxs) < self._num_frames, "'frame_idxs' range beyond number of available frames!"
+    @check_get_frames_args
```
weiglszonja, line 56: @CodyCBakerPhD @h-mayorquin 
I was wondering if I could use this decorator here, but I noticed when a single frame is extracted while using this decorator, it only returns the first row of the frame:

https://github.com/catalystneuro/roiextractors/blob/e9080cc53daadb9685222429163dd00564506d9b/src/roiextractors/extraction_tools.py#L289-L290
What was the reasoning behind this? If this is intended, I can go back to do the checks without the decorator, just curious.

CodyCBakerPhD: I haven't been using the decorators for this personally - from past experience, they can produce very hard to debug issues. There's certainly a time and place for them, though, but I'm okay with not doing that decorators here.

CodyCBakerPhD:
Almost there now, just need timestamps with tests

src/roiextractors/multiimagingextractor.py
```diff
@@ -33,12 +33,18 @@ def __init__(self, imaging_extractors: List[ImagingExtractor]):
...
         for imaging_extractor in self._imaging_extractors:
             self._start_frames.append(num_frames)
             num_frames = num_frames + imaging_extractor.get_num_frames()
             self._end_frames.append(num_frames)
+            if getattr(imaging_extractor, "_times") is not None:
+                times.extend(imaging_extractor._times)
         self._num_frames = num_frames
 
+        if times:
+            self.set_times(times=times)
```
weiglszonja, line 46: or do the assertion that the length matches and:
`self._times = times `

CodyCBakerPhD: Hmmm good question, I'll look into it and get back to you

CodyCBakerPhD: Seems from the core logic of https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L71-L88 we may want to pre-scan the list of extractors to see if any one of them has times set (`any((hasattr(extractor, "_times") for extractor in imaging_extractors))`) and if even a single one does, we'll need to assemble an entire `_times` vector for the `MultiImagingExtractor` as well, assemebled by either appending/concatenating the `_times` field if available in a sub extractor, or by getting `frame_to_time` artificially using its sampling frequency.

CodyCBakerPhD:
@weiglszonja For when you're back and feeling better - to save you some time on debugging, it looks like you may have up to update the keyword arguments to the toy example data in the tests. Very minor issue, they should be working otherwise

src/roiextractors/multiimagingextractor.py
```diff
@@ -59,6 +57,17 @@ def _check_consistency_between_imaging_extractors(self):
...
+    def _get_times(self):
+        frame_indices = np.array([*range(self._start_frames[0], self._end_frames[-1])])
+        times = self.frame_to_time(frames=frame_indices)
+
+        for extractor_index, extractor in enumerate(self._imaging_extractors):
+            if getattr(extractor, "_times") is not None:
+                to_replace = [*range(self._start_frames[extractor_index], self._end_frames[extractor_index])]
+                times[to_replace] = extractor._times
+
+        return times
```
weiglszonja, line 69: @CodyCBakerPhD, this is the current attempt to get times based on your recent comment. Let me know what you think about it; I'm not sure this is the best way to do it.

CodyCBakerPhD: Yeah, I'm fine with this! Good job!

The only alternative I can think of is to have an empty list that gets extended with the result of calls to `extractor.frame_to_time(frame_idxs=range(extractor.get_num_frames())` as you iterator over all the `extractors` in `self._imaging_extractors`, and then you cast that as an array at the end - but I don't believe this is much more efficient (if any) than what you have

CodyCBakerPhD (approved):

PR #128, opened 2022-06-03
[Feature] Add frame slicing for imaging

### LINKED ISSUE #103: add stub option
A stub option should limit the conversion to the first ~100 frames so that conversions can be tested quickly without reading/writing all of the data.

CodyCBakerPhD:
fix #103 

WIP

Following general structure of modern approach: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/core/frameslicerecording.py

Need to

- [x] Polish functionality
- [x] Finish unit tests


h-mayorquin:
> @h-mayorquin You may notice this initial PR does not adhere to the upcoming standardization of structure `(frames, rows, cols, channels)`; this was mostly to minimize changes in this PR w.r.t. extraction tools and NumpyImagingExtractor but I'm happy to do it in the next PR

Can you avoid that just by using the new `generate_dummy_imaging_extractor` in your test setup? 
Right now you have a diff again in `src/roiextractors/extractors/numpyextractors/numpyextractors.py` which will revert some changes in master, right?

src/roiextractors/imagingextractor.py
```diff
@@ -147,3 +147,57 @@ def write_imaging(imaging, save_path: PathType, overwrite: bool = False):
...
+            assert 0 <= start_frame < parent_size
+        if end_frame is None:
+            end_frame = parent_size
+        else:
+            assert 0 < end_frame <= parent_size
+        assert end_frame > start_frame, "'start_frame' must be smaller than 'end_frame'!"
+
+        super().__init__()
+
+    def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> np.ndarray:
```
h-mayorquin, line 185: Type annotation. We should do the same here of `Optional[int]` for future functionality of getting complete frames and not a channel.

h-mayorquin (commented):
Two things:
1) Does this return frames from `start_frame` to `end_frame` inclusive. Looking at the test it seems it does not include one of the sides. This should be written in the docstring I think.
2) I think the only missing time functionality is if your `imaging_extractor` already had some times attached to it through a previous call to `set_times`. If that's the case, the times should also be cut to match the new frames appropriately.

CodyCBakerPhD:
> Can you avoid that just by using the new generate_dummy_imaging_extractor in your test setup?

I'll just do that anyway, sure

CodyCBakerPhD:
> Does this return frames from start_frame to end_frame inclusive. Looking at the test it seems it does not include one of the sides. This should be written in the docstring I think.

I'll check against SpikeInterface (which is what I'm basing all this off of) and will get back to you and add it to docstring clarifying the behavior.

> I think the only missing time functionality is if your imaging_extractor already had some times attached to it through a previous call to set_times. If that's the case, the times should also be cut to match the new frames appropriately.

Ah, good catch. I'll add that in a bit

CodyCBakerPhD:
Slicing is exlcusive on the right index, similar to other zero-based Python indexing (*e.g.*, `range(1,5)` gives `0, 1, 2, 3, 4`), and consistent with SpikeInterface.

Added timestamps propagation and test.

h-mayorquin (approved):
Looks good to me. 

Thinking about this in the context of nwb-conversion-tools I was wondering where this would have a repercusion beyond the `add_two_photon_series`:

https://github.com/catalystneuro/nwb-conversion-tools/blob/16126dbd3039444a3c74fa8a371cc43a596b07e5/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py#L162-L165

(Where stubbing could be achieved by taking the first 100 frames in line 184).

The other possible one is `add_epochs` in the same module which supposedly indicates that there is an `imaging._epochs.items()` functionality that I can't find outside of the nwb imaging extractor.  

But if that's supposed to be used in the future this is something that should affect the frame slicing somehow.

CodyCBakerPhD:
`add_epochs` isn't really something we make use of these days.

But frame slicing in general can be used for any number of things from making visual snapshots, trial-aligning, synchronization offsets, etc.

Pretty general feature that gets a lot of use on SpikeInterface.

PR #129, opened 2022-06-04
Add dummy imaging_extractor

h-mayorquin:
For creating test here and in `nwb-conversion-tools` let's mimic spike interface functionality to add simple artificial dummy extractors that are easy to create and test. This PR pushes one for `imaging_extractors`. A similar one will be added for the homologous segmentation structure in a further PR. 

Also, I changed the numpy extractor to conform to the now agreed upon image format of (frames, rows, columns, channels).

### DIFF
--- a/src/roiextractors/extractors/numpyextractors/numpyextractors.py
+++ b/src/roiextractors/extractors/numpyextractors/numpyextractors.py
@@ -50,10 +50,10 @@ class NumpyImagingExtractor(ImagingExtractor):
         self._channel_names = channel_names
 
         (
-            self._num_channels,
             self._num_frames,
             self._size_x,
             self._size_y,
+            self._num_channels,
         ) = get_video_shape(self._video)
 
         if len(self._video.shape) == 3:
--- a/src/roiextractors/testing.py
+++ b/src/roiextractors/testing.py
@@ -14,18 +14,30 @@ inttype = (int, np.integer)
 import numpy as np
 
 from roiextractors import NumpyImagingExtractor
+from roiextractors.extraction_tools import DtypeType
 
 
 def generate_dummy_imaging_extractor(
-    num_frames: int = 100,
+    num_frames: int = 30,
     rows: int = 10,
     columns: int = 10,
     num_channels: int = 1,
-    sampling_frequency: float = 30000,
-    dtype="int16",
+    sampling_frequency: float = 30,
+    dtype: DtypeType = "uint16",
 ):
     channel_names = [f"channel_num_{num}" for num in range(num_channels)]
-    video = np.random.randint(low=0, high=256, size=(num_frames, rows, columns, num_channels)).astype(dtype)
+
+    dtype = np.dtype(dtype)
+    number_of_bytes = dtype.itemsize
+
+    low = 0 if "u" in dtype.name else 2 ** (number_of_bytes - 1) - 2**number_of_bytes
+    high = 2**number_of_bytes - 1 if "u" in dtype.name else 2**number_of_bytes - 2 ** (number_of_bytes - 1) - 1
+    size = (num_frames, rows, columns, num_channels)
+    video = (
+        np.random.random(size=size)
+        if "float" in dtype.name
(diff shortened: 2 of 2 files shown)


src/roiextractors/extractors/numpyextractors/numpyextractors.py
```diff
@@ -50,10 +50,10 @@ def __init__(
         self._channel_names = channel_names
 
         (
-            self._num_channels,
             self._num_frames,
             self._size_x,
             self._size_y,
+            self._num_channels,
```
CodyCBakerPhD, line 56: Thanks, do you know which other extractors might need this kind of change?

src/roiextractors/testing.py
```diff
@@ -11,6 +11,29 @@
 inttype = (int, np.integer)
 
 
+import numpy as np
+
+from roiextractors import NumpyImagingExtractor
+
+
+def generate_dummy_imaging_extractor(
+    num_frames: int = 100,
```
CodyCBakerPhD, line 20: In the spirit of keeping an instance of an object generated via this method using only defaults as minimal as possible in order to improve testing speeds, what would you think of `30` frames instead? That's 1 full second of imaging sampled at 30Hz (somewhat average for this).

I know you can of course always override this default, but having to do so is an extra step that is easy to forget and I can't think of many cases that would require a longer video for testing purposes.

src/roiextractors/testing.py
```diff
@@ -11,6 +11,29 @@
...
+    rows: int = 10,
+    columns: int = 10,
+    num_channels: int = 1,
+    sampling_frequency: float = 30000,
```
CodyCBakerPhD, line 24: The average imaging rate of data we've seen is around 30Hz, not 30 kHz.

src/roiextractors/testing.py
```diff
@@ -11,6 +11,29 @@
...
+    dtype="int16",
```
CodyCBakerPhD, line 25: Our data is usually unsigned as well - though, yes, it is often `uint16` instead of `uint8` (mostly because it is usually grayscale).

CodyCBakerPhD: Also can you add annotation typing? Either strict string input, or union of string and numpy dtype?

src/roiextractors/extractors/numpyextractors/numpyextractors.py
h-mayorquin: Most of them. I will address this in a future PR.

src/roiextractors/testing.py
h-mayorquin: Totally onboard with this. Changing.

src/roiextractors/testing.py
h-mayorquin: Yup!

src/roiextractors/testing.py
```diff
@@ -11,6 +11,29 @@
...
+):
+    channel_names = [f"channel_num_{num}" for num in range(num_channels)]
+    video = np.random.randint(low=0, high=256, size=(num_frames, rows, columns, num_channels)).astype(dtype)
```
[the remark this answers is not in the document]
h-mayorquin: I think that's a good idea that, the only thing that I am not so sure about is the random values for the float case but I don't really have any other suggestion.

src/roiextractors/testing.py
h-mayorquin: Done.

src/roiextractors/testing.py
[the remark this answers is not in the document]
CodyCBakerPhD: Oh I meant to mention, the random values in the float case are uniformly distrbuted which is good because then if someone wants to run some kind of targetted test on specific ranges of data values (which onyl they would know) they can easily to dhat by shift/scale modification of the default uniform values.

*I.e.*, `video_with_other_values = video * (b - a) + a` will now uniformly range any interval `[a, b]`.

Of course if they want a different distribution altogether, then they can either just make their own toy data or use the inverse transform method `video_with_new_distribution = inverse_cumulative_distribution(video)`.

Issue #121, opened 2022-05-26
Standardize frame and image terminology and their related methods in `ImagingExtractor`


h-mayorquin:
##TL;DR 
In videos with multiple channels (three colors) the terms **image** and **frame** can mean an object with shape `(n_channels, rows, columns)` or an object with shape `(row, columns)`. I think we should decided which one is which and enforce their related methods to return objects with that shape by default.

Particularly, I think we should change `get_frames` to return an object of size  `(n_channels, rows, columns)` as now it mostly (as, in most of the methods in the instance classes) returns objects of size `(row, columns)`.

Here, I don't mind about the order, just that the n_channels dimension should be there.
## Context
The base class for imaging extractor has the following abstract methods:

https://github.com/catalystneuro/roiextractors/blob/35accb9c6b25838dfe209cd7ea626b99518d9120/src/roiextractors/imagingextractor.py#L29-L35

Looking at the first method (there is a channel in the argument) it seems that the original intention was for frame to mean an object of size `(row, columns)` and some of the tests that I recently silenced are consistent with this assumption:

https://github.com/catalystneuro/roiextractors/blob/35accb9c6b25838dfe209cd7ea626b99518d9120/src/roiextractors/testing.py#L181-L188

This has not been a problem so far because all of our `io` tests only have data with one channel but the recently added `NumpyMemmapImagingExtractor` method can handle videos with three channels and the test above was supprsed. 

My personal preference is for frame to mean  `(n_channels, rows, columns)`  and that image means `(rows, columns)`  (you can see that that's a [common understanding](https://stackoverflow.com/questions/13033278/image-size-python-opencv) of the word and that `openCV` used to have like that). As mentioned above, following this convention would mean that `get_frames` should return objects of size  `(n_channels, rows, columns)` by default which is a small change in most of the classes. 

That said, maybe there is a better terminology or something that I am missing.

CodyCBakerPhD:
To be clear though, only the capability of the `NumpyMemmapImagingExtractor` involves multiple channels. In practice, these data types are most often grayscale (I haven't yet seen any true Imaging data that is in color), including the current Seidemann project that uses this binary format.

So another option may be just for the moment to raise `NotImplementedErrors` for when there are more more than one color channel.

h-mayorquin:
@CodyCBakerPhD 

I see, so we should expect to see all the future formats to only have a gray channel?

[...]

PR #130, opened 2022-06-04
Add channel argument to preserve backwards compability in `memmapextractors`

h-mayorquin:
I am adding this argument to keep backwards compatibility with the previous `get_frames` policy of returning only a channel of the signal (see #121 for a more general discussion).

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -35,7 +35,7 @@ except ImportError:
 
 ArrayType = Union[list, np.array]
 PathType = Union[str, Path]
-NumpyArray = Union[np.ndarray, np.memmap]
+NumpyArray = np.ndarray
 DtypeType = Union[str, np.dtype]
 IntType = Union[int, np.integer]
 FloatType = float
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -35,7 +35,7 @@ except ImportError:
 
 ArrayType = Union[list, np.array]
 PathType = Union[str, Path]
-NumpyArray = Union[np.array, np.memmap]
+NumpyArray = Union[np.ndarray, np.memmap]
 DtypeType = Union[str, np.dtype]
 IntType = Union[int, np.integer]
 FloatType = float
--- a/src/roiextractors/extractors/memmapextractors/memmapextractors.py
+++ b/src/roiextractors/extractors/memmapextractors/memmapextractors.py
@@ -21,14 +21,14 @@ class MemmapImagingExtractor(ImagingExtractor):
     def __init__(
         self,
         video,
-    ):
+    ) -> None:
         """
         Abstract class for memmapable imaging extractors.
         """
         self._video = video
         super().__init__()
 
-    def get_frames(self, frame_idxs=None, channel: Optional[int] = 0) -> NumpyArray:
+    def get_frames(self, frame_idxs=None, channel: Optional[int] = 0) -> np.ndarray:
         if frame_idxs is None:
             frame_idxs = [frame for frame in range(self.get_num_frames())]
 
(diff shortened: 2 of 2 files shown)


src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -27,17 +27,19 @@ def __init__(
         self._video = video
         super().__init__()
 
-    def get_frames(self, frame_idxs=None):
+    def get_frames(self, frame_idxs=None, channel: int = 0):
```
CodyCBakerPhD, line 30: If `channel` can be `None` as an allowed input, add this to the annotation type as `Optional[int]`

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -27,17 +27,19 @@ def __init__(
...
             frame_idxs = [frame for frame in range(self.get_num_frames())]
 
         frames = self._video.take(indices=frame_idxs, axis=0)
+        if channel is not None:
+            frames = frames[:, :, :, channel].squeeze()
 
         return frames
 
-    def get_video(self, start_frame: int = None, end_frame: int = None) -> np.array:
+    def get_video(self, start_frame: int = None, end_frame: int = None, channel: int = 0) -> np.array:
```
CodyCBakerPhD, line 40: Since this is getting handed off, essentially, to `get_frames`, mimic requested annotation changes here as well

CodyCBakerPhD: Also, recommend annotation typing `np.array` as `np.ndarray` - interesting read: https://stackoverflow.com/questions/35673895/type-hinting-annotation-pep-484-for-numpy-ndarray

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Good idea!

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -27,17 +27,19 @@ def __init__(
         self._video = video
         super().__init__()
 
-    def get_frames(self, frame_idxs=None):
+    def get_frames(self, frame_idxs=None, channel: int = 0):
         if frame_idxs is None:
             frame_idxs = [frame for frame in range(self.get_num_frames())]
 
         frames = self._video.take(indices=frame_idxs, axis=0)
+        if channel is not None:
```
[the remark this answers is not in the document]
h-mayorquin: As far as I can see they don't. The functional always return a specific channel.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
h-mayorquin: Good. I took the chance to add a bunch of typing information together with this. Check the latest commit.

src/roiextractors/extractors/memmapextractors/memmapextractors.py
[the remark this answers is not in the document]
CodyCBakerPhD: I kind of like this behavior, though. Can you add a note to refactor this throughout?

It also implies that the `channel` annotation could reasonably be extended to `Optional[Union[int, List[int]]]` if, for whatever reason, someone only wanted to subset two of three color channels.

src/roiextractors/extraction_tools.py
```diff
@@ -35,7 +35,7 @@
 
 ArrayType = Union[list, np.array]
 PathType = Union[str, Path]
-NumpyArray = Union[np.array, np.memmap]
+NumpyArray = Union[np.ndarray, np.memmap]
```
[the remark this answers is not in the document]
h-mayorquin: Good. I will open a small PR to correct this everywhere else then.

CodyCBakerPhD (approved):

PR #131, opened 2022-06-08
Update numpy type hints

h-mayorquin:
A small PR with some improvements in PR type hints that uses the new type hinting module for `numpy`:

https://numpy.org/doc/stable/user/basics.types.html

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -4,9 +4,11 @@ from typing import Union, Tuple
 from dataclasses import dataclass, field
 
 import numpy as np
+from numpy.typing import ArrayLike, DTypeLike
 import scipy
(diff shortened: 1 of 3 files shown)


src/roiextractors/extraction_tools.py
```diff
@@ -24,10 +26,9 @@
 except ImportError:
     HAVE_Scipy = False
 
-ArrayType = Union[list, np.array]
+ArrayType = ArrayLike
 PathType = Union[str, Path]
-NumpyArray = Union[np.array, np.memmap]
-DtypeType = Union[str, np.dtype]
+DtypeType = DTypeLike
```
CodyCBakerPhD, line 31: Thanks, I prefer this approach much more.

CodyCBakerPhD (approved):

PR #132, opened 2022-06-08
[Refactor] standardize video structures

CodyCBakerPhD:
@h-mayorquin First attempts to start the process of standardizing internal tools and data access patterns to the new standard.

For the ImagingExtractor inits, however, careful attention must be given to the source data structure (which may not be our 'canonical' form). As such, we may want to start using the `VideoStructure` to resolve the mapping for all these cases.

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -260,14 +260,19 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
     return image_mask
 
 
-def get_video_shape(video):
-    if len(video.shape) == 3:
-        # 1 channel
+def get_video_shape(video) -> Union[Tuple[int, int, int], Tuple[int, int, int, int]]:
+    """Return the shape of the video according to the standardized structure of (frames, rows, columns, channels)."""
+    n_dims = len(video.shape)
+    if n_dims == 3:
         num_channels = 1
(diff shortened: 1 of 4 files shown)


h-mayorquin:
> @h-mayorquin First attempts to start the process of standardizing internal tools and data access patterns to the new standard.
> 
> For the ImagingExtractor inits, however, careful attention must be given to the source data structure (which may not be our 'canonical' form). As such, we may want to start using the `VideoStructure` to resolve the mapping for all these cases.

I was thinking that we could do this piece wise by adding a general flag to the base class of `ImagingExtractor`. Something like `self.standard_format=False` (or something shorter). Then we add the flag at the `__init__` of the classes where we can promise that as True and add a general test that goes over all the classes that have this flag and test they do uphold their contract. 

The advantage of this is that we have some explicit condition for the contract and also we can just leave it false if in the future we encounter some interface that is hard to reshape.

h-mayorquin:
See #139

CodyCBakerPhD:
I don't think anything is actually gained by having this: https://github.com/catalystneuro/roiextractors/pull/132/files#diff-a78deab6d95f9b7d33f492bee43ebed5cb1f38d1dfe788f67c56e52936abe90dR263-R275 as a function unto itself.

I saw we configure the internals for each format specifically (not all of them even have arrays per se) and as long as the API produces the output in a consistent way we should be fine

PR #136, opened 2022-06-10
Refactor nwb extractor frames and tests

h-mayorquin:
This is the first attempt at getting correct frames in `nwb_extractor`. This branch should be tested in `nwb-conversion-tools` to see if everything works as we expected over there.  I refactored the function and added a test to see that it works as expected.

I intend to add a test to support loading `TwoPhotonTimeSeries` with channels but that's not the priority right now.

### DIFF
--- a/src/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/src/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -83,14 +83,12 @@ class NwbImagingExtractor(ImagingExtractor):
         # Load the two video structures that TwoPhotonSeries supports.
         self._data_has_channels_axis = True
         if len(self.two_photon_series.data.shape) == 3:
-            self._data_has_channels_axis = False
             self._num_channels = 1
-            self._num_frames, self._rows, self._columns = self.two_photon_series.data.shape
+            self._num_frames, self._columns, self._rows = self.two_photon_series.data.shape
         else:
-            self._num_frames, self._rows, self._columns, self._num_channels = self.two_photon_series.data.shape
-            raise "videos with multiple channel not supported yet"
+            raise "TwoPhothonSeries with depth dimension (z) not supported"
 
-        # Set channel names
+        # Set channel names (This should disambiguate which optical channel)
         self._channel_names = [i.name for i in self.two_photon_series.imaging_plane.optical_channel]
 
         # Set sampling frequency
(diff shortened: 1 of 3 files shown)


h-mayorquin:
This is the PR in `nwb-conversion-tools`:

https://github.com/catalystneuro/nwb-conversion-tools/pull/547

tests/test_internals/test_nwb_extractors.py
```diff
@@ -0,0 +1,109 @@
...
+
+        expected_image_size = self.image_size
+        expected_num_frames = self.num_frames
+        expected_num_channels = self.num_channels
+
+        assert image_size == expected_image_size
+        assert num_frames == expected_num_frames
+        assert num_channels == expected_num_channels
+
+        frame_idxs = 0
```
h-mayorquin, line 87: Maybe this case does not make sense, I am not sure if we should allow `frame_idx` take scalar values instead of list or arrays.

h-mayorquin: I just checked and `TiffImagingExtractor` does the same. In a second thought my current take is that  the argument frame_idxs of `get_frames` should behave exactly like slicing of `numpy` to the extent that is possible so users use a mental model that is readily available within the scientific community. Moreover, if we do get that we can abstract this away with [__getitem__ ](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) to support slicing notation.

CodyCBakerPhD: > Maybe this case does not make sense, I am not sure if we should allow frame_idx take scalar values instead of list or arrays.

It's mostly because `frame_idxs=[single_idx]` is more awkward than `frame_idxs=single_idx`.

CodyCBakerPhD: > `get_frames` should behave exactly like slicing of numpy

Right, and you can slice a numpy array with a single scalar as well as an `ArrayType` of multiple indexes.

src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -139,24 +150,33 @@ def make_nwb_metadata(self, nwbfile, opts):
...
+            slice_stop = min(np.max(frame_idxs) + 1, self.get_num_frames())
         else:
-            sorted_idxs = np.sort(frame_idxs)
-            argsorted_idxs = np.argsort(frame_idxs)
-            return opts.data[sorted_idxs][argsorted_idxs].transpose([0, 2, 1])
-
-    @check_get_videos_args
-    def get_video(self, start_frame=None, end_frame=None, channel=0):
-        opts = self.nwbfile.acquisition[self._optical_series_name]
-        video = opts.data[start_frame:end_frame].transpose([0, 2, 1])
```
CodyCBakerPhD, line 155: Turns out this un-transposition is actually correct - can you adjust that and the tests and then we can call this good?

h-mayorquin:
@CodyCBakerPhD 
I corrected to take into account the transposition and the fact that the `TwoPhotonSeries` does not really have channels as per the discussion in:
https://github.com/NeurodataWithoutBorders/nwb-schema/issues/522

Also, I realized that most of the other `get_frames` functions do not behave like numpy when indexed with a singleton. I will open another issue for that.

tests/test_internals/test_nwb_extractors.py
h-mayorquin: See:
https://github.com/catalystneuro/roiextractors/issues/147

CodyCBakerPhD (approved):

PR #139, opened 2022-06-11
Proposal for standardizing data format piecewise.

h-mayorquin:
Following the comment in #132:

> I was thinking that we could do this piece wise by adding a general flag to the base class of ImagingExtractor. Something like self.standard_format=False (or something shorter). Then we add the flag at the __init__ of the classes where we can promise that as True and add a general test that goes over all the classes that have this flag and test they do uphold their contract.

> The advantage of this is that we have some explicit condition for the contract and also we can just leave it false if in the future we encounter some interface that is hard to reshape.

Here is the proposal in a PR for this. I will do a chain PR to this for `Hdf5ImagingExtractor`.

### DIFF
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -25,6 +25,7 @@ class ImagingExtractor(ABC, BaseExtractor):
         BaseExtractor.__init__(self)
         assert self.installed, self.installation_mesg
         self._memmapped = False
+        self._images_in_standard_structure = False
 
     @abstractmethod
     def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


src/roiextractors/imagingextractor.py
```diff
@@ -68,6 +69,21 @@ def get_dtype(self) -> DtypeType:
...
+        if unique_frame_value == rows:
+            unique_frame_value += 1
+        if unique_frame_value == columns:
+            unique_frame_value += 1
+
+        frame_idxs = [_ for _ in range(unique_frame_value)]
+        video_shape = np.array(self.get_frames(frame_idxs=frame_idxs, channel=None).shape)
+        video_shape[video_shape == unique_frame_value] = self.get_num_frames()
+
+        return tuple(video_shape)
```
CodyCBakerPhD, line 85: This function is rather strange...

I'd rather have the shape information handled properly by whatever the specific format is so that our existing API functions simply 'handle it' in the same way each time

CodyCBakerPhD:
@h-mayorquin Check https://github.com/catalystneuro/roiextractors/pull/146 to see what is basically the tests you added here, but they already work. No need for attribute tracking what is enabled/disabled for this.

src/roiextractors/imagingextractor.py
h-mayorquin: It is written like this to avoid loading as few frames as necessary (for memory concerns). Maybe there is a better way to do it though.

CodyCBakerPhD: Hence why reliance on the format-specific API functions (which should be built lazily as is) would be both simpler and just as safe memory-wise.

See https://github.com/catalystneuro/roiextractors/pull/140#discussion_r901166368

PR #140, opened 2022-06-11
Hdf5 to standard video format

h-mayorquin:
This ensures that the `hdf5` now has two things:
1) The option to use `get_frames` with channel=None and it should return the files in the standard structure.
2) Can use a function `get_video_shape` that uses the former functionality to return the shape of the video.

Also, the extractor now uses lazy operations for loading through `lazy_ops`.

### DIFF
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -37,7 +37,6 @@ class Hdf5ImagingExtractor(ImagingExtractor):
         channel_names: ArrayType = None,
     ):
         ImagingExtractor.__init__(self)
-        self._images_in_standard_structure = True
 
         self.filepath = Path(file_path)
         self._sampling_frequency = sampling_frequency
@@ -71,19 +70,9 @@ class Hdf5ImagingExtractor(ImagingExtractor):
         else:
             self.metadata = metadata
 
-        (
-            self._num_channels,
-            self._num_frames,
-            self._rows,
-            self._columns,
-        ) = get_video_shape(self._video)
-
-        # I am assuming that the videos come here with self._num_channels in the first axis
-        self._data_has_channels_axis = self._video.ndim > 3
-        if self._data_has_channels_axis:
-            self._standard_video = self._video.lazy_transpose([1, 2, 3, 0])
-        else:
-            self._standard_video = self._video
+        # The test data has four dimensions and the first axis is channels
+        self._num_channels, self._num_frames, self._num_rows, self._num_cols = self._video.shape
+        self._video = self._video.lazy_transpose([1, 2, 3, 0])
 
         if self._channel_names is not None:
             assert len(self._channel_names) == self._num_channels, (
@@ -112,19 +101,11 @@ class Hdf5ImagingExtractor(ImagingExtractor):
             slice_start = 0
             slice_stop = self.get_num_frames()
 
-        if self._data_has_channels_axis and channel is not None:
-            frames = self._standard_video.lazy_slice[slice_start:slice_stop, :, :, channel]
-        elif self._data_has_channels_axis and channel is None:
-            frames = self._standard_video.lazy_slice[slice_start:slice_stop, ...]
-        elif not self.data_has_channel_axis and channel is None:
-            frames = self._standard_video.lazy_slice[slice_start:slice_stop, ...][:, np.newaxis]
-        else:
-            frames = self._standard_video.lazy_slice[slice_start:slice_stop, ...]
-
+        frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel]
         return frames
 
     def get_image_size(self) -> Tuple[int, int]:
-        return (self._rows, self._columns)
+        return (self._num_rows, self._num_cols)
 
     def get_num_frames(self):
         return self._num_frames
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -1,15 +1,16 @@
 from pathlib import Path
+from typing import Optional, Tuple
 
 import numpy as np
 
 from ...extraction_tools import PathType, FloatType, ArrayType
 from ...extraction_tools import (
-    check_get_frames_args,
     get_video_shape,
     write_to_h5_dataset_format,
 )
 from ...imagingextractor import ImagingExtractor
-from typing import Tuple
+from lazy_ops import DatasetView
+
 
 try:
     import h5py
@@ -36,6 +37,8 @@ class Hdf5ImagingExtractor(ImagingExtractor):
         channel_names: ArrayType = None,
     ):
         ImagingExtractor.__init__(self)
+        self._images_in_standard_structure = True
+
         self.filepath = Path(file_path)
         self._sampling_frequency = sampling_frequency
         self._mov_field = mov_field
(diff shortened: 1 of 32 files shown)


src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -71,13 +74,16 @@ def __init__(
         (
             self._num_channels,
             self._num_frames,
-            self._size_x,
-            self._size_y,
+            self._rows,
+            self._columns,
         ) = get_video_shape(self._video)
```
CodyCBakerPhD, line 79: `get_video_shape` doesn't really do much TBH

How about just doing the aliasing/assignment as

```python
self._num_channels, self._num_frames, self._num_rows, self._num_columns = self._video.shape
```

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -71,13 +74,16 @@ def __init__(
...
 
-        if len(self._video.shape) == 3:
-            # check if this converts to np.ndarray
-            self._video = self._video[np.newaxis, :]
+        # I am assuming that the videos come here with self._num_channels in the first axis
+        self._data_has_channels_axis = self._video.ndim > 3
+        if self._data_has_channels_axis:
+            self._standard_video = self._video.lazy_transpose([1, 2, 3, 0])
```
CodyCBakerPhD, line 84: This sort of map could be nice, but I would avoid having both `_video` and `_standard_video`. Perhaps define a strictly internal `initial_video` then define `self._video = initial_video[proper_axis_order]`?

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -36,6 +37,8 @@ def __init__(
         channel_names: ArrayType = None,
     ):
         ImagingExtractor.__init__(self)
+        self._images_in_standard_structure = True
```
CodyCBakerPhD, line 40: The images will return the proper structure, so ordering items are strictly for internal/private usage, so IDK that we need to track it

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -96,20 +102,29 @@ def __init__(
...
+        else:
+            slice_start = 0
+            slice_stop = self.get_num_frames()
+
+        if self._data_has_channels_axis and channel is not None:
+            frames = self._standard_video.lazy_slice[slice_start:slice_stop, :, :, channel]
+        elif self._data_has_channels_axis and channel is None:
+            frames = self._standard_video.lazy_slice[slice_start:slice_stop, ...]
+        elif not self.data_has_channel_axis and channel is None:
+            frames = self._standard_video.lazy_slice[slice_start:slice_stop, ...][:, np.newaxis]
```
CodyCBakerPhD, line 120: This is a lot of detail just for handling a case that isn't often seen. Is this tested in any way right now?

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
CodyCBakerPhD: > You want different labels/variable names for them or am I missing something?

I don't see a point to having both views as class attributes, better to just have one attribute that is the expected shape (the `self._standard_video` here, but named `self._video` since no need to adjust the name to differentiate).

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
CodyCBakerPhD: We wouldn't want anything downstream to use a class flag, even a private one, that is only expected to be temporary.

src/roiextractors/imagingextractor.py
```diff
@@ -73,7 +73,19 @@ def get_video(self, start_frame: int = None, end_frame: int = None, channel: int
...
+        if unique_frame_value == rows:
+            unique_frame_value += 1
+        if unique_frame_value == columns:
+            unique_frame_value += 1
+
+        frame_idxs = [_ for _ in range(unique_frame_value)]
+        video_shape = np.array(self.get_frames(frame_idxs=frame_idxs, channel=None).shape)
+        video_shape[video_shape == unique_frame_value] = self.get_num_frames()
+
+        return tuple(video_shape)
```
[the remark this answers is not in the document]
CodyCBakerPhD: If you mean a simple collection helper like

```python
def get_video_size(self)
    image_size = self.get_image_size()
    num_channels = self.get_num_channels()
    if num_channels == 1:
        return (self.get_num_frames(), image_size[0], image_size[1])
    else:
        return (self.get_num_frames(), image_size[0], image_size[1], num_channels)
```

That would be fine.

Note that this also has the function name ending in `_size` not `_shape` to be consistent with the sister function `get_image_size`.

I would also point out that `_shape` would be more appropriate in both situations, so we should discuss if that should also be standardized.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
CodyCBakerPhD: Should either (a) add some tests or (b) add a `NotImplementedError` for when `channel is not None` since that isn't something the rest of the package really supports yet.

CodyCBakerPhD: Personally favoring option (b) since we have no examples anywhere of multi-channel data. We could build tests for HDF5 and numpy extractors with synthetic data, but it's not really an important feature and we have other higher priority things across the ecosystem right now.

src/roiextractors/imagingextractor.py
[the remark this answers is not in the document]
CodyCBakerPhD: The most general thing we could do is for the `__init__` of every format to properly coerce the shape of each object into a standard API, and everything should follow from that. Just not something that was double checked until now.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: All right, I get you and yes, I think that's better. Just having an internal variable that is already on the standard shape is better. Not point in keeping a point to the original form.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Btw, do you know where this assumption comes from? that videos in this format come in that shape? is there a specification?

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Done.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Done.

src/roiextractors/imagingextractor.py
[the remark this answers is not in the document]
h-mayorquin: Sure. Let's make it implicit.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Eliminated the flags.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: So I implemented this.

CodyCBakerPhD:
@h-mayorquin Looks good for now - in a follow-up PR, though, can you optionally allow a `VideoStructure` to be attached to resolve the mapping of these two lines? https://github.com/catalystneuro/roiextractors/pull/140/files#diff-3b502517c20d13549ed923b67c19415af11532255bfaaec1393d7ad7be39514dR73-R74

which are not, per se, a part of any proprietary format but rather are merely a convention used for that one demo file?

CodyCBakerPhD (approved):

PR #141, opened 2022-06-11
add typing information to `frame_slice in base` class

h-mayorquin:
This was missing in the frame slice PR. The class has the possibility to set start_frame and end_frame to None but not the function in the base class.

### DIFF
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -128,7 +128,7 @@ class ImagingExtractor(ABC, BaseExtractor):
         if extractor._times is not None:
             self.set_times(deepcopy(extractor._times))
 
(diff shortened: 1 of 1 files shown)


src/roiextractors/imagingextractor.py
```diff
@@ -128,7 +128,7 @@ def copy_times(self, extractor: BaseExtractor) -> None:
         if extractor._times is not None:
             self.set_times(deepcopy(extractor._times))
 
-    def frame_slice(self, start_frame, end_frame):
+    def frame_slice(self, start_frame: int = None, end_frame: int = None):
```
CodyCBakerPhD, line 131:
```python
Optional[int]
```

h-mayorquin: Yes. Done.

CodyCBakerPhD (approved):

PR #144, opened 2022-06-14
ScanImageTiffExtractor

CodyCBakerPhD:
WIP adding a new extractor specifically for reading TIFF files produced via the ScanImage software.


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -0,0 +1,96 @@
...
+            self._num_channels = 1
+        else:  # no example file for multiple color channels or depths
+            raise NotImplementedError(
+                "Extractor cannot handle 4D TIFF data. Please raise an issue to request this feature: "
+                "https://github.com/catalystneuro/roiextractors/issues "
+            )
+
+    @check_get_frames_args
+    def get_frames(self, frame_idxs, channel=0) -> np.ndarray:
+        if not all(np.diff(frame_idxs) == 0):
```
bendichter, line 78:
```suggestion
        if not all(np.diff(frame_idxs) == 1):
```
shouldn't this be 1?

CodyCBakerPhD: Yes

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        assert self.installed, self.installation_mesg
+        super().__init__()
+        self.file_path = Path(file_path)
+        self._sampling_frequency = sampling_frequency
+        assert self.file_path.suffix in [".tiff", ".tif", ".TIFF", ".TIF"]
+
+        with ScanImageTiffReader(str(self.file_path)) as io:
+            shape = io.shape()  # [frames, rows, cols]
+        if len(shape) == 3:
+            self._num_frames, self._num_rows, self._num_cols = shape
```
h-mayorquin, line 64: We should either use `rows` and `columns` here as well or change `_num_rows` and `_num_cols` in other extractors.

h-mayorquin (commented):
Maybe we should create a directory called `tests_on_data` to mimic nwb-conversion-tools.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+    def get_image_size(self) -> Tuple[int, int]:
+        return (self._num_rows, self._num_cols)
+
+    def get_num_frames(self) -> int:
+        return self._num_frames
+
+    def get_sampling_frequency(self) -> float:
+        return self._sampling_frequency
+
+    def get_channel_names(self) -> Optional[ArrayType]:
```
bendichter, line 98: It seems odd to me that the return type here is `Optional`. Does this mean that it's either `None` or an `ArrayType`?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        file_path: PathType,
+        sampling_frequency: FloatType,
+    ):
+        """
+        Specialized extractor for reading TIFF files produced via ScanImage.
+
+        The generic TiffImagingExtractor has issues loading these types of TIFF files in a lazy (memory mapped) manner.
+
+        The API calls to this extractor, however, have fully lazy behavior; however, direct slicing of the underlying
+        data structure is not equivalent to a numpy memory map.
```
bendichter, line 46:
```suggestion
        This extractor allows for lazy accessing of slices, unlike :py:class:`~roiextractors.extractors.tiffimagingextractors.TiffImagingExtractor`. However, direct slicing of the underlying data structure is not equivalent to a numpy memory map.
```

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -0,0 +1,102 @@
...
+        file_path : PathType
+            Path to the TIFF file.
+        sampling_frequency : float
+            The frequency at which the frames were sampled, in Hz.
+        """
+        assert self.installed, self.installation_mesg
+        super().__init__()
+        self.file_path = Path(file_path)
+        self._sampling_frequency = sampling_frequency
+        assert self.file_path.suffix in [".tiff", ".tif", ".TIFF", ".TIF"]
```
bendichter, line 59: I'd probably raise a warning here instead. It is possible that a tiff file might have a different suffix.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Since the other arguments are, if present, `_num_frames` and `_num_channels`, then I think we should transition to `_num_rows` and `_num_cols`.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Channel names aren't actually used here, so removing this function (which doesn't currently do anything anyway).

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Done

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Fixed the annotation to match the base class, but this extractor only has a single unnamed channel so this function isn't actually used, it just has to be defined here since it's set as an abstract method in the base.

CodyCBakerPhD:
> Maybe we should create a directory called tests_on_data to mimic nwb-conversion-tools.

Not a bad idea now that we finally have a few tests that don't require data. This PR does, though.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Yes, I think this is more consistent. Let's change it on all the others.

PR #145, opened 2022-06-16
MultiImagingExtractor to enable querying different slices of frames

weiglszonja:
This PR directly relates to #127, the reason why it is in a separate PR is to enhance the visibility of the changes in the `get_frames` method. 
The aim of this PR is to enable querying different slices of frames e.g.
`[[0,1,2,3], [10,11,22,25], [...]] `  just as it should work with numpy arrays. 

Is this use case justified?

### DIFF
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -53,33 +53,46 @@ class MultiImagingExtractor(ImagingExtractor):
                 len(unique_values) == 1
             ), f"{property_message} is not consistent over the files (found {unique_values})."
 
-    @check_get_frames_args
     def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> NumpyArray:
-        extractor_indices = np.searchsorted(self._end_frames, frame_idxs, side="right")
-        # Match frame_idxs to imaging extractors
-        extractors_dict = defaultdict(list)
-        for extractor_index, frame_index in zip(extractor_indices, frame_idxs):
-            extractors_dict[extractor_index].append(frame_index)
+        if isinstance(frame_idxs, (int, np.integer)):
+            frame_idxs = [frame_idxs]
+        frame_idxs = np.array(frame_idxs)
+        assert np.max(frame_idxs) < self._num_frames, "'frame_idxs' range beyond number of available frames!"
+        extractors = np.searchsorted(self._end_frames, frame_idxs, side="right")
 
-        frames_to_concatenate = []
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@weiglszonja It would be weird to do this just for the MultiImagingExtractor and no other ImagingExtractor. But likewise seems like a lot of setup to do for all of the extractors.

Is it not possible and just as simple to iterate over the double indices and call `get_frames(frame_idxs=single_slice_of_frame_idx)`?

Also, what is the intended use case for this return structure as opposed to just flattening the frame idx list?

weiglszonja:
Thanks @CodyCBakerPhD, you're right. I was trying to cover all edge-cases like what if the frame idx list is actually a list of lists, how it should be handled then. I probably overthought it.

CodyCBakerPhD:
> Thanks @CodyCBakerPhD, you're right. I was trying to cover all edge-cases like what if the frame idx list is actually a list of lists, how it should be handled then. I probably overthought it.

Thanks for considering it, advanced features like that are always cool and we can keep this in the backlog for the far future or if/when an analysis use case pops up for it. But yeah, every little bit of added complexity makes projects like this harder to maintain so for now we're just trying to make everything in the repo as basic as possible to do it's job.

src/roiextractors/multiimagingextractor.py
```diff
@@ -53,17 +53,23 @@ def _check_consistency_between_imaging_extractors(self):
...
-        extractor_indices = np.searchsorted(self._end_frames, frame_idxs, side="right")
-        # Match frame_idxs to imaging extractors
-        extractors_dict = defaultdict(list)
-        for extractor_index, frame_index in zip(extractor_indices, frame_idxs):
-            extractors_dict[extractor_index].append(frame_index)
+        if isinstance(frame_idxs, (int, np.integer)):
+            frame_idxs = [frame_idxs]
+        frame_idxs = np.array(frame_idxs)
+        if all([isinstance(frame_idx, Iterable) for frame_idx in frame_idxs]):
+            frame_idxs = frame_idxs.flatten()
```
CodyCBakerPhD, line 61: I'd still say supporting it in this way is unnecessary. If you remove it you can go back to just using the decorator

CodyCBakerPhD:
> @CodyCBakerPhD, Let me know if I can merge this back to the original PR https://github.com/catalystneuro/roiextractors/pull/127

Might as well just port the tests and the `defaultdict` over directly so we don't inflate the changelog

PR #146, opened 2022-06-16
[Testing] testing if video shape is standard

CodyCBakerPhD:
Just tests

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -2,6 +2,7 @@ import unittest
 from pathlib import Path
 from copy import copy
 
+import numpy as np
 from roiextractors.testing import check_imaging_equal, check_segmentations_equal
 from parameterized import parameterized, param
 from hdmf.testing import TestCase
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Attn @h-mayorquin when you get back - this already works as is, though probably because we don't have any example data with multiple color channels. Going to merge this since its just an extra test

tests/test_io.py
```diff
@@ -68,6 +69,18 @@ def test_imaging_extractors(self, extractor_class, extractor_kwargs):
...
             return
 
+    @parameterized.expand(imaging_extractor_list, name_func=custom_name_func)
+    def test_imaging_extractors_canonical_shape(self, extractor_class, extractor_kwargs):
+        extractor = extractor_class(**extractor_kwargs)
+        image_size = extractor.get_image_size()
+        num_channels = extractor.get_num_channels()
+        video = extractor.get_video()
+
+        canonical_video_shape = [extractor.get_num_frames(), image_size[0], image_size[1]]
```
h-mayorquin, line 79: So, if image_size[0] is columns and image_size[0] is rows this will work but it should not. Also, no need to load the whole video to test this.

h-mayorquin: Oh, I realized this is merged already.

PR #148, opened 2022-06-20
Swap suffix assertions to warnings (ImagingExtractors)

CodyCBakerPhD:
Swaps many of the current suffix assertions to soft warnings.

This is for the ImagingExtractors, I'll do a follow up for the segmentations.

### DIFF
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from warnings import warn
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


h-mayorquin (approved):
Do we really need to raise any warnings at all? If the __init__ fails to initialize it will be because the reader fails and therefore throw an appropriate assertion over there.

CodyCBakerPhD:
> Do we really need to raise any warnings at all?

History: https://github.com/catalystneuro/roiextractors/pull/83

Also suggested by Ben: https://github.com/catalystneuro/roiextractors/pull/144#discussion_r901157915

> If the init fails to initialize it will be because the reader fails and therefore throw an appropriate assertion over there.

Not something you can guarantee since it offloads it to an outside package that may or may not respond that way - also can be much more subtle; perhaps it *can* read a file of a different suffix, just not *correctly*. Best we at least notify the user of this possibility and trust the user to decide. Common case of this is a lab might use a different suffix standard for their experiments, but the actual file itself reads fine through existing tools and they are aware of that fact.

PR #149, opened 2022-06-20
Raise NotImplementedError for formats lacking examples

CodyCBakerPhD:
None of our example data for the various proprietary formats include multiple channels, so we're really unable to be entirely sure about how to go about reading them in that case. `Tiff` may not be currently working as intended, `Sbx` might lead to incorrect data retrieval (if the memmap shape for the `.sbx` file is not correct), and `Nwb` raises an incorrect error type.

This PR standardized the error that gets raised in this case and leads the user to the issues on GitHub to request the feature (where we would also request they share the data to add to GIN).

For the non-proprietary numpy-related and HDF5 extractors we can, of course, make tests involving synthetic data on whatever axes we wish to specify, but this has also not really been done so far.

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -42,6 +42,15 @@ IntType = Union[int, np.integer]
 FloatType = float
 
 
+def raise_multi_channel_or_depth_not_implemented(extractor_name: str):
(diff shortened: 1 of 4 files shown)


h-mayorquin (approved):
Looks good to me.

PR #152, opened 2022-06-20
`num_rows`, `num_cols` terminology standarization

h-mayorquin:
PR to use standard terminology. Changes rows and columns to num_rows and num_cols respectively.

### DIFF
--- a/src/roiextractors/extraction_tools.py
+++ b/src/roiextractors/extraction_tools.py
@@ -60,7 +60,7 @@ class VideoStructure:
 
     Attributes:
         num_rows (int): The number of rows of each frame as a matrix.
-        num_cols (int): The number of columns of each frame as a matrix.
+        num_columns (int): The number of columns of each frame as a matrix.
         num_channels (int): The number of chanenls (1 for gray, 3 for colors).
         rows_axis (int): The axis or dimension corresponding to the rows.
         columns_axis (int):  The axis or dimension corresponding to the columns.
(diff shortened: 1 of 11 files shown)


src/roiextractors/extraction_tools.py
```diff
@@ -132,15 +132,15 @@ def _validate_video_structure(self) -> None:
     def build_video_shape(self, n_frames: int) -> Tuple[int, int, int, int]:
         video_shape = [None] * 4
         video_shape[self.frame_axis] = n_frames
-        video_shape[self.rows_axis] = self.rows
-        video_shape[self.columns_axis] = self.columns
-        video_shape[self.num_channels_axis] = self.num_channels
+        video_shape[self.rows_axis] = self.num_rows
+        video_shape[self.cols_axis] = self.num_cols
+        video_shape[self.channels_axis] = self.num_channels
```
CodyCBakerPhD, line 137: Sorry to keep going back and forth on this kind of thing, but what's your opinion on collapsing `columns` to `cols` everywhere here? I wasn't paying much attention to it before, but I do always think that full words are superior to abbreviations. That said, the mathematician in me likes the symmetry in variable name length between `rows` and `cols`, lol.

src/roiextractors/extraction_tools.py
```diff
@@ -59,44 +59,44 @@ class VideoStructure:
     consistency checks to ensure the validity of the sturcture.
 
     Attributes:
-        rows (int): The number of rows of each frame as a matrix.
-        columns (int): The number of columns of each frame as a matrix.
+        num_rows (int): The number of rows of each frame as a matrix.
+        num_columns (int): The number of columns of each frame as a matrix.
```
CodyCBakerPhD, line 63: Hmm... See other comment on `columns` vs `cols` - this case is at least inconsistent with the current state

src/roiextractors/extraction_tools.py
h-mayorquin: My own preference was to have full words as well. From the point of view that we spend more time reading code than writing it I think that it pays to be redundant. I thought about that when I saw you change in the other PR but does not feel like a big deal in this case and I did not want to nitpick.

src/roiextractors/extraction_tools.py
h-mayorquin: This escape me, I tried to change everything to cols.

src/roiextractors/extraction_tools.py
CodyCBakerPhD: I think we might as well go with full-word `columns` then, I agree on that but haven't probably been the best about it even myself.

src/roiextractors/extraction_tools.py
CodyCBakerPhD: It might take a bit of iteration to purge the entire repo of the references, no prob

src/roiextractors/extraction_tools.py
h-mayorquin: Ok, I tried to do this in the last PR. Let's keep an eye open to see if something got through.

CodyCBakerPhD (approved):

Issue #147, opened 2022-06-18
Should the output of `get_frames` be consistent with numpy indexing for list with single elements?


h-mayorquin:
So, in numpy we have the following:

```python
In [1]: import numpy as np

In [2]: x = np.zeros((10, 4, 3))

In [3]: x[0, ...].shape
Out[3]: (4, 3)

In [4]: x[[0], ...].shape
Out[4]: (1, 4, 3)
```

However, our extractors:
Tiff:
```
extractor.get_frames(0).shape
(60, 80)
extractor.get_frames([0]).shape
(60, 80)
```

HDF5:
```
extractor.get_frames(0).shape
(60, 80)
extractor.get_frames([0]).shape
(60, 80)
```

SBX:
```
extractor.get_frames(0)  <- throws an error
extractor.get_frames([0]).shape
(512, 796)
``` 

NumpyMemmamextractor:
```
extractor.get_frames(0) <- throws an error
extractor.get_frames([0]).shape
(10, 10)
```

This is important as the equality test for extractors that we use requires identical return values in the case with a single element:
https://github.com/catalystneuro/roiextractors/blob/a6c08b864e30fe4648d0f9413ff89f2bb0165b91/src/roiextractors/testing.py#L192

We use this downstream in `nwb-conversion-tools` in our round-trip tests and it will cause problems if `NwbImagingExtractor` does not match the behavior of the other imaging extractors:
https://github.com/catalystneuro/roiextractors/pull/136

Whereas the debate on how much we should follow numpy indexing / slicing convention should be ordered by priorities (see https://github.com/catalystneuro/roiextractors/pull/145) I think that in this case we should stick to numpy convention. 

What do you think? 
@CodyCBakerPhD @weiglszonja

bendichter:
I think the most important thing is that this is documented and consistent between all of the different roiextractors. I'd be OK with going with whichever approach we think will be the most convenient in the long run

CodyCBakerPhD:
We should try to mimic SpikeInterface as much as possible, I think.

Over there, `get_traces` works the way numpy does, by returning a unit axis in that position.

h-mayorquin:
OK, so let's align with numpy for this case. I will open PRs for all the things above.

h-mayorquin:
@CodyCBakerPhD 
> We should try to mimic SpikeInterface as much as possible, I think.


So, I was taking a look at the `get_traces` in spikeinterface:
[...]

PR #154, opened 2022-06-21
Fix `get_frames` behavior for single element list.

h-mayorquin:
As described in #147 we decided to mimic the behavior of numpy. This PR introduces corresponding tests for all the imaging extractors for which we test with data (plus memmap and nwb imaging extractor) and modifies the behavior to obtain the desired behavior.

### DIFF
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -101,6 +101,8 @@ class Hdf5ImagingExtractor(ImagingExtractor):
             slice_stop = self.get_num_frames()
 
         frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel]
+        if isinstance(frame_idxs, int):
+            frames = frames.dsetread().squeeze()
         return frames
 
     def get_image_size(self) -> Tuple[int, int]:
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -100,9 +100,9 @@ class Hdf5ImagingExtractor(ImagingExtractor):
             slice_start = 0
             slice_stop = self.get_num_frames()
 
-        frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel]
+        frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel].dsetread()
         if isinstance(frame_idxs, int):
-            frames = frames.dsetread().squeeze()
+            frames = frames.squeeze()
         return frames
 
     def get_image_size(self) -> Tuple[int, int]:
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
@@ -75,13 +75,21 @@ class ScanImageTiffImagingExtractor(ImagingExtractor):
                 "https://github.com/catalystneuro/roiextractors/issues "
             )
 
-    @check_get_frames_args
     def get_frames(self, frame_idxs: ArrayType, channel: int = 0) -> np.ndarray:
+
+        squeeze_data = False
+        if isinstance(frame_idxs, int):
+            squeeze_data = True
+            frame_idxs = [frame_idxs]
+
         if not all(np.diff(frame_idxs) == 1):
             return np.concatenate([self._get_single_frame(idx=idx) for idx in frame_idxs])
         else:
             with ScanImageTiffReader(filename=str(self.file_path)) as io:
-                return io.data(beg=frame_idxs[0], end=frame_idxs[-1] + 1)
+                frames = io.data(beg=frame_idxs[0], end=frame_idxs[-1] + 1)
+                if squeeze_data:
+                    frames = frames.squeeze()
+            return frames
 
     def _get_single_frame(self, idx: int) -> np.ndarray:
         """
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -86,14 +86,26 @@ class TestExtractors(TestCase):
             return
 
     @parameterized.expand(imaging_extractor_list, name_func=custom_name_func)
-    def test_get_frames_shape(self, extractor_class, extractor_kwargs):
+    def test_imaging_extractors_canonical_shape(self, extractor_class, extractor_kwargs):
+        """Test that get_video and get_frame methods for their shapes and types under different indexing scenarios"""
         extractor = extractor_class(**extractor_kwargs)
-        assert_get_frames_return_shape(imaging_extractor=extractor)
+        image_size = extractor.get_image_size()
+        num_channels = extractor.get_num_channels()
+
+        # Test canonical shape for get video
+        video = extractor.get_video()
+        canonical_video_shape = [extractor.get_num_frames(), image_size[0], image_size[1]]
+        if num_channels > 1:
+            canonical_video_shape.append(num_channels)
+        assert video.shape == tuple(canonical_video_shape)
 
         # Test spikeinterface-like behavior
         one_element_video_shape = extractor.get_video(start_frame=0, end_frame=1, channel=0).shape
         expected_shape = (1, image_size[0], image_size[1])
         assert one_element_video_shape == expected_shape
+
+        # Test frames behavior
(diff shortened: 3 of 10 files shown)


h-mayorquin:
@CodyCBakerPhD This is too big. Breaking this into pieces:
1. Add general test for indexing behavior and test `NwbImagingExtractor` #158.
2. Fix and test `NumpyMemmapImagingExtractor`
3. Fix the rest with a general test in `test_io.py`

tests/test_io.py
```diff
@@ -64,6 +68,8 @@ class TestExtractors(TestCase):
     @parameterized.expand(imaging_extractor_list, name_func=custom_name_func)
     def test_imaging_extractors(self, extractor_class, extractor_kwargs):
         extractor = extractor_class(**extractor_kwargs)
+        assert_get_frames_indexing_with_single_channel(imaging_extractor=extractor)
+
```
CodyCBakerPhD, line 72: Could you make this a separate test function that still uses the same `parameterized.expand` over the `imaging_extractor_list`? That way `test_imaging_extractors` stays as the simple read/write roundtrip and nothing more

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -100,7 +100,9 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0):
             slice_start = 0
             slice_stop = self.get_num_frames()
 
-        frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel]
+        frames = self._video.lazy_slice[slice_start:slice_stop, :, :, channel].dsetread()
+        if isinstance(frame_idxs, int):
+            frames = frames.squeeze()
```
CodyCBakerPhD, line 105: Just to confirm, I had thought that the goal here was to keep the inner frame dimension of `(1,num_rows,num_columns)` in the event `frame_idxs=scalar`? Won't the squeeze reduce it to `(num_rows, num_cols)`?

tests/test_io.py
h-mayorquin: I think that I am more concerned with you that our tests will become slower and slower and this moves us toward that direction. On the other hand, it is better to keep the tests isolated and I think this is a stronger concern here. I will change this.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Numpy does not return the axis corresponding to the scalar when indexed by it:
https://github.com/catalystneuro/roiextractors/issues/147

Maybe there was a misunderstanding?

tests/test_io.py
CodyCBakerPhD: Test runtime should be unaffected either way, the only common overhead is `__init__` of each extractor, which I certainly hope is not a heavy operation.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
CodyCBakerPhD: Looks like a misunderstainding indeed, I did not check what numpy naturally does in that case, I only checked what SpikeInterface does (which keeps the axis in-place). I was saying we should mimic SpikeInterface as much as possible, but I didn't know SpikeInterface wasn't aligned with numpy

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -75,13 +75,21 @@ def __init__(
...
+
         if not all(np.diff(frame_idxs) == 1):
             return np.concatenate([self._get_single_frame(idx=idx) for idx in frame_idxs])
         else:
             with ScanImageTiffReader(filename=str(self.file_path)) as io:
-                return io.data(beg=frame_idxs[0], end=frame_idxs[-1] + 1)
+                frames = io.data(beg=frame_idxs[0], end=frame_idxs[-1] + 1)
+                if squeeze_data:
+                    frames = frames.squeeze()
+            return frames
```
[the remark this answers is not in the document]
h-mayorquin: It fails with the case of a list with a single scalar on it.

tests/test_io.py
h-mayorquin: Yes, my thinking is that it should not be, depends on how we are loading the videos. But overall I agree, we keep the test isolated now and then we should reduce **if** they become too slow.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: Interestingly the decorator  `check_get_frames_args` eliminates the axis in both cases.

tests/test_io.py
h-mayorquin: Should be done, I also removed the testing of the video shape as this is test should cover the same. Moreover, I changed the name of the assert function to something less specific as I think if we decide to move on with only supporting slide we can just change the the current test to account for that.

h-mayorquin:
Closed by mistake.

CodyCBakerPhD:
> Closed by mistake.

Whoa 🤔That's weird - was it because it was linked in the header of the other merged PR or something?

weiglszonja:
@CodyCBakerPhD, yeah it had a "close" keyword in the description of that merged PR.  https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

tests/test_io.py
h-mayorquin: @CodyCBakerPhD 
Thinking further about this, I think that the `__-init__` of each of the extractors is going to be the heaviest operation together with `get_video`. As all the information and in some cases all the data is loaded at the level.

src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
h-mayorquin: I am coming back to this. I am now confused by how you were thinking concerning spikeinterface. As `spikeinterface` uses slice semantics, lists with a single element do not make sense over there.

CodyCBakerPhD: @h-mayorquin The return structure when the slice spawns only a single element keeps the axis in place both for frames and channels (no squeezing)

CodyCBakerPhD: I think @weiglszonja would also benefit from that behavior with her new ImagingDataChunkIterator, as it was an unnecessary complexity to have to add force a new-axis when/if iterating over one axis at a time (important for the iterator's returned data to have the same structure as the selection)

h-mayorquin: Isn't the new `ImagingDataChunkIterator` working with `get_video` instead of with `get_frames`? Get video has the desired behavior so this should not be a problem  anymore.

CodyCBakerPhD: @weiglszonja Can you confirm this?

The relevant lines from her PR: https://github.com/catalystneuro/nwb-conversion-tools/pull/575/files#diff-3419e31114e44476c4474658e8b2892e041cdcf50db4c99176d0d17a18e4fb67R110-R117

but have not tested it myself

CodyCBakerPhD: But yes, we are transitioning to `get_video` thankfully, so that is the place to fix it now (if still as issue)

h-mayorquin: Yes, if you want that the iterator behaves as similar to the one that you use for spikeinterface then you should use `get_video`. 

The tests for the desired behavior should be working here. Please take a look.
https://github.com/catalystneuro/roiextractors/pull/181

CodyCBakerPhD:
What's the CHANGELOG for this PR?

h-mayorquin:
> What's the summary of changes now for this PR? Removing the decorator?
> 
> (I know we don't have a CHANGELOG.md here yet - can you make one?)

@CodyCBakerPhD I just did, from there

 Added unittests to the `get_frames` method from `ImagingExtractors` to assert that they are consistent with numpy
indexing behavior.

CodyCBakerPhD (approved):

PR #157, opened 2022-06-21
Add MultiImagingExtractor to extractorlist

weiglszonja:


h-mayorquin:
> It looks like spikeextractors doesn't include this kind of stuff in the extractor listing. https://github.com/SpikeInterface/spikeextractors/blob/master/spikeextractors/extractorlist.py
> 
> I guess I don't care either way, do you have an opinion @h-mayorquin?

Maybe it could fail with some future test that will iterate over the whole list?
Maybe you add the import the __init__ so is imported directly from the root space but not on the list?

I don't feel very strongly either way.

weiglszonja:
It is also perfectly fine if we don't add it, I didn't know how this extractor listing is maintained so if it can cause an error in the future we can ignore this.

PR #161, opened 2022-06-23
MultiImagingExtractor remove decorator from `get_frames`

weiglszonja:
The `check_get_frames_args` decorator is problematic when frames are returned from the data generator:
https://github.com/catalystneuro/nwb-conversion-tools/blob/07e958d7564eb12d00a5ff5fabe661677d023efa/src/nwb_conversion_tools/tools/roiextractors/roiextractors.py#L188
because when a single frame is retrieved the decorator returns only the first dimension of the frame:
https://github.com/catalystneuro/roiextractors/blob/8db5f9cb3a7ee5efee49b7fd0b694c7a8105519a/src/roiextractors/extraction_tools.py#L298-L299

This solution removes this decorator from being used in the `MultiImagingExtractor`.

### DIFF
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -68,8 +68,9 @@ class MultiImagingExtractor(ImagingExtractor):
 
         return times
 
-    @check_get_frames_args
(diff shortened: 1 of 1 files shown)


weiglszonja:
This will be closed after #186 is merged

PR #162, opened 2022-06-23
Add `folder_path` as an argument for `Suite2pSegmentationExtractor`

h-mayorquin:
Right now in master the `Suite2pSegmentationExtractor` takes `file_path` as one of its arguments but a `folder_path` is passed instead. This PR adds the possibility for using `file_path` while keeping the possibility of using `file_path` for backwards compatibility. A deprecation warning is now issued with no specific date (any ideas?). The tests are also modfied to take this into account and now there are two tests for `Suit2P` data (supporting the old and the new arguments).

Other minor things:
Removed some os.path utilities in favour of pathlib.
The `write_segmentation` method is moved to the bottom of the code to comply with the standard of code organization that we have in other extractors.

### DIFF
--- a/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
+++ b/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
@@ -41,12 +41,12 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
         from warnings import warn
 
         if file_path is not None:
-            warning_stiring = (
-                "The keyword argument 'file_path' is being deprecated in favour of 'folder_path', "
-                "'folder_path' takes precence over 'file_path'"
+            warning_string = (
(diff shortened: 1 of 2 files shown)


src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -72,6 +96,46 @@ def _summary_image_read(self, bstr="meanImg"):
     def roi_locations(self):
         return np.array([j["med"] for j in self.stat]).T.astype(int)
 
+    # defining the abstract class enforced methods:
```
h-mayorquin, line 99: This is just moving.

h-mayorquin:
> Thanks, that was a minor annoyance. Do the conversion tools need to be updated as well?

It should work untill we deprecate but I will propagate now anyway.

PR #165, opened 2022-06-29
Fix `get_dtype` for `NumpyImagingExtractor`

weiglszonja:
`NumpyImagingExtractor.get_dtype()` fails when it tries to retrieve the data type of a single frame using an integer as frame index: 
`self.get_frames(0, 0).dtype`, it should be changed to an ArrayType. 
This PR might be connected to an existing issue #147 
I haven't seen a fix for this (do let me know if there was), neither a unittest for get_dtype.

### DIFF
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -62,7 +62,7 @@ class ImagingExtractor(ABC, BaseExtractor):
         pass
 
     def get_dtype(self) -> DtypeType:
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Works for me. We'll be bringing up some items like this for discussion at the next team meeting.

weiglszonja:
Great, thanks @CodyCBakerPhD

PR #167, opened 2022-06-30
Synch setup installation methods with nwb conversion tools

h-mayorquin:
Small PR so we have consistent naming conventions for installing with extra requirements. Changed `testing` to `test`.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -34,7 +34,7 @@ jobs:
       - name: Test full installation
         run: pip install .[full]
       - name: Install testing requirements (-e needed for codecov report)
-        run: pip install -e .[testing]
+        run: pip install -e .[test]
 
       - name: Get ophys_testing_data current head hash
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
@h-mayorquin One question I had about all this `extra_requires` stuff (which is really nice, BTW) - can you do mix-and-match? Like, if I we wanted to break it into 'full', 'test', 'ecephys' (only installing part of 'full' pertaining to ecephys interfaces), 'ophys', etc... could I do 'ecephys'+'test'? Or do all of this have to be single commands, and to add intersections/unions you'd have to define that as a separate extra altogether?

CodyCBakerPhD (approved):

h-mayorquin:
You can mix [full, test] but I don't think that you can use the names to do set operations. Every handler refers to a list that will have to be defined somewhere (in this case we do the logic from text-file to dicti in the setup).

PR #168, opened 2022-07-02
Scanbox refactoring and docstring

h-mayorquin:
This PR contains three small corrections for sbx:
1) Added a docstring.
2) Refactored assertion errors with file paths and sampling frequency.
3) Changed some links of the documentation that were death and added new ones.

Btw, I think this functions were copied as it is from the changed links. Maybe something to be aware off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,11 @@ Developed by [CatalystNeuro](http://catalystneuro.com/).
 ## Usage:
 ### Supported file types:
 #### Imaging
-1. TIFF  (Generic and ScanImage)
-2. SBX (Scanbox)
-3. HDF5 (Generic)
+1. HDF5
+2. TIFF
+3. STK
+4. FLI
+5. SBX
 
 #### Segmentation
 1. [calciumImagingAnalysis](https://github.com/bahanonu/calciumImagingAnalysis) (CNMF-E, EXTRACT)
--- a/README.md
+++ b/README.md
@@ -17,11 +17,9 @@ Developed by [CatalystNeuro](http://catalystneuro.com/).
 ## Usage:
 ### Supported file types:
 #### Imaging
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -17,11 +17,9 @@ Developed by [CatalystNeuro](http://catalystneuro.com/).
 ## Usage:
 ### Supported file types:
 #### Imaging
-1. HDF5
-2. TIFF
-3. STK
-4. FLI
```
h-mayorquin, line 23: I am not sure which formats are FLI and STK. Maybe removing this is wrong but I could not find them in the library.

CodyCBakerPhD:
> Btw, I think this functions were copied as it is from the changed links. Maybe something to be aware off.

I'm guessing that's because their code is not published on PyPI - if we wanted to import it, we would have to do a `git` install.

README.md
CodyCBakerPhD: @h-mayorquin These can be found implemented here: https://github.com/catalystneuro/roiextractors-gpl/blob/master/roiextractors_gpl/extractors/bioformatsimagingextractors/bioformatsimagingextractors.py

You are free to move them over to this project, but the dependencies *must* be imported lazily due to the licensing difference (never put GPL license in our requirements/setup). It's a secondary software that the user must have installed separately.

We'll see if we can dig up the sample data they were based on but given it's been 2+ years that may be difficult.

h-mayorquin:
> > Btw, I think this functions were copied as it is from the changed links. Maybe something to be aware off.
> 
> I'm guessing that's because their code is not published on PyPI - if we wanted to import it, we would have to do a `git` install.

Makes sense, let me know if there is anything missing for merging.

CodyCBakerPhD (approved):

PR #170, opened 2022-07-04
Add EXTRACT-Public support

CodyCBakerPhD:
Replaces #82

Assigning @weiglszonja to test, debug, and finish.

I just generated the proper file structure through the EXTRACT tutorials and will be posting the test data on the GIN repo soon.


CodyCBakerPhD:
GIN added but might need more work (asking neo team about git-annexing): https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/7

bendichter:
see https://github.com/catalystneuro/ndx-extract

tests/test_extractsegmentationextractor.py
```diff
@@ -155,3 +161,20 @@ def test_extractor_accepted_list(self, accepted_list):
...
+            test_config_struct.create_dataset("max_iter", data=np.array([10.0]))
+            thresholds.create_dataset("S_dup_corr_thresh", data=np.array([0.95]))
+            thresholds.create_dataset("T_dup_corr_thresh", data=np.array([0.95]))
+
+            config = self.extractor._config_struct_to_dict(config_struct=test_config_struct)
+
+        self.assertEqual(config["S_corr_thresh"], [0.1])
+        self.assertEqual(config["max_iter"], [10.0])
+        self.assertEqual(config["thresholds"]["S_dup_corr_thresh"], [0.95])
+        self.assertEqual(config["thresholds"]["T_dup_corr_thresh"], [0.95])
```
weiglszonja, line 181: @CodyCBakerPhD is this a good approach here, or should I return values that exactly match [ndx-extract](https://github.com/catalystneuro/ndx-extract) expected input?

src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
```diff
@@ -169,11 +167,6 @@ def _trace_extractor_read(self) -> DatasetView:
         """Returns the traces with a shape of number of ROIs and number of frames."""
         return DatasetView(self._output_struct["temporal_weights"])
 
-    def _summary_image_read(self) -> np.ndarray:
-        """Returns the summary image as a numpy array where the first dimension is the
-        height of the frame and the second dimension is the width of the frame."""
-        return self._info_struct["summary_image"][:].transpose()
-
```
weiglszonja: This assumption was based off the previous EXTRACT segmentation extractor, but I'm not convinced that they are the same. There is not much information about parameter descriptions in the info struct, only that they contain algorithm statistics. Until it is confirmed I'd remove extracting it and reach out in the meantime to the developers whether what they call "summary image" is indeed the correlation image.

weiglszonja: [EXTRACT_output_to_nwb.m](https://github.com/schnitzer-lab/EXTRACT-public/blob/master/EXTRACT/helper_functions/EXTRACT_output_to_nwb.m#L208) 
Based on their export to NWB, they write: `summary_image`, `F_per_pixel`, and `max_image` to an Images container:
https://github.com/schnitzer-lab/EXTRACT-public/blob/d7a83736918fbef24643860ec6ac3aabfb71c617/EXTRACT/helper_functions/EXTRACT_output_to_nwb.m#L208-L210
To add these images I think we could add them as private attributes to the `NewSegmentationExtractor` class, and override the abstract class `SegmentationExtractor` `get_images_dict()` method, such that we could write these images to NWB. 

https://github.com/catalystneuro/roiextractors/blob/1f8a479e90a45595de661ad339fa53fc022294e2/src/roiextractors/segmentationextractor.py#L203-L212

What do you think about this @CodyCBakerPhD or @bendichter ?

src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py
```diff
@@ -7,22 +14,216 @@
...
+    """Abstract class that defines which extractor class to use for a given file."""
+
+    extractor_name = "ExtractSegmentation"
+    installed = HAVE_H5PY  # check at class level if installed or not
+    installation_mesg = "To use ExtractSegmentationExtractor install h5py: \n\n pip install h5py \n\n"  # error message when not installed
+
+    def __new__(
+        cls,
+        file_path: PathType,
+        sampling_frequency: float,
```
weiglszonja, line 33: @bendichter Is making this a required attribute a good choice when we have irregular timing?
Another example based on [EXTRACT_output_to_nwb.m]:
(https://github.com/schnitzer-lab/EXTRACT-public/blob/master/EXTRACT/helper_functions/EXTRACT_output_to_nwb.m)
https://github.com/schnitzer-lab/EXTRACT-public/blob/d7a83736918fbef24643860ec6ac3aabfb71c617/EXTRACT/helper_functions/EXTRACT_output_to_nwb.m#L46-L52
maybe we could also have `sampling_frequency` and `timestamps` and then add a check within the class that either of them must be supplied. Also correct me if I'm wrong about this, but I don't think currently using timestamps with a `SegmentationExtractor` is supported. Do you think it will make sense to have something similar as it is done in `add_two_photon_series`? 
https://github.com/catalystneuro/neuroconv/blob/69fff3b45828dd27df65cdeea92ab062140dc283/src/neuroconv/tools/roiextractors/roiextractors.py#L418-L425

bendichter: > maybe we could also have sampling_frequency and timestamps and then add a check within the class that either of them must be supplied

that's a good idea

bendichter: > Do you think it will make sense to have something similar as it is done in add_two_photon_series?

Yes, that might be necessary

src/roiextractors/extractorlist.py
```diff
@@ -7,7 +8,9 @@
 from .extractors.nwbextractors import NwbImagingExtractor, NwbSegmentationExtractor
 from .extractors.schnitzerextractor import (
     CnmfeSegmentationExtractor,
+    LegacyExtractSegmentationExtractor,
     ExtractSegmentationExtractor,
+    NewExtractSegmentationExtractor,
```
CodyCBakerPhD, line 13: Just leaving this as a note - I believe  the intention here is to allow people the option of manually overriding the choice of which sub type to use for their data in the event that the main `ExtractSegmentationExtractor` routing function does not behave as they expect?

weiglszonja: Yes, it was my intention, on the other hand maybe but if this is confusing to have 3 types of classes of the same type?

CodyCBakerPhD: At first it was confusing, yes, but I can see the need for it. Alternative is to force them to directly import it from the deeper level in that event.

IDK what do you think? I could see a user looking at it and wondering which class they are meant to use for their data

weiglszonja: having confusing imports is not good probably, the `ExctractSegmentationExtractor` should handle cases when someone wanted to use the older version

weiglszonja (approved):
🎆 🎆 🎆

PR #181, opened 2022-07-24
Test spikeinterface-like behavior for get_video with one element

h-mayorquin:
This is something important for the chunk iterator that is being developed by @weiglszonja. Note that this should be merged after #180 as it has it as a base.

### DIFF
--- a/tests/test_internals/test_nwb_extractors.py
+++ b/tests/test_internals/test_nwb_extractors.py
@@ -84,6 +84,7 @@ class TestNwbImagingExtractor(unittest.TestCase):
         assert num_frames == expected_num_frames
         assert num_channels == expected_num_channels
 
+        # Test numpy like behavior for frame_idxs
         frame_idxs = 0
         frames_with_scalar = nwb_imaging_extractor.get_frames(frame_idxs)
         expected_frames = self.video[frame_idxs, ...]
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
Great, anything else to add?

h-mayorquin:
> Great, anything else to add?

@CodyCBakerPhD @weiglszonja 
I don't know if you care about the case when start_frame == end frame, in that case spikeinterface returns an array of shape `(0, num_channels)`. Is that a edge case of the iterator you are building? What do you think?

CodyCBakerPhD:
> I don't know if you care about the case when start_frame == end frame, in that case spikeinterface returns an array of shape (0, num_channels). Is that a edge case of the iterator you are building? What do you think?

The slicing semantics of the GenericDataChunkIterator should not have an issue with endpoints like that but good to think of...

So in SpikeInterface, if you want to get all channel data for a single frame, you need to slice from `start_frame` to `start_frame+1`? And to access the final frame that way you would need to do from `final_frame` to `final_frame+1`?

h-mayorquin:
> So in SpikeInterface, if you want to get all channel data for a single frame, you need to slice from `start_frame` to `start_frame+1`? And to access the final frame that way you would need to do from `final_frame` to `final_frame+1`?

`start_frame=0, end_frame=1` -> you get the first frame
`start_frame=last_frame - 1,  end_frame=last_frame` -> you get the last frame

PR #187, opened 2022-08-05
`SegmentationExtractor`: add frame_to_time, get_roi_ids as class methods

weiglszonja:
- Adds frame_to_time method in SegmentationExtractor: can be used to define timestamps when writing segmentation data to NWB
- get_roi_ids can be used from SegmentationExtractor, no need to define in child classes if they are using the same logic. 
- Other smaller changes: refactor docstrings, get_sampling_frequency will not break if private attribute _sampling_frequency is None

### DIFF
--- a/src/roiextractors/segmentationextractor.py
+++ b/src/roiextractors/segmentationextractor.py
@@ -233,6 +233,9 @@ class SegmentationExtractor(ABC, BaseExtractor):
         sampling_frequency: float
             Sampling frequency of the recording in Hz.
         """
+        if self._sampling_frequency is not None:
+            return float(self._sampling_frequency)
+
         return self._sampling_frequency
 
     def get_num_rois(self):
--- a/src/roiextractors/segmentationextractor.py
+++ b/src/roiextractors/segmentationextractor.py
@@ -1,9 +1,10 @@
 from abc import ABC, abstractmethod
+from typing import Union
 
 import numpy as np
 from spikeextractors.baseextractor import BaseExtractor
 
-from .extraction_tools import ArrayType
+from .extraction_tools import ArrayType, IntType, FloatType
 from .extraction_tools import _pixel_mask_extractor
 
 
@@ -103,7 +104,7 @@ class SegmentationExtractor(ABC, BaseExtractor):
         Returns
         -------
         roi_ids: list
-            ROI ids list.
+            List of roi ids.
         """
         return list(range(self.get_num_rois()))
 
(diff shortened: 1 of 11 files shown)


CodyCBakerPhD:
Assuming there are tests for all of the `get_roi_ids()` for each extractor, can you add a test for the new `frame_to_time` function?

src/roiextractors/segmentationextractor.py
```diff
@@ -231,18 +230,21 @@ def get_sampling_frequency(self):
...
-        samp_freq: float
-            Sampling frequency of the recordings in Hz.
+        sampling_frequency: float
+            Sampling frequency of the recording in Hz.
         """
-        return float(self._sampling_frequency)
+        if self._sampling_frequency is not None:
+            return float(self._sampling_frequency)
+
+        return self._sampling_frequency
```
weiglszonja, line 239: Should I rollback this change @CodyCBakerPhD ?

src/roiextractors/segmentationextractor.py
```diff
@@ -270,13 +272,31 @@ def get_num_channels(self):
...
         Returns the default number of planes of imaging for the segmentation extractor.
-        Detaults to 1 for all but the MultiSegmentationExtractor
+        Defaults to 1 for all but the MultiSegmentationExtractor
         Returns
         -------
         self._num_planes: int
         """
         return self._num_planes
 
+    def frame_to_time(self, frame_indices: Union[IntType, ArrayType]) -> Union[FloatType, ArrayType]:
```
h-mayorquin, line 282: ArrayType includes int I think.

What do you think of having `frame_indices=None` means getting all the times?

h-mayorquin (commented):
Something that I am wondering is if the `roi_ids` were meant to be int values to beging with or they could be string. In the ephys pipeline we have the indexes of the table and the unit_name for example.

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: If I recall, the approach we took with ImagingExtractors was to define it at each format level: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L36 and assert in every format specific test that it is, indeed, a float https://github.com/catalystneuro/roiextractors/blob/52f4704c0148ce2a95fd6294ed68677b235fc056/src/roiextractors/testing.py#L353

So yeah I'd say let's synchronize this with the segmentation. It can be in a separate PR though, this one can just establish the frame_to_time

CodyCBakerPhD (approved):

CodyCBakerPhD:
> Something that I am wondering is if the roi_ids were meant to be int values to beging with or they could be string.

Do we have any other infrastructure that indicates string references to ROI IDs?

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: > ArrayType includes int I think.

If you examine the full nesting of `ArrayLike` it does not contain our exact definition of `IntType`, though comes close to the spirit.

However, annotation typing is really just for us to be able to look at this signature and understand what it accepts. As I read it, I can specify numpy slicing semantics (single integer or array or integers) to get those frames as times.

Adding `None` to that would also reinforce numpy-like semantics, yes.

As for whether or not we want to support numpy semantics in general (as per previous conversations about the imaging extractor `get_frames`, that's the more central question I think.

h-mayorquin: It is not about numpy-like semantics which I think would not apply here.

It is more so it is consistent with other methods of `segmentation_extractor` where passing an argument as None means give me at all of it. See `get_roi_pixel_masks`.

h-mayorquin: Concerning the typing issue that's fine if you think is more clear to read. However, while I have not checked the nesting the documentation seems to be clear that it should contain scalars such as a int:

https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike

h-mayorquin:
> > Something that I am wondering is if the roi_ids were meant to be int values to beging with or they could be string.
> 
> Do we have any other infrastructure that indicates string references to ROI IDs?

@CodyCBakerPhD 

Not very clear. The functions `get_roi_locations` and `get_roi_image_masks` seem to indicate in their code that the roi_ids are not the indexes but maybe the intent is that they were integers but not ordered. Check them out.

PR #188, opened 2022-08-08
`SegmentationExtractor`: add `set_times`

weiglszonja:
Follow up on #187

Adds `set_times` method to SegmentationExtractor.
The use case is to be able to write segmentation data with timestamps.


CodyCBakerPhD:
@weiglszonja Ready for review?

weiglszonja:
@CodyCBakerPhD  yes, let me know if there are other tests that should be added

CodyCBakerPhD (approved):

PR #190, opened 2022-08-12
[refactor] Testing SegmentationExtractor images

weiglszonja:
Testing segmentation images should be more flexible than relying on the assumption that segmentation extractors all have "correlation" and "mean" images.

### DIFF
--- a/src/roiextractors/testing.py
+++ b/src/roiextractors/testing.py
@@ -252,12 +252,13 @@ def check_segmentation_return_types(seg: SegmentationExtractor):
         element_dtypes=floattype,
         shape_max=(np.prod(seg.get_image_size()), 3),
     )
-    _assert_iterable_complete(
-        seg.get_image(),
-        dtypes=(np.ndarray, NoneType),
-        element_dtypes=floattype,
-        shape_max=(*seg.get_image_size(),),
-    )
+    for image_name in seg.get_images_dict():
+        _assert_iterable_complete(
+            seg.get_image(image_name),
+            dtypes=(np.ndarray, NoneType),
+            element_dtypes=floattype,
+            shape_max=(*seg.get_image_size(),),
+        )
     _assert_iterable_complete(
         seg.get_accepted_list(),
         dtypes=(list, NoneType),
@@ -285,7 +286,6 @@ def check_segmentation_return_types(seg: SegmentationExtractor):
     assert isinstance(seg.get_traces_dict(), dict)
     assert isinstance(seg.get_images_dict(), dict)
     assert {"raw", "dff", "neuropil", "deconvolved"} == set(seg.get_traces_dict().keys())
-    assert {"mean", "correlation"} == set(seg.get_images_dict().keys())
 
 
 def check_imaging_equal(
--- a/src/roiextractors/testing.py
+++ b/src/roiextractors/testing.py
@@ -231,7 +231,7 @@ def check_segmentations_images(
     assert len(images_in_extractor1) == len(images_in_extractor2)
 
     image_names_are_equal = all(image_name in images_in_extractor1.keys() for image_name in images_in_extractor2.keys())
-    assert image_names_are_equal, "The name of segmentation images in the segmentation extractors are not equal."
+    assert image_names_are_equal, "The names of segmentation images in the segmentation extractors are not the same."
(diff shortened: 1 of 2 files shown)


src/roiextractors/testing.py
```diff
@@ -252,12 +256,13 @@ def check_segmentation_return_types(seg: SegmentationExtractor):
         element_dtypes=floattype,
         shape_max=(np.prod(seg.get_image_size()), 3),
     )
-    _assert_iterable_complete(
-        seg.get_image(),
-        dtypes=(np.ndarray, NoneType),
```
h-mayorquin, line 257: Isn't the NoneType here taking care of if seg.get_image() is empty? or what is the function of the loop here?

src/roiextractors/testing.py
```diff
@@ -203,7 +203,11 @@ def check_segmentations_equal(
     ) == set(
         segmentation_extractor2.get_roi_pixel_masks(roi_ids=segmentation_extractor1.get_roi_ids()[:1])[0].flatten()
     )
-    assert_array_equal(segmentation_extractor1.get_image(), segmentation_extractor2.get_image())
```
h-mayorquin, line 206: Maybe you can test for the length of the dictionary being equal, if that does not work then the other test is not necesssary. Also, maybe you can check for the names of the images being equals so the `get_images_dict()`  are compared both in keys and items?

src/roiextractors/testing.py
```diff
@@ -203,7 +203,11 @@ def check_segmentations_equal(
...
+    for image_name in segmentation_extractor1.get_images_dict().keys():
+        assert_array_equal(
```
h-mayorquin, line 207: Wouldn't `assert_array_almost_equal` be more appropriate?

src/roiextractors/testing.py
weiglszonja: for `seg.get_image()` the name by default is "correlation", if "correlation" is not in `get_images_dict()` then this errors out before that.

weiglszonja: In `ExtractSegmentationExtractor` instead of correlation or mean image we have other images: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py#L243

The intention here is if there is no correlation image, we should check the images that are actually in the segmentation

src/roiextractors/testing.py
weiglszonja: Why is that?

src/roiextractors/testing.py
h-mayorquin: I see, that make sense. Thanks.

src/roiextractors/testing.py
h-mayorquin: If the dtype of the images end up being float their specific floating point representation might different even if they are same value. Not sure how robust is `assert_array_equal` by itself.

h-mayorquin: Try something like this
```python
In [13]: x = np.random.rand(10, 10)

In [14]: np.testing.assert_array_equal(x, x.round(22))
```

CodyCBakerPhD: If strict equality passes without issue, why would we lessen it? There should in principle be no data loss even at some arbitrary level of rounding, and it would even be useful to know if such a event can occur

src/roiextractors/testing.py
weiglszonja: I moved this to a separate check method, thanks for the suggestions.

CodyCBakerPhD (dismissed):

CodyCBakerPhD (approved):

weiglszonja:
@CodyCBakerPhD changelog updated

CodyCBakerPhD (approved):

src/roiextractors/testing.py
h-mayorquin: The role of this function is to test if two segmentations are equal. The fact that it passes the tests doesn't mean that it will work "in the field" . This is like comparing floats with equality because it works on the test set.

h-mayorquin: In other words you avoid false positives for the assertion.

PR #191, opened 2022-08-12
[refactor] `NwbSegmentationExtractor`

weiglszonja:
Updating `NwbSegmentationExtractor`  is necessary after merging [catalystneuro/neuoroconv#41](https://github.com/catalystneuro/neuroconv/pull/41) to keep up with the change of trace names.
The `get_images_dict()` is also changed to be more flexible if there are other segmentation images than correlation or mean images. 

This PR prepares the update of `ExtractSegmentationInterface` in catalystneuro/neuroconv#87.

### DIFF
--- a/src/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/src/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -328,9 +328,11 @@ class NwbSegmentationExtractor(SegmentationExtractor):
                 return rej_list
 
     def get_images_dict(self):
-        if self._segmentation_images is None:
-            return super().get_images_dict()
-        images_dict = {image_name: image_data[:].T for image_name, image_data in self._segmentation_images.items()}
+        images_dict = super().get_images_dict()
+        if self._segmentation_images is not None:
+            images_dict.update(
+                (image_name, image_data[:].T) for image_name, image_data in self._segmentation_images.items()
+            )
 
         return images_dict
 
--- a/src/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/src/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -248,7 +248,7 @@ class NwbSegmentationExtractor(SegmentationExtractor):
             .nwb file location
         """
         check_nwb_install()
-        SegmentationExtractor.__init__(self)
+        super().__init__()
         file_path = Path(file_path)
         if not file_path.is_file():
             raise Exception("file does not exist")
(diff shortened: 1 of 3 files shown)


src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -260,61 +260,53 @@ def __init__(self, file_path: PathType):
...
+        # Extract roi_responses:
+        fluorescence = None
+        df_over_f = None
+        any_roi_response_series_found = False
+        if "Fluorescence" in ophys.data_interfaces:
+            fluorescence = ophys.data_interfaces["Fluorescence"]
+        if "DfOverF" in ophys.data_interfaces:
+            df_over_f = ophys.data_interfaces["DfOverF"]
+        if fluorescence is None and df_over_f is None:
+            raise Exception("Could not find Fluorescence/DfOverF module in nwbfile.")
```
h-mayorquin, line 275: Don't the `data_interfaces["Key"]` above return a KeyError when the "Fluoresence and DfOverF do not exist? Do them just return None?

h-mayorquin (commented):
Looks OK, did you try to see if all the tests in `neuroconv` work with this?

src/roiextractors/extractors/nwbextractors/nwbextractors.py
weiglszonja: no, actually i didn't change that part, the intention was to declare above `df_over_f`, `fluorescence`=`None`, `None` and only override with the actual interface if we have that key. does that answer your question?

weiglszonja:
I'll need to test with other segmentation datasets, so far I only checked it with the extract segmentation data.

src/roiextractors/extractors/nwbextractors/nwbextractors.py
h-mayorquin: Right, if you did not change then is not in the scope of this PR. Let's come back to this later. 

It is good that you eliminated the necessary outer loop for this PR (making the code flatter) is just that the diff then shows everything and it was hard to see what you did change. That line got my eye.

h-mayorquin:
> I'll need to test with other segmentation datasets, so far I only checked it with the extract segmentation data.

One way to do it would be to open a PR in neuroconv pinning to this commit.

weiglszonja:
Testing commit 61e61d5a40c382183e4e6e659a9e8001b36107bb on catalystneuro/neuroconv#87

src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -328,9 +328,11 @@ def get_rejected_list(self):
...
 
     def get_images_dict(self):
-        if self._segmentation_images is None:
-            return super().get_images_dict()
-        images_dict = {image_name: image_data[:].T for image_name, image_data in self._segmentation_images.items()}
+        images_dict = super().get_images_dict()
+        if self._segmentation_images is not None:
+            images_dict.update(
+                (image_name, image_data[:].T) for image_name, image_data in self._segmentation_images.items()
+            )
```
weiglszonja, line 335: @h-mayorquin 
This seems like finally fixed the gin tests in catalystneuro/neuroconv/#87. So why is that? (thinking out loud)
Our segmentation extractor writes segmentation data with only `correlation` or images with some other name to an nwbfile. Then, if we wanted to test whether data was written correctly we are reading this nwbfile in with `NwbSegmentationExtractor`. Why override get_images_dict() in the first place? This was necessary to allow saving images that are not named "correlation" or "mean" images -- as it is defined in `SegmentationExtractor.get_images_dict()`. Without the super() class call if a segmentation extractor only wrote "correlation" image then the comparison would fail because its get_images_dict() would return also "mean" image that is None --> we should also have this `NwbSegmentationExtractor`. 

The hotfix for `ExtractSegmentationExtractor` will be done in a separate PR, I just wanted to see if I add the super() call there then the ExtractSegmentationInterface would pass also this test.

CodyCBakerPhD: I'd agree - neither ROIExtractors base nor the NWB extractor should expect 'mean' and 'correlation' to be the only (or any) specified summary images - different segmentation methods seem to have a wide variety of statistical methods for displaying/summarizing the input.

weiglszonja: Yes, I'm just pointing out that the way I managed to solve this issue is if there are other images other than "correlation" or "mean" then the super() class call will be necessary when overriding the parent class SegmentationExtractor's `get_images_dict()` method. This was the best solution I come up today to make all previous tests pass with this change (of having other images in segmentation).

tests/test_extractsegmentationextractor.py
```diff
@@ -199,7 +199,10 @@ def test_extractor_get_images_dict(self):
             )[:]
 
         images_dict = self.extractor.get_images_dict()
-        self.assertEqual(len(images_dict), 3)
+        self.assertEqual(len(images_dict), 5)
+
+        self.assertEqual(images_dict["correlation"], None)
+        self.assertEqual(images_dict["mean"], None)
```
CodyCBakerPhD, line 205: Long-term TODO: I wish we didn't require these for all segmentation extractors, since everyone has a different way of summarizing images that don't clearly fit into notions of mean/correlation

CodyCBakerPhD (approved):

PR #195, opened 2022-08-15
Improvements to MultiImagingExtractor

CodyCBakerPhD:
Collection of fixes needed for volumetric data as well as integration with the `ImagingDataChunkIterator` in https://github.com/catalystneuro/neuroconv/pull/90

Probably easiest way to point out each fix is to comment the code.


src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -47,10 +47,13 @@ def __init__(
         self._channel_names = channel_names
 
         self._file = h5py.File(file_path, "r")
-        if "mov" in self._file.keys():
+        if mov_field in self._file.keys():
```
CodyCBakerPhD, line 50: The standard `Hdf5ImagingExtractor` wasn't actually useful for my case, but I saw this and fixed it anyway. Can splinter into a different PR if needed.

src/roiextractors/imagingextractor.py
```diff
@@ -67,8 +67,10 @@ def get_video(
         pass
 
     def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> np.ndarray:
-
-        return self.get_video()[frame_idxs, ..., channel]  # To-do simplify for contigious frame_idxs case
+        if np.all(np.diff(frame_idxs) == 0):
+            return self.get_video(start_frame=frame_idxs[0], end_frame=frame_idxs[-1])
```
CodyCBakerPhD, line 71: This implements the `# To-do` for contiguous frame indices

src/roiextractors/imagingextractor.py
```diff
@@ -67,8 +67,10 @@ def get_video(
...
+        relative_indices = np.array(frame_idxs) - frame_idxs[0]
+        return self.get_video(start_frame=frame_idxs[0], end_frame=frame_idxs[-1])[relative_indices, ..., channel]
```
CodyCBakerPhD, line 73: However, the original `self.get_video()` call almost crashed my computer by attempting to load an entire video into memory just to do the fancy indexing on it - all in order to determine the `dtype` from the first frame. If the return type from `get_video` is a memmap this obviously isn't a problem, but it may not always be guaranteed to be.

This correction only loads the amount of video needed to perform the fancy indexing.

src/roiextractors/imagingextractor.py
```diff
@@ -210,8 +212,13 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> np.nd
         mapped_frame_idxs = np.array(frame_idxs) + self._start_frame
         return self._parent_imaging.get_frames(frame_idxs=mapped_frame_idxs, channel=channel)
 
-    def get_video(self, start_frame=None, end_frame=None, channel: Optional[int] = 0) -> np.ndarray:
-        assert max(end_frame) < self._num_frames, "'end_frame' range beyond number of available frames!"
```
CodyCBakerPhD, line 214: The `end_frame` of `get_video` ought to follow slicing semantics, so it is perfectly safe to request frames beyond the range of the video (it just returns an empty array in that event)

src/roiextractors/imagingextractor.py
```diff
@@ -210,8 +212,13 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> np.nd
...
+    def get_video(
+        self, start_frame: Optional[int] = None, end_frame: Optional[int] = None, channel: Optional[int] = 0
+    ) -> np.ndarray:
+        assert start_frame >= 0, (
+            f"'start_frame' must be greater than or equal to zero! Received '{start_frame}'.\n"
+            "Negative slicing semantics are not supported."
+        )
```
CodyCBakerPhD, line 221: However, allowing negative slicing is a different beast. Not opposed to it, but not going to implement it here.

src/roiextractors/multiimagingextractor.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, imaging_extractors: List[ImagingExtractor]):
         # Checks that properties are consistent between extractors
         self._check_consistency_between_imaging_extractors()
 
+        self._dtype = self._imaging_extractors[0].get_dtype()
```
CodyCBakerPhD, line 34: Maybe I'll just overload the function itself instead... WIP

src/roiextractors/multiimagingextractor.py
```diff
@@ -68,45 +69,37 @@ def _get_times(self):
...
+        current_frame = 0
+        for extractor_index in extractors_spanned:
+            relative_start = self._start_frames[extractor_index] - start
+            relative_stop = min(self._end_frames[extractor_index], stop) - start
+            relative_span = relative_stop - relative_start
+            total_selection = slice(current_frame, current_frame + relative_span)
+            print(f"relative_start = {relative_start}")
+            print(f"relative_stop = {relative_stop}")
+            print(f"relative_span = {relative_span}")
+            print(f"total_selection = {total_selection}")
```
CodyCBakerPhD, line 97: Will remove these when closer to merge

src/roiextractors/multiimagingextractor.py
weiglszonja: Yes, it is a better idea. If we are using this perhaps we should check also for dtype consistency in `_check_consistency_between_imaging_extractors`?

CodyCBakerPhD: Oh good catch, didn't even notice that

CodyCBakerPhD: Added to consistency check and defined override method

src/roiextractors/multiimagingextractor.py
```diff
@@ -97,16 +106,53 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> Numpy
...
+            array_frame_slice = slice(current_frame, current_frame + relative_span)
+            video[array_frame_slice, ...] = self._imaging_extractors[extractor_index].get_video()
+            current_frame += relative_span
+
+        # Right endpoint; since more than one extractor is spanned, only care about indexing final end frame
+        relative_stop = stop - self._start_frames[extractors_spanned[-1]]
+        array_frame_slice = slice(current_frame, None)
+        video[array_frame_slice, ...] = self._imaging_extractors[extractors_spanned[-1]].get_video(
+            end_frame=relative_stop
+        )
```
CodyCBakerPhD, line 154: There is undoubtedly a mathematical expression for including each of these steps into a single iteration loop, but I feel like this is easier to read and comprehend

CodyCBakerPhD:
@weiglszonja Ready for full review now

CodyCBakerPhD:
@weiglszonja You'll also want to try out this branch with your stuff from the Fee lab in conjunction with https://github.com/catalystneuro/neuroconv/pull/90, let me know if it behaves as expected for you

weiglszonja:
@CodyCBakerPhD works for me, the imaging data looks correct (compared with an older version).
Used this branch for `get_video` https://github.com/catalystneuro/fee-lab-to-nwb/pull/19

src/roiextractors/multiimagingextractor.py
weiglszonja: thanks!

CodyCBakerPhD:
@h-mayorquin Any chance you could take a look at this while Szonja is enjoying vacation?

src/roiextractors/imagingextractor.py
h-mayorquin: Good point.

src/roiextractors/imagingextractor.py
h-mayorquin: I agree, low priority.

src/roiextractors/multiimagingextractor.py
```diff
@@ -97,16 +106,53 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> Numpy
...
+            raise NotImplementedError(
+                f"MultiImagingExtractors for multiple channels have not yet been implemented! (Received '{channel}'."
+            )
+
         start = start_frame if start_frame is not None else 0
         stop = end_frame if end_frame is not None else self.get_num_frames()
-        frame_idxs = range(start, stop)
-        return self.get_frames(frame_idxs=frame_idxs, channel=channel)
+        total_range = list(range(start, stop))
+        extractors_spanned = np.unique(np.searchsorted(self._end_frames, total_range, side="right"))
```
h-mayorquin, line 117: Clever!

h-mayorquin (changes requested):
Loos good to me. 
I was looking at `test_multiimagingextractor` and we don't have tests for this though.

Can we add tests with three extractors that:
1) That only slices from second extractor
2) Slices from second and third.
3) Slices from middle of first to the middle of third.

I would be good to have those calls.

src/roiextractors/multiimagingextractor.py
CodyCBakerPhD: Could be more clever. As is, still generated an entire vector of length `stop - start`, so for a really long imaging extractor could get more expensive than necessary

CodyCBakerPhD: Anyway, fixed that. Only uses the two bounds now

CodyCBakerPhD:
> Can we add tests with three extractors that:

> That only slices from second extractor
> Slices from second and third.
> Slices from middle of first to the middle of third.
> I would be good to have those calls.

Thanks, I added these extra tests which revealed a necessary fix for some endcap edge cases.

CodyCBakerPhD:
@weiglszonja Just waiting for your approval that the recent changes work with your imaging data as well

weiglszonja (approved):
Works with https://github.com/catalystneuro/fee-lab-to-nwb/pull/19 👍

PR #196, opened 2022-08-20
Remove spikeextractors dependency

CodyCBakerPhD:
ROIExtractors does not test or make use of any of the methods inherited from the `BaseExtractor` imported from the legacy `spikeextractors`. This removes it from the minimal installation and base `ImagingExtractor`.


requirements-testing.txt
```diff
@@ -1,3 +1,4 @@
 pytest
 pytest-cov
 parameterized==0.8.1
+spikeextractors>=0.9.10
```
CodyCBakerPhD, line 4: While we haven't been utilizing any of the I/O or class infrastructure from `spikeextractors`, the `toy_example` function makes extreme use of the synthetic data code from `spikeextractors`. Thus making it an optional dependency for that singular purpose. I tried to port over some of the code but it's a LOT to move over and I'm not really sure if it ought to be within the scope of ROIExtractors (as an I/O package) but rather might be better in a package devoted to generated synthetic data (though using existing packages like BRIAN might be worth a look into for that...)

h-mayorquin: I look at it once when I was trying to remove this. Most of the functionality of the `toy_example` is easier to remove. The only thing I could not remove is the generation of spikes in terms of frames which is requires more functionality if we want to keep it backwards compatible. If we just want to generate fake spikes for illustration purposes we could get rid of this entirely.

CodyCBakerPhD: Seemed to do more than just simple spike generation - also included some waveform geometry type stuff? It seemed decently complicated, but if you think it can be done away with and you know how to do it, please do lol

CodyCBakerPhD:
> Are we still using the decorator for video arguments somewhere? The casting code that you added is related to it but I am not sure that decorator is needed anymore.

No, we got permission a while back to do away with those completely.

But the casting function I had to port over was something inherited from spikeextractors for some extraction tool function, I didn't looks too deep into it but some time we should really review that module and see what we can do away with.

requirements-testing.txt
h-mayorquin: Here the follow-up on this:
https://github.com/catalystneuro/roiextractors/issues/204

PR #197, opened 2022-08-20
Pull data from S3 bucket

CodyCBakerPhD:
Same idea as https://github.com/catalystneuro/neuroconv/pull/102


CodyCBakerPhD:
Annoyingly, all of the GitHub actions used for S3 transfers are based off of some container that is only compatible with linux - as such, I adjusted the workflow to do the direct installation and usage of the AWS CLI (which is what all the actions ever seemed to need to do anyway).

Now working well on all platforms.

Also, worth noting that I notice the transfer speeds are several orders of magnitude faster than they were with datalad, so that's a plus as well.

bendichter (approved):

PR #198, opened 2022-08-20
Make external dependencies imported lazy

CodyCBakerPhD:
In the spirit of https://github.com/catalystneuro/neuroconv/pull/74, this is all that would be required here to take advantage of fully lazy imports of external dependencies for all OphysInterfaces on NeuroConv.


src/roiextractors/extraction_tools.py
```diff
@@ -480,3 +487,59 @@ def todict(matobj):
         else:
             dict[strg] = elem
     return dict
+
+
+def get_package(
+    package_name: str,
+    installation_source: str = "pip",
+    package_installation_display: Optional[str] = None,
```
h-mayorquin, line 495: I would for `name` instead of `display` here (or something else).

CodyCBakerPhD: Thought here (though this can apply to `installation_source`) was that it could be a string that also has other information other than just the name. Example

```
installation_source = "conda -c conda-forge"
```

or

```
package_installation_display = "--no-deps special-name-of-Package[special_options_too]"
```

Though IDK, it could also be clearer to add those other things as `installation_options: str = ""` separately and keep the other options as strictly the name mappings?

h-mayorquin: My suggestion would be to use a single argument called `installations_instructions` for line 544 so you can get rid of `installation_source` and `installation_display`as they don't seem to lend themselves that well to that decomposition.

Also, would use `package_name` instead of `package_installation_display` on line 538.

CodyCBakerPhD: Done

bendichter (approved):

PR #201, opened 2022-08-28
[Merge after #200]: Subframe segmentation

CodyCBakerPhD:
Relies on trace orientation fix/standardization in #200.

Introduces subframe selection capabilities to SegmentationExtractors as well as tests for this.

Necessary to introduce frame stubbing into the SegmentationDataInterface on NeuroConv.

### DIFF
--- a/src/roiextractors/segmentationextractor.py
+++ b/src/roiextractors/segmentationextractor.py
@@ -1,7 +1,8 @@
 from abc import ABC, abstractmethod
-from typing import Union
+from typing import Union, Optional, Tuple
 
 import numpy as np
+from numpy.typing import ArrayLike
 
 from .extraction_tools import ArrayType, IntType, FloatType
 from .extraction_tools import _pixel_mask_extractor
@@ -158,6 +159,10 @@ class SegmentationExtractor(ABC):
             2-D array: image y x image x
         """
         pass
+
+    def frame_slice(self, start_frame, end_frame):
+        """Return a new SegmentationExtractor ranging from the start_frame to the end_frame."""
+        return FrameSliceSegmentationExtractor(parent_imaging=self, start_frame=start_frame, end_frame=end_frame)
 
     def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name="raw"):
         """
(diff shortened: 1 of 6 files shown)


src/roiextractors/segmentationextractor.py
```diff
@@ -285,7 +289,7 @@ def set_times(self, times: ArrayType):
         assert len(times) == self.get_num_frames(), "'times' should have the same length of the number of frames!"
         self._times = np.array(times, dtype=np.float64)
 
-    def frame_to_time(self, frame_indices: Union[IntType, ArrayType]) -> Union[FloatType, ArrayType]:
+    def frame_to_time(self, frames: Union[IntType, ArrayType]) -> Union[FloatType, ArrayType]:
```
h-mayorquin, line 292: We are calling `frame_idxs` in the imaging extractor, we should make this consistent across the library I think.

src/roiextractors/segmentationextractor.py
```diff
@@ -319,3 +323,120 @@ def write_segmentation(segmentation_extractor, save_path, overwrite=False):
...
+            end_frame = parent_size
+        else:
+            assert 0 < end_frame <= parent_size
+        assert end_frame > start_frame, "'start_frame' must be smaller than 'end_frame'!"
+
+        super().__init__()
+        if getattr(self._parent_segmentation, "_times") is not None:
+            self._times = self._parent_segmentation._times[start_frame:end_frame]
+
+    def get_accepted_list(self) -> list:
```
h-mayorquin, line 366: I guess this is for testing purposes only but what if the frames that you cut do not contain the rois here. I don't remember what spikeinterface does in this case. Most likely they don't check for this type of condition.

src/roiextractors/segmentationextractor.py
```diff
@@ -319,3 +323,120 @@ def write_segmentation(segmentation_extractor, save_path, overwrite=False):
...
+    def get_traces(
+        self,
+        roi_ids: Optional[Iterable[int]] = None,
+        start_frame: Optional[int] = None,
+        end_frame: Optional[int] = None,
+        name: str = "raw",
+    ) -> np.ndarray:
+        start_frame = min(start_frame or 0, self._num_frames)
+        end_frame = min(end_frame or self._num_frames, self._num_frames)
+        return self._parent_segmentation.get_traces(
```
h-mayorquin, line 381: I am glad that the general method already has `start_frame` and `end_frame` functionality implemented.

src/roiextractors/segmentationextractor.py
```diff
@@ -319,3 +323,120 @@ def write_segmentation(segmentation_extractor, save_path, overwrite=False):
...
+                        "The `get_traces_dict` method for SubFrameSegementations does not yet support underlying trace "
+                        "types other than numpy arrays (which includes memory maps)."
+                    )
+
+        def _safe_subtrace(trace: Optional[np.ndarray], start: int, end: int):
+            if trace is None and len(trace.shape) > 0:
+                return None
+            return trace[start:end, :]
+
+        return dict(
```
h-mayorquin, line 402: What do you think of going over the dict of the parent and using the names instead. That would make this shorter and will be future proof if we add one more type of trace (trace_name) in the future.

src/roiextractors/segmentationextractor.py
```diff
@@ -319,3 +323,120 @@ def write_segmentation(segmentation_extractor, save_path, overwrite=False):
...
+                if not isinstance(trace, np.ndarray):
+                    raise NotImplementedError(
+                        "The `get_traces_dict` method for SubFrameSegementations does not yet support underlying trace "
+                        "types other than numpy arrays (which includes memory maps)."
+                    )
+
+        def _safe_subtrace(trace: Optional[np.ndarray], start: int, end: int):
+            if trace is None and len(trace.shape) > 0:
+                return None
+            return trace[start:end, :]
```
h-mayorquin, line 400: Why not use parent.get_traces() here?

h-mayorquin (changes requested):
Some comments here and there.

Can you add a test where the segmentation extractor does not have one of the traces? there is an argument in `generate_dummy_segmentation_extractor` to control that. The sliced segmentation extractor traces_dict should not have more keys than the father.

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: `frames` is the general argument call passed to `frame_to_time` in the ImagingExtractor: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L69

So I'm being consistent with that.

As for `get_frames` taking `frame_idxs`, yeah that should probably be standardized

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: How do you mean? With regards to motion correction?

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: Doing it better now

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: Done

CodyCBakerPhD:
> Can you add a test where the segmentation extractor does not have one of the traces? there is an argument in generate_dummy_segmentation_extractor to control that. The sliced segmentation extractor traces_dict should not have more keys than the father.

Done

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: Currently there is no time mapping from frames to roi_ids, so that kind of thing is not currently possible or a 'problem'

src/roiextractors/segmentationextractor.py
h-mayorquin: Yep. Makes sense:
https://github.com/catalystneuro/roiextractors/issues/202

src/roiextractors/segmentationextractor.py
h-mayorquin: Right, that makes sense, I was thinking that in spikeinterface there are some units whose first spike is later in time and maybe it should be cut-out if you sub-frame but it seems that here that's not a problem as we have traces instead for each roi and image masks who don't really have an "time of first spike" concept.

src/roiextractors/segmentationextractor.py
h-mayorquin: Great.

src/roiextractors/segmentationextractor.py
h-mayorquin: Looks good!

tests/test_internals/test_frame_slice_segmentation.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+    @parameterized.expand([param(name="mean"), param(name="correlation")], name_func=segmentation_name_function)
+    def test_get_image(self, name: str):
+        assert_array_equal(
+            x=self.frame_sliced_segmentation.get_image(name=name), y=self.toy_segmentation_example.get_image(name=name)
+        )
+
+
+class TestMissingTraceFrameSlicesegmentation(BaseTestFrameSlicesegmentation):
```
h-mayorquin, line 97: I did not think on this pattern, it is neat.

tests/test_internals/test_frame_slice_segmentation.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def segmentation_name_function(testcase_function, param_number, param):
+    return f"{testcase_function.__name__}_{param_number}_{parameterized.to_safe_name(param.kwargs['name'])}"
+
+
+class BaseTestFrameSlicesegmentation(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.toy_segmentation_example = generate_dummy_segmentation_extractor(num_frames=15, num_rows=5, num_columns=4)
```
h-mayorquin, line 36: I see that you have some tests for rejected and accepted list. Up to your but the `generate_dummy_segmentation_extractor` has a rejected list argument to generate that:

https://github.com/catalystneuro/roiextractors/blob/b71f17db4d3dadeaa4ae02c806c1e269f03d3d9b/src/roiextractors/testing.py#L54-L65

So you can test the non-default case when `rejected_list != [ ]`

h-mayorquin (approved):
Looks good to me, one minor suggestion if you want to add.

tests/test_internals/test_frame_slice_segmentation.py
CodyCBakerPhD: Multiplies the total number of tests run when doing it, but going for maximum safety here.

tests/test_internals/test_frame_slice_segmentation.py
CodyCBakerPhD: Yeah, I could. But I actually have yet to see any format in practice that bothers to save ROI info that has not been accepted. (so far, every single ROI written as series data or image mask is 'accepted' so 'rejected' is always empty). Maybe something to think about and revisit in the future

PR #203, opened 2022-08-31
Debug to edge case of MultiImaging

CodyCBakerPhD:
@weiglszonja Does this fix your issue?

### DIFF
--- a/src/roiextractors/multiimagingextractor.py
+++ b/src/roiextractors/multiimagingextractor.py
@@ -113,7 +113,7 @@ class MultiImagingExtractor(ImagingExtractor):
 
         start = start_frame if start_frame is not None else 0
         stop = end_frame if end_frame is not None else self.get_num_frames()
-        extractors_range = np.searchsorted(self._end_frames, (start, stop), side="right")
+        extractors_range = np.searchsorted(self._end_frames, (start, stop - 1), side="right")
         extractors_spanned = list(
             range(extractors_range[0], min(extractors_range[-1] + 1, len(self._imaging_extractors)))
         )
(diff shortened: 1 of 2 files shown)


weiglszonja:
Alternative solution:

```
np.where(np.logical_and(np.array(self._end_frames) > start, np.array(self._start_frames) < stop),)[0]
```
@CodyCBakerPhD 
`timeit` results with the alternative solution, roughly 2 times slower than `np.searchsorted`!
start_frames, end_frames are arrays with 10000 values, looking up a range of (4800, 6000)

np.searchsorted(end_frames, (start, stop), side="right") -> 0.002 s
np.where(np.logical_and(np.array(end_frames) > start, np.array(start_frames) < stop)) -> 0.004 s

To replicate:
```python
import timeit

setup = """
import numpy as np

start_frames = np.arange(0, 10000, 1000)
end_frames = np.arange(0, 10000, 1000) + 1000

start = 4800
stop = 6000

"""

functions = [
    'np.searchsorted(end_frames, (start, stop), side="right")',
    'np.where(np.logical_and(np.array(end_frames) > start, np.array(start_frames) < stop))',
]

rdict = {}
for i in functions:
    rdict[i] = timeit.timeit(i, setup=setup, number=1000)
    print("%s -> %s s" %(i,rdict[i]))
```

CodyCBakerPhD:
> np.searchsorted(end_frames, (start, stop), side="right") -> 0.002 s
> np.where(np.logical_and(np.array(end_frames) > start, np.array(start_frames) < stop)) -> 0.004 s

Keeping things in perspective, those are still both very, very fast. As long as it's not an exponential slowdown it's perfectly fine.

So the question is then, which option solves your issue?

weiglszonja:
> So the question is then, which option solves your issue?

I just tried yours, I managed to run the conversion 💯 , I don't have a strong preference for the alternative version, so we can merge this one as well.

weiglszonja (approved):
Thanks for fixing this!

Issue #204, opened 2022-08-31
Finalize droping `spikextractors`, remove its use in `toy_example.py`


h-mayorquin:
On #196 @CodyCBakerPhD did most of the hard work of removing the spikextractors as a dependency and pushed it as an optional work flow. 

What is left is to eliminate this completely. In the toy example `spike-extractor` is used to generate a set of frames in which the spikes occur to transform them to roi  (that is, not the spikes times but the actual frames, given a sampling frequency, in which the spikes occurred). We should be able to reproduce this functionality here and eliminate the dependency for good.

So, where is this in the code, the actual functions from spikeextractors are here:
https://github.com/catalystneuro/roiextractors/blob/b71f17db4d3dadeaa4ae02c806c1e269f03d3d9b/src/roiextractors/example_datasets/toy_example.py#L126-L131

A recording and a spikextractor. They only use is here:
https://github.com/catalystneuro/roiextractors/blob/b71f17db4d3dadeaa4ae02c806c1e269f03d3d9b/src/roiextractors/example_datasets/toy_example.py#L140-L151

Note that the the recording extractor (rec in the code) is only used to get the sampling frequency and the number of frames which is not really necessary. I will re-structure this in a small PR to make this obvious. The only difficult thing to remove is the production of the spike trains in line 144:

https://github.com/catalystneuro/roiextractors/blob/b71f17db4d3dadeaa4ae02c806c1e269f03d3d9b/src/roiextractors/example_datasets/toy_example.py#L144

So, a function that produces frames in which spikes occur in some reasonable fashion should suffice for this small toy example.

h-mayorquin:
In #205 I added indications of what needs to be done for when someone has time to take this on:
https://github.com/catalystneuro/roiextractors/blob/5b741335acdc3e7067020e63614f5c017578095f/src/roiextractors/example_datasets/toy_example.py#L140-L152

h-mayorquin:
This has long been settled.

PR #205, opened 2022-08-31
Make `toy_example.py` dependency on spikeextractors clearer 

h-mayorquin:
This is the first small step on to accomplish the goals described on #204. It is just changes that makes it clear what should be done to complete issue #204 .

### DIFF
--- a/src/roiextractors/example_datasets/toy_example.py
+++ b/src/roiextractors/example_datasets/toy_example.py
@@ -109,7 +109,6 @@ def toy_example(
         The output segmentation extractor
 
     """
-    import spikeextractors as se
 
     # generate ROIs
     num_rois = int(num_rois)
@@ -123,6 +122,8 @@ def toy_example(
     )
 
     # generate spike trains
+    import spikeextractors as se
+
     rec, sort = se.example_datasets.toy_example(
         duration=duration,
         K=num_rois,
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
I was under the impression that the spikeextractors toy example utilized a lot of over-arching calls to things like generating waveforms and such. Are you saying we can just generate some dummy spike trains here and call it good?

CodyCBakerPhD:
Though I wouldn't say it's the responsibility of our package to accurately generate biophysically realistic (or even statistically consistent) synthetic data. We just need example data structure to test and demonstrate usage that can be used in lieu of real data

h-mayorquin:
> I was under the impression that the spikeextractors toy example utilized a lot of over-arching calls to things like generating waveforms and such. Are you saying we can just generate some dummy spike trains here and call it good?

Yep.

I agree that we should not care too much about realistic spikes.

h-mayorquin:
> > I was under the impression that the spikeextractors toy example utilized a lot of over-arching calls to things like generating waveforms and such. Are you saying we can just generate some dummy spike trains here and call it good?
> 
> Yep.
> 
> I agree that we should not care too much about realistic spikes. The toy example should be used for education / tutorial purposes anyway, right?

CodyCBakerPhD:
> The toy example should be used for education / tutorial purposes anyway, right?

Yeah, I believe the original motivation was the have 'realistic looking' output in the widgets applied to a demo file but... I don't know if they have to be THAT realistic looking... simple Poisson processes would be fine by me

h-mayorquin:
Can we merge this? I just wanted to leave the terrain prepare for someone -maybe me later- to take this on and reference this from issue #204.

CodyCBakerPhD (approved):

PR #207, opened 2022-08-31
Finish correcting segmentation orientation

### LINKED ISSUE #206: Suite2pSegmentationExtractor still uses (num_rois, num_frames) format for its traces
We have changed this in #200 but the following code still raises an error

```python
folder_path=str(OPHYS_DATA_PATH / "segmentation_datasets" / "suite2p")
extractor = Suite2pSegmentationExtractor(folder_path=folder_path)

assert extractor.get_traces(name="raw").shape == (extractor.get_num_frames(),extractor.get_num_rois())
```

More concretely:
```
extractor.get_num_frames(), extractor.get_num_rois(), extractor.get_traces(name="raw").shape
(250, 15, (15, 250))
```

CodyCBakerPhD:
fix #206  ?

@h-mayorquin Unable to reproduce it in the tests, can you write a failing test here that works like you're seeing?

### DIFF
--- a/src/roiextractors/segmentationextractor.py
+++ b/src/roiextractors/segmentationextractor.py
@@ -173,14 +173,13 @@ class SegmentationExtractor(ABC):
         """
         if name not in self.get_traces_dict():
             raise ValueError(f"traces for {name} not found, enter one of {list(self.get_traces_dict().keys())}")
-        if roi_ids is None:
-            roi_idx_ = range(self.get_num_rois())
-        else:
+        if roi_ids is not None:
             all_ids = self.get_roi_ids()
-            roi_idx_ = [int(all_ids.index(i)) for i in roi_ids]
+            roi_idxs = [all_ids.index(i) for i in roi_ids]
         traces = self.get_traces_dict().get(name)
         if traces is not None and len(traces.shape) != 0:
-            return np.array([traces[start_frame:end_frame, idx] for idx in roi_idx_])
+            idxs = slice(None) if roi_ids is None else roi_idxs
+            return np.array(traces[start_frame:end_frame, :])[:, idxs]  # numpy fancy indexing is quickest
 
     def get_traces_dict(self):
         """
--- a/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
+++ b/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
@@ -158,13 +158,13 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
 
         # saving traces:
         if segmentation_object.get_traces(name="raw") is not None:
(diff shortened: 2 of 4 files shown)


h-mayorquin:
The same piece of code that I provided in #206 is still failing here. The functions that you added to the tests (which are great!) use the `get_traces_dict` method. But the error occurs with  the `get_traces` method directly. The problem is that those not consistent because the method transposes the traces here:

https://github.com/catalystneuro/roiextractors/blob/5b741335acdc3e7067020e63614f5c017578095f/src/roiextractors/segmentationextractor.py#L183

Thinking about this, I don't know why we have two methods that do essentially the same thing.

src/roiextractors/segmentationextractor.py
```diff
@@ -173,14 +173,13 @@ def get_traces(self, roi_ids=None, start_frame=None, end_frame=None, name="raw")
...
-        else:
+        if roi_ids is not None:
             all_ids = self.get_roi_ids()
-            roi_idx_ = [int(all_ids.index(i)) for i in roi_ids]
+            roi_idxs = [all_ids.index(i) for i in roi_ids]
         traces = self.get_traces_dict().get(name)
         if traces is not None and len(traces.shape) != 0:
-            return np.array([traces[start_frame:end_frame, idx] for idx in roi_idx_])
+            idxs = slice(None) if roi_ids is None else roi_idxs
+            return np.array(traces[start_frame:end_frame, :])[:, idxs]  # numpy fancy indexing is quickest
```
CodyCBakerPhD, line 182: Well, it's pretty subtle, but this is the reason it wasn't working. For whatever reason, the return was always an array casting of a list compression whose outer level is the ROI's.

CodyCBakerPhD: I mean, applying the fancy indexing for subset ROI's makes more sense to do to something you know is already a numpy array (some traces slicing might apply to `h5py.Dataset` objects) so I think that's covered here without the use of `np.stack`.

h-mayorquin: Good point. Is the casting to `np.array` actually needed over there?

CodyCBakerPhD: In the event that the underlying `traces` are an `h5py.Dataset` I think it makes it clearer/more readable to cast an array for maximum safety to communicate that it is the `np.ndarray` which has the fancy indexing applied to it

h-mayorquin: All right.

h-mayorquin (approved):

src/roiextractors/segmentationextractor.py
CodyCBakerPhD: I do think there are probably even better ways around everything in this function. Probably in a follow-up

PR #208, opened 2022-08-31
Debugs for the NWBSegmentationExtractor

CodyCBakerPhD:
These ultimately affect the round-trip tests on NeuroConv - we should look into adding an NWBFile example on the ophys GIN so we can add it into the `test_io` here.

### DIFF
--- a/src/roiextractors/extractors/nwbextractors/nwbextractors.py
+++ b/src/roiextractors/extractors/nwbextractors/nwbextractors.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Union, Optional
+from typing import Union, Optional, Iterable
 
 import numpy as np
 from lazy_ops import DatasetView
@@ -281,7 +281,7 @@ class NwbSegmentationExtractor(SegmentationExtractor):
                 setattr(
                     self,
                     f"_roi_response_{trace_name}",
-                    DatasetView(container.roi_response_series[trace_name_segext].data).lazy_transpose(),
+                    DatasetView(container.roi_response_series[trace_name_segext].data),
                 )
                 if self._sampling_frequency is None:
                     self._sampling_frequency = container.roi_response_series[trace_name_segext].rate
@@ -298,7 +298,7 @@ class NwbSegmentationExtractor(SegmentationExtractor):
             ps = image_seg.plane_segmentations["PlaneSegmentation"]
             assert "image_mask" in ps.colnames, "Could not find any image_masks in nwbfile."
             self._image_masks = DatasetView(ps["image_mask"].data).lazy_transpose([2, 1, 0])
-            self._roi_locs = ps["RoiCentroid"] if "RoiCentroid" in ps.colnames else None
+            self._roi_locs = ps["ROICentroids"] if "ROICentroids" in ps.colnames else None
             self._accepted_list = ps["Accepted"].data[:] if "Accepted" in ps.colnames else None
             self._rejected_list = ps["Rejected"].data[:] if "Rejected" in ps.colnames else None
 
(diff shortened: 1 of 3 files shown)


src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -298,7 +298,7 @@ def __init__(self, file_path: PathType):
             ps = image_seg.plane_segmentations["PlaneSegmentation"]
             assert "image_mask" in ps.colnames, "Could not find any image_masks in nwbfile."
             self._image_masks = DatasetView(ps["image_mask"].data).lazy_transpose([2, 1, 0])
-            self._roi_locs = ps["RoiCentroid"] if "RoiCentroid" in ps.colnames else None
+            self._roi_locs = ps["ROICentroids"] if "ROICentroids" in ps.colnames else None
```
CodyCBakerPhD, line 301: New from https://github.com/catalystneuro/neuroconv/pull/118

src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, file_path: PathType):
                 setattr(
                     self,
                     f"_roi_response_{trace_name}",
-                    DatasetView(container.roi_response_series[trace_name_segext].data).lazy_transpose(),
+                    DatasetView(container.roi_response_series[trace_name_segext].data),
```
CodyCBakerPhD, line 284: Missed in #200

src/roiextractors/extractors/nwbextractors/nwbextractors.py
```diff
@@ -336,10 +336,14 @@ def get_images_dict(self):
 
         return images_dict
 
-    @property
-    def roi_locations(self):
-        if self._roi_locs is not None:
-            return self._roi_locs.data[:].T
```
CodyCBakerPhD, line 342: This method was never working properly - it was always leading to the base method anyway

h-mayorquin:
> These ultimately affect the round-trip tests on NeuroConv - we should look into adding an NWBFile example on the ophys GIN so we can add it into the `test_io` here.

We have this for the ImagingExtractor, we should also add the same for segmentation.
https://github.com/catalystneuro/roiextractors/blob/master/tests/test_internals/test_nwb_extractors.py

CodyCBakerPhD:
> We have this for the ImagingExtractor, we should also add the same for segmentation.

I'd feel a lot better about using an actual file. There's some very particular nuances about initializing the traces from an actual NWB file that not be easily replicable with a mock

h-mayorquin:
> > We have this for the ImagingExtractor, we should also add the same for segmentation.
> 
> I'd feel a lot better about using an actual file. There's some very particular nuances about initializing the traces from an actual NWB file that not be easily replicable with a mock

What do you mean? That's not a mock, (or maybe you mean something different by mock?)  A file is created, saved and loaded.
What would be those nuances?

CodyCBakerPhD:
> What do you mean? That's not a mock, (or maybe you mean something different by mock?) A file is created, saved and loaded.

I'm mentally equating, or possible conflating, a dummy file (albeit, it's on disk) with a hardcoded structure with a Mock file class (albeit, that's only in-memory) that has the same attribute structure. Principle of use is kind of the same +/- going through PyNWB.

> What would be those nuances?

Like writing the ROICentroids column (not required by PyNWB or NWB Schema, but done by default through NeuroConv), which it nonetheless tries to load in over here on ROIExtractors. And the choice of series names, which ROIExtractors asssumes matches the NeuroConv convention.

Basically, our NwbSegmentationExtractor is more of a NeuroConvSegmentationExtractor as it is right now, so using a manually constructed NWB file with PyNWB would only work if that hard-coded structure exactly matched what NeuroConv outputs by default.

Don't get me wrong this should all be fixed, but that's going to be a much bigger PR. Can I add those tests then? (this PR is for required for things like https://github.com/catalystneuro/neuroconv/pull/120 to pass)

h-mayorquin:
Oh yeah, sure, this 

> > What do you mean? That's not a mock, (or maybe you mean something different by mock?) A file is created, saved and loaded.
> 
> I'm mentally equating, or possible conflating, a dummy file (albeit, it's on disk) with a hardcoded structure with a Mock file class (albeit, that's only in-memory) that has the same attribute structure. Principle of use is kind of the same +/- going through PyNWB.
> 
> > What would be those nuances?
> 
> Like writing the ROICentroids column (not required by PyNWB or NWB Schema, but done by default through NeuroConv), which it nonetheless tries to load in over here on ROIExtractors. And the choice of series names, which ROIExtractors asssumes matches the NeuroConv convention.
> 
> Basically, our NwbSegmentationExtractor is more of a NeuroConvSegmentationExtractor as it is right now, so using a manually constructed NWB file with PyNWB would only work if that hard-coded structure exactly matched what NeuroConv outputs by default.

Even more so, we can just code this here as well. In fact, having a file in gin would be harder to modify if the schema of neuroconv changes and we can just update it very simply here. I really don't see what do we gain from uploading if we are going to build it ourselves. We just need the structure in code to conform to the structure in neuroconv.

> Don't get me wrong this should all be fixed, but that's going to be a much bigger PR. Can I add those tests then? (this PR is for required for things like [catalystneuro/neuroconv#120](https://github.com/catalystneuro/neuroconv/pull/120) to pass)

Yes, sure, those comments are tangential and more about long-term.

h-mayorquin (approved):

PR #210, opened 2022-09-05
ExtractSegmentationExtractor improvements

weiglszonja:
Connected to https://github.com/catalystneuro/neuroconv/pull/128, raise an error when `output_struct_name` is not found in the .mat file. 
- changed assumed newer version of extract files to "1.0.0"
- transpose segmentation images to have height x width for consistency


CodyCBakerPhD:
> transpose segmentation images to have height x width for consistency

From the tutorial at: https://github.com/schnitzer-lab/EXTRACT-public/blob/master/Learning%20materials/Lecture%20Tutorials/Tutorial%201%20-%20Starting%20code/example_tutorial.m#L38

it would seem the spatial weights already follow ROIExtractors convention

weiglszonja:
@CodyCBakerPhD leaving here as a reminder
when reading extract data in MATLAB, these images would read as `height x width`; however in `h5py` they would read as `width x height` by default, hence the transpose is necessary in this case to be consistent within extractors.

CodyCBakerPhD (approved):

PR #212, opened 2022-09-07
Relax rounding of frame_to_time()

CodyCBakerPhD:
For fixing https://github.com/catalystneuro/neuroconv/pull/130 and also just ensuring a closer behavior of our timestamps to those found in SpikeInterface: https://github.com/catalystneuro/spikeinterface/blob/master/spikeinterface/core/baserecording.py#L448-L449


weiglszonja:
It seems like there is still a rounding issue left in the tests:
https://github.com/catalystneuro/neuroconv/runs/8233084881?check_suite_focus=true
`assert_array_almost_equal` the default is 6 decimals.

CodyCBakerPhD:
> It seems like there is still a rounding issue left in the tests:

Yeah I saw. I guess it doesn't entirely fix the issue there but at least it improves the accuracy.

I still think we should mimic SpikeInterface as much as possible here though. (eventually even moving to `set_times` and `get_times`)

@weiglszonja What do you think?

PR #215, opened 2022-10-19
Debug masks for SubFrameSegmentation

CodyCBakerPhD:
@h-mayorquin Turns out that `_image_masks` problem occurred again, but only after I enabled stubbing for my segmentation interfaces.

This fixes it though.

### DIFF
--- a/src/roiextractors/segmentationextractor.py
+++ b/src/roiextractors/segmentationextractor.py
@@ -1,7 +1,8 @@
 from abc import ABC, abstractmethod
-from typing import Union, Optional, Tuple, Iterable
+from typing import Union, Optional, Tuple, Iterable, List
 
 import numpy as np
+from numpy.typing import ArrayLike
 
 from .extraction_tools import ArrayType, IntType, FloatType
 from .extraction_tools import _pixel_mask_extractor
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Upcoming
 
+### Fixes
+
+* Fixed two issue with the `SubFrameSegementation` class: (i) attempting to set the private attribute `_image_masks` even when this was not present in the parent, and (ii) not calling the parent function for `get_pixel_masks` and instead using the base method even in cases where this had been overridden by the parent. [PR #215](https://github.com/catalystneuro/roiextractors/pull/215)
```
h-mayorquin, line 5:
```suggestion
* Fixed two issues with the `SubFrameSegementation` class: (i) attempting to set the private attribute `_image_masks` even when this was not present in the parent, and (ii) not calling the parent function for `get_pixel_masks` and instead using the base method even in cases where this had been overridden by the parent. [PR #215](https://github.com/catalystneuro/roiextractors/pull/215)
```

h-mayorquin (commented):
Do we have any test for this?

CodyCBakerPhD:
> Do we have any test for this?

Added

h-mayorquin (approved):

PR #220, opened 2023-03-31
add `BrukerTiffImagingExtractor`

weiglszonja:
Add support for Bruker 2P imaging data which consists of:
- individual TIFF files in `OME-TIF` format (`.ome.tif` files) where each TIFF file corresponds to a single frame
- metadata in XML format (`.xml` and `.env` files)

The metadata from the .xml file in dict form for reference:
```python
{
'version': '5.6.64.400',
'date': '2/20/2023 3:58:25 PM',
'notes': '',
'activeMode': 'ResonantGalvo',
'bitDepth': '13',
'currentScanAmplitude': [{'XAxis': '4.17170481'}, {'YAxis': '4.17170481'}],
'currentScanCenter': [{'XAxis': '0'}, {'YAxis': '0'}],
'dwellTime': '0.4',
'framePeriod': '0.032953338',
'interlacedScanTrackCount': '0',
'laserPower': [{'920 Axon': '500'}, {'1064 Axon': '900.390625'}, {'Monaco': '0'}],
'linesPerFrame': '512',
'maxVoltage': [{'XAxis': '2.085852405'}, {'YAxis': '2.085852405'}],
'micronsPerPixel': [{'XAxis': '1.1078125'}, {'YAxis': '1.1078125'}, {'ZAxis': '5'}],
'minVoltage': [{'XAxis': '-2.085852405'}, {'YAxis': '-2.085852405'}],
'objectiveLens': '16X', 'objectiveLensMag': '16', 'objectiveLensNA': '0.8',
'opticalZoom': '2',
'pixelsPerLine': '512',
'pmtGain': [{'PMT 1 HV (Red)': '800'}, {'PMT 2 HV (Green)': '700'}],
'positionCurrent': [{'XAxis': '0'}, {'YAxis': '0'}, {'Z Focus': '147.55'},
                     {'Optotune ETL 10-30': '-80.0653860309135'}],
 'preampFilter': 'NoFilter',
 'preampOffset': [{'Ch1': '0'}, {'Ch2': '0'}],
 'rastersPerFrame': '1',
 'rotation': '0',
 'samplesPerPixel': '3',
 'scanLinePeriod': '6.3129E-05',
 'useInterlacedScanPattern': 'False',
 'xYStageGridIndex': '-1', 'xYStageGridXIndex': '-1', 'xYStageGridYIndex': '-1',
 'yAspectExpansion': '1', 'zDevice': '0'}
```

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
@@ -112,8 +112,10 @@ class BrukerTiffImagingExtractor(ImagingExtractor):
         start_frame: Optional[int] = None,
         end_frame: Optional[int] = None,
     ):
+        tiffile = get_package(package_name="tifffile", installation_instructions="pip install tifffile")
+
         for file in self._file_paths[start_frame:end_frame]:
-            yield _get_tiff_reader().memmap(self.folder_path / file, mode="r", _multifile=False)
+            yield tiffile.memmap(self.folder_path / file, mode="r", _multifile=False)
 
     def get_video(
         self, start_frame: Optional[int] = None, end_frame: Optional[int] = None, channel: int = 0
--- a/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
@@ -40,26 +40,33 @@ class BrukerTiffImagingExtractor(ImagingExtractor):
 
         super().__init__()
         self.folder_path = Path(folder_path)
-
-        sequences = self._get_xml_root().findall("Sequence")
-        num_sequences = len(sequences)
-        # TODO: extend it to general plane index, but we need example data for it first
-        self._sequence = sequences[0]
-        self._num_frames = len(self._sequence.findall("./Frame"))
-
-        files = self._sequence.findall(".//File")
-        self._file_paths = [file.attrib["filename"] for file in files]
         tif_file_paths = list(self.folder_path.glob("*.ome.tif"))
         assert tif_file_paths, f"The TIF image files are missing from '{self.folder_path}'."
-        if num_sequences != 1:
-            missing_files = [
-                file_path for file_path in self._file_paths if self.folder_path / file_path not in tif_file_paths
-            ]
-            assert (
-                not missing_files
-            ), f"Some of the TIF image files at '{self.folder_path}' are missing for this plane. The list of files that are missing: {missing_files}"
+
+        xml_root = self._get_xml_root()
+        sequences = xml_root.findall("Sequence")
+        num_sequences = len(sequences)
+        num_frames_first_sequence = len(sequences[0].findall("./Frame"))
+        # The number of "Sequence" elements in the document determine whether multiple z-planes are present.
+        if num_sequences == 1:
+            self._num_frames = num_frames_first_sequence
+            self._num_z_planes = 1
         else:
-            assert len(self._file_paths) == len(
+            # Note: could it happen that for the last sequence we have less frames then for the first one?
+            num_frames_last_sequence = len(sequences[-1].findall("Frame"))
+            if num_frames_first_sequence != num_frames_last_sequence:
+                raise NotImplementedError(
+                    f"Not sure how to handle final stack because it was found with fewer z-planes ({num_frames_last_sequence}, expected: {num_frames_first_sequence}).",
+                )
+            self._num_frames = num_sequences
+            self._num_z_planes = num_frames_first_sequence
+
+        # The ordered list of files in the XML document
+        files = xml_root.findall(".//File")
+        self._file_paths = [file.attrib["filename"] for file in files]
+        tif_file_paths = list(self.folder_path.glob("*.ome.tif"))
+
+        assert len(self._file_paths) == len(
                 tif_file_paths
             ), f"The number of TIF image files at '{self.folder_path}' should be equal to the number of frames ({len(self._file_paths)}) specified in the XML configuration file."
 
@@ -78,7 +85,7 @@ class BrukerTiffImagingExtractor(ImagingExtractor):
         """
         Parses the XML configuration file into element tree and returns the root of this tree.
         """
-        xml_file_path = self.folder_path / f"{self.folder_path.stem}.xml"
+        xml_file_path = self.folder_path / f"{self.folder_path.name}.xml"
         assert Path(xml_file_path).is_file(), f"The XML configuration file is not found at '{self.folder_path}'."
         tree = ElementTree.parse(xml_file_path)
         return tree.getroot()
@@ -120,8 +127,8 @@ class BrukerTiffImagingExtractor(ImagingExtractor):
                                 )
         return xml_metadata
 
-    def get_image_size(self) -> Tuple[int, int]:
-        return self._height, self._width
+    def get_image_size(self) -> Tuple[int, int, int]:
+        return self._height, self._width, self._num_z_planes
 
     def get_num_frames(self) -> int:
         return self._num_frames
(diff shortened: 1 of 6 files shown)


weiglszonja:
**TODO**: add tests (haven't found test data to upload to [test](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/imaging_datasets) folder)
**TODO**: Bruker metadata extension? with the metadata shown in the description

Sample data from Pinto upload is in progress ([sub-NCCR32/sub-NCCR32_ses-20230331T153324_ophys.nwb](https://gui-staging.dandiarchive.org/dandiset/204134/draft/files?location=))

CodyCBakerPhD:
> TODO: Bruker metadata extension? with the metadata shown in the description

How long would that take compared to writing a `DynamicTable` with a single row, whose columns are each of the fields? (both approaches include descriptions of each field, but the data type would be auto-inferred in the table case)

CodyCBakerPhD:
> TODO: add tests (haven't found test data to upload to [test](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/imaging_datasets) folder)

How big is this session of data? Would including only a handful of the first tiffs (basically a stub) cause a problem for the reader? (like a mismatch with something expected from the header)

weiglszonja:
> > TODO: Bruker metadata extension? with the metadata shown in the description
> 
> How long would that take compared to writing a `DynamicTable` with a single row, whose columns are each of the fields? (both approaches include descriptions of each field, but the data type would be auto-inferred in the table case)

You're right, I haven't considered that approach, makes sense!


> > TODO: add tests (haven't found test data to upload to [test](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/imaging_datasets) folder)
> 
> How big is this session of data? Would including only a handful of the first tiffs (basically a stub) cause a problem for the reader? (like a mismatch with something expected from the header)

~11.81GB on disk total. Currently it would, but it can be changed that I ignore what is in the header and assume the number of frames from the tif files within the folder (but also make sure they are in consecutive order, which is the case when collecting them from the header)

CodyCBakerPhD:
> Currently it would, but it can be changed that I ignore what is in the header and assume the number of frames from the tif files within the folder (but also make sure they are in consecutive order, which is the case when collecting them from the header)

What field of the header describes how many frames to expect?

bendichter:
I think it's better to build an extension for Bruker rather than using a DynamicTable

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -0,0 +1,123 @@
...
+        return [file.attrib["filename"] for file in self._get_xml_root().findall(".//File")]
+
+    def get_image_size(self) -> Tuple[int, int]:
+        return self._height, self._width
+
+    def get_num_frames(self) -> int:
+        return len(self._file_paths)
+
+    def get_sampling_frequency(self) -> float:
+        frame_period = float(self.xml_metadata["framePeriod"])
```
CodyCBakerPhD, line 95: Minor point - I would recommend extracting and storing this during `__init__` as `_sampling_frequency` so you don't have to keep recalculating it every time this getter is called

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -0,0 +1,123 @@
...
+        return 1 / frame_period
+
+    def get_channel_names(self) -> list:
+        channel_names = [file.attrib["channelName"] for file in self._get_xml_root().findall(".//File")]
+        unique_channel_names = list(set(channel_names))
+        return unique_channel_names
```
CodyCBakerPhD, line 101: Is this actually multi-channel data?

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -0,0 +1,123 @@
...
+    def get_dtype(self) -> DtypeType:
+        return self._dtype
+
+    def _frames_iterator(
+        self,
+        start_frame: Optional[int] = None,
+        end_frame: Optional[int] = None,
+    ):
+        for file in self._file_paths[start_frame:end_frame]:
+            yield _get_tiff_reader().memmap(self.folder_path / file, mode="r", _multifile=False)
```
CodyCBakerPhD, line 116:
```suggestion
    ):
        tiffile = get_package(package_name="tifffile", installation_instructions="pip install tifffile")

        for file in self._file_paths[start_frame:end_frame]:
            yield tiffile.memmap(self.folder_path / file, mode="r", _multifile=False)
```

reads a little awkward to re-import on each iteration; I know it technically won't affect performance hardly at all but this should work the same way

CodyCBakerPhD:
> It is not a single field, each frame has an entry like this:

Interesting...

But right now you decide the iteration from

```python
    def _get_files_names(self):
        return [file.attrib["filename"] for file in self._get_xml_root().findall(".//File")]
```

which seems similar to globbing...

From there, if you wanted to confirm that the `index` of the last `Frame` matched the number of files from globbing, it would only be to ensure the two are consistent, right?

So to create a stub, we would remove all the `frame` subfields whose indexes correspond to the files we would remove as well? Since there isn't a global counter that seems a little easier

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -0,0 +1,123 @@
...
+                            {indexed_value.attrib["description"]: indexed_value.attrib["value"]}
+                        )
+                    elif "value" in indexed_value.attrib:
+                        xml_metadata[child.attrib["key"]].append(
+                            {indexed_value.attrib["index"]: indexed_value.attrib["value"]}
+                        )
+        return xml_metadata
+
+    def _get_files_names(self):
+        return [file.attrib["filename"] for file in self._get_xml_root().findall(".//File")]
```
CodyCBakerPhD, line 86: Just to make me feel safer, the `.findall` is guaranteed to iterate a 'list' of properly ordered file (frame) order?

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -0,0 +1,123 @@
...
+        self,
+        start_frame: Optional[int] = None,
+        end_frame: Optional[int] = None,
+    ):
+        for file in self._file_paths[start_frame:end_frame]:
+            yield _get_tiff_reader().memmap(self.folder_path / file, mode="r", _multifile=False)
+
+    def get_video(
+        self, start_frame: Optional[int] = None, end_frame: Optional[int] = None, channel: int = 0
+    ) -> np.ndarray:
```
CodyCBakerPhD, line 120: If this is single-channel and we don't yet have an example of multi-channel, I would recommend adding `if channel != 0: raise NotImplementedError` to be safe

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
weiglszonja: You're right, thanks!

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
weiglszonja: The data that we currently has only one channel and is called `"Ch2"`. Since I'm not handling multi-channel data in `get_video()` I guess I can replace `.findall` to `.find`, and then it returns the first entry.

CodyCBakerPhD: `.findall` is fine to leave as a generality

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
CodyCBakerPhD: Cool

CodyCBakerPhD:
Last 2 items then are

1) docstrings

and

2) tests (requires a stubbed session)

CodyCBakerPhD:
@weiglszonja OK CI now has the data on its cache

Just running into some actual errors now

CodyCBakerPhD:
> Can we change it here too?

Do w/e you need to do to get the CI passing here, yeah

CodyCBakerPhD (approved):

PR #222, opened 2023-04-21
Add `MicroManagerTiffImagingExtractor`

weiglszonja:
Add extractor for the Micro-Manager TIF image format:
- image file stacks are saved into multipage TIF files in OME-TIFF format (.ome.tif files), each of which are up to around 4GB in size.
- The 'DisplaySettings' JSON file contains the properties of Micro-Manager.

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/micromanagertiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/micromanagertiffimagingextractor.py
@@ -2,7 +2,6 @@ import json
 import logging
 import re
 from collections import Counter
-from encodings.utf_8 import decode
 from itertools import islice
 from pathlib import Path
 from types import ModuleType
--- a/src/roiextractors/extractors/tiffimagingextractors/micromanagertiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/micromanagertiffimagingextractor.py
@@ -4,7 +4,7 @@ from collections import Counter
 from itertools import islice
 from pathlib import Path
 from types import ModuleType
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Dict
 
 from xml.etree import ElementTree
 import numpy as np
@@ -26,8 +26,9 @@ class MicroManagerTiffImagingExtractor(MultiImagingExtractor):
     def __init__(self, folder_path: PathType):
         """
         The imaging extractor for the Micro-Manager TIF image format.
-        This format consists of multiple TIF image files in multipage OME-TIF format (.ome.tif files)
-        and 'DisplaySettings' JSON file.
+        The image file stacks are saved into multipage TIF files in OME-TIFF format (.ome.tif files),
+        each of which are up to around 4GB in size.
+        The 'DisplaySettings' JSON file contains the properties of Micro-Manager.
 
         Parameters
         ----------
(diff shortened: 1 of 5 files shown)


src/roiextractors/extractors/tiffimagingextractors/micromanagertiffimagingextractor.py
```diff
@@ -0,0 +1,182 @@
...
+            raise NotImplementedError(
+                f"The {self.extractor_name}Extractor does not currently support multiple color channels."
+            )
+        if start_frame is not None and end_frame is not None and start_frame == end_frame:
+            page = self.pages[start_frame]
+            return page.asarray()[np.newaxis, ...]
+
+        pages = self.pages[start_frame:end_frame]
+        frames = [page.asarray() for page in pages]
+        return np.stack(frames)
```
CodyCBakerPhD, line 182: Can be improved as per our meeting

tests/test_micromanagertiffimagingextractor.py
```diff
@@ -0,0 +1,112 @@
+import shutil
+import tempfile
+from pathlib import Path
```
CodyCBakerPhD, line 3:
```suggestion
from warnings import warn
```

tests/test_micromanagertiffimagingextractor.py
```diff
@@ -0,0 +1,117 @@
+import shutil
+import tempfile
+from warnings import warn
```
CodyCBakerPhD, line 3:
```suggestion
from pathlib import Path
from warnings import warn
```

src/roiextractors/multiimagingextractor.py
```diff
@@ -127,7 +127,7 @@ def get_video(
                 start_frame=relative_start, end_frame=relative_stop
             )
 
-        video_shape = (stop - start,) + self._imaging_extractors[0].get_image_size()
+        video_shape = (stop - start,) + self.get_image_size()
```
CodyCBakerPhD, line 130: Why do you need to change this?

Since it's an abstract method wouldn't you need to also define `get_image_size` generally for the `MultiImagingExtractor`?

I'd prefer if you can leave the base class here unchanged and do what you need to in the child classes, is that possible?

CodyCBakerPhD:
Overall looks good, one comment that has the potential to break back-compatibility (though I admit it's likely not being tested explicitly)

src/roiextractors/multiimagingextractor.py
weiglszonja: you're right, it's not necessary to change, better to keep it as is.

tests/test_micromanagertiffimagingextractor.py
```diff
@@ -0,0 +1,116 @@
...
+                frames.append(tif.asarray(key=range(5)))
+        return np.concatenate(frames, axis=0)
+
+    @classmethod
+    def tearDownClass(cls):
+        try:
+            (Path(cls.test_dir) / cls.file_paths[0]).unlink()
+            shutil.rmtree(cls.test_dir)
+        except PermissionError:  # Windows
+            warn(f"Unable to cleanup testing data at {cls.test_dir / cls.file_paths[0]}! Please remove them manually.")
```
CodyCBakerPhD, line 49:
```suggestion
        try:
            shutil.rmtree(cls.test_dir)
        except PermissionError:  # Windows
            warn(f"Unable to cleanup testing data at {cls.test_dir}! Please remove it manually.")
```

CodyCBakerPhD (approved):

PR #223, opened 2023-05-05
Rename `MicroManagerTiffImagingExtractor` channel to 'OpticalChannelDefault'

weiglszonja:
Micro-Manager uses "Default" as channel name, it is better to rename it to 'OpticalChannelDefault' for clarity


CodyCBakerPhD:
I meant that you can modify it at the extractor level within the interface; the notion of the `OpticChannel` is NWB specific and the notion of the naming scheme is also NWB specific, so IDK that ROIExtractors would need to adopt it directly

PR #225, opened 2023-05-31
Add `MiniscopeImagingExtractor`

weiglszonja:
Add extractor for the Miniscope format:

- `MiniscopeImagingExtractor` takes a `folder_path` that points to the main folder where the microscope and behavior video subfolders are located.

The expected folder structure for Miniscope V3 recordings, based on the example session received from the Tye lab, is organized as follows:

- `C6-J588_Disc5/` (main folder)
  - `15_03_28/` (subfolder corresponding to the recording time)
    - `Miniscope/` (subfolder containing the microscope video stream)
         - `0.avi` (file containing the microscope video)
         - `metaData.json` (file containing the metadata for the microscope device)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10, 3.11]
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: s-weigand/setup-conda@v1
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: s-weigand/setup-conda@v1
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, 3.10, 3.11]
+        python-version: [3.8, 3.9, "3.10", 3.11]
         os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - uses: s-weigand/setup-conda@v1
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,2 +1,3 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader==1.4.1
+neuroconv[video]
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,4 +1,4 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader==1.4.1
-neuroconv[video]@git+https://github.com/catalystneuro/neuroconv.git@main
+neuroconv[video]>=0.3.0
 natsort>=8.3.1
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,3 +1,4 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader==1.4.1
 neuroconv[video]
+natsort>=8.3.1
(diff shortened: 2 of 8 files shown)


src/roiextractors/extractors/miniscopeimagingextractor/miniscopeimagingextractor.py
```diff
@@ -0,0 +1,113 @@
+import json
+import re
+from pathlib import Path
+from typing import Optional, Tuple, List
+
+import numpy as np
+from natsort import natsorted
+
+from neuroconv.datainterfaces.behavior.video.video_utils import VideoCaptureContext
```
CodyCBakerPhD, line 9: Should be moved inside a local scope if that's the approach we're taking

requirements-full.txt
```diff
@@ -1,2 +1,3 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader==1.4.1
+neuroconv[video]
```
CodyCBakerPhD, line 3: Looks like you'll have to add `natsort` as well

CodyCBakerPhD:
Mostly looks good; is there an example file to share yet?

requirements-full.txt
weiglszonja: > Looks like you'll have to add `natsort` as well

You're right, thanks!

> Mostly looks good; is there an example file to share yet?

I've just uploaded at `tye-lab-to-nwb/C6-J588_Disc5_miniscope_stub.nwb`

src/roiextractors/extractors/miniscopeimagingextractor/miniscopeimagingextractor.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        if channel != 0:
+            raise NotImplementedError(
+                f"The {self.extractor_name}Extractor does not currently support multiple color channels."
+            )
+
+        end_frame = end_frame or self.get_num_frames()
+        start_frame = start_frame or 0
+
+        video = np.empty(shape=(end_frame - start_frame, *self.get_image_size()))
```
CodyCBakerPhD, line 108: @weiglszonja I think you might need to set `dtype="int16"` here; the file shows the data ending up in `float64`

I'd also ensure that the type returned by `cv2.cvtColor`  with `cv2.COLOR_RGB2GRAY` is `int16` not `float64` (and manually cast if so)

weiglszonja: `float64` is definitely incorrect, thank you for catching it! 

> I'd also ensure that the type returned by cv2.cvtColor with cv2.COLOR_RGB2GRAY is int16 not float64 (and manually cast if so)

The return type is actually `uint8` and not int16 which is the same as the `VideoCaptureContext.get_video_frame_dtype()` returns

CodyCBakerPhD: From meeting: yes, I suppose it does. OK then

CodyCBakerPhD:
From meeting: dropping 3.7 here

@weiglszonja Can you try adding 3.11 in a follow-up just to see if it works?

weiglszonja:
@CodyCBakerPhD Thanks for checking the file, turned out I was using an roiextractors version the commit before I fixed the[ data type issue](https://github.com/catalystneuro/roiextractors/pull/225/commits/30b3ea1fddc68c8efed10abd81d8ab28ed941e63). 
The example file has been replaced, now it is showing up as `uint8` values.

.github/workflows/testing.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
```
CodyCBakerPhD, line 14:
```suggestion
        python-version: [3.8, 3.9, 3.10, 3.11]
```

Extending testing versions

.github/workflows/testing.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10, 3.11]
```
CodyCBakerPhD, line 14:
```suggestion
        python-version: [3.8, 3.9, "3.10", 3.11]
```

CodyCBakerPhD (approved):

CodyCBakerPhD:
Awesome! Thanks for all the work on this

PR #227, opened 2023-06-03
added new args to allow for incomplete data import and fixed ypix bug

RichieHakim:
This PR contains focuses on some changes to the `Suite2pSegmentationExtractor` class. The following changes are made:
1. Now allows for incomplete data ingestion. New argument (`allow_incomplete_import`). If any files are missing (ie F.npy is not there), it can simply ignore it an move on. Also, a new argument (`warn_missing_files`) toggles printing warnings for missing files.
2. New arguments for searching within a subdirectory specified by the plane number. This is often not how directories are structured and this toggle removes the user burden of restructuring directories if they choose not to.
3. Potential bug fix in the orientation of roi_masks. It seems that the code was writen to flip the roi_masks in the y direction. I'm not sure why this is the case. I removed code that ingested this way and code that undid this during saving.
4. Typo fix in readme.

### DIFF
--- a/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
+++ b/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
@@ -1,6 +1,7 @@
 import shutil
 from pathlib import Path
 from typing import Optional
+import warnings
 
 import numpy as np
 
@@ -22,6 +23,9 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
         folder_path: Optional[PathType] = None,
         combined: bool = False,
         plane_no: IntType = 0,
+        search_plane_subdirectory: bool = True,
+        allow_incomplete_import: bool = False,
+        warn_missing_files: bool = True,
         file_path: Optional[PathType] = None,
     ):
         """
@@ -34,6 +38,14 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
             if the plane is a combined plane as in the Suite2p pipeline
         plane_no: int
             the plane for which to extract segmentation for.
+        search_plane_subdirectory: bool
+            If True, will search for files in 'folder_path/plane{plane_no}',
+             else will search for files in 'folder_path'.
+        allow_incomplete_import: bool
+            If True, will not raise an error if the file is incomplete.
+        warn_missing_files: bool
+            If True, will raise a warning if a file is incomplete and
+             allow_incomplete_import is True.
         file_path: str or Path [Deprecated]
             ~/suite2p folder location on disk
 
@@ -56,12 +68,27 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
         self.plane_no = plane_no
         self.folder_path = Path(folder_path)
 
-        self.stat = self._load_npy("stat.npy")
-        self._roi_response_raw = self._load_npy("F.npy", mmap_mode="r").T
-        self._roi_response_neuropil = self._load_npy("Fneu.npy", mmap_mode="r").T
-        self._roi_response_deconvolved = self._load_npy("spks.npy", mmap_mode="r").T
-        self.iscell = self._load_npy("iscell.npy", mmap_mode="r")
-        self.ops = self._load_npy("ops.npy").item()
+        self._search_plane_subdirectory = search_plane_subdirectory
+
+        def try_load_npy(filename, mmap_mode=None, fn_transform=lambda x: x):
+            """
+            This function allows for incomplete import of files.
+            """
+            try:
+                return fn_transform(self._load_npy(filename, mmap_mode=mmap_mode))
+            except FileNotFoundError:
+                if allow_incomplete_import:
+                    warnings.warn(f"File {filename} not found.") if warn_missing_files else None
+                    return None
+                else:
+                    raise FileNotFoundError(f"File {filename} not found.")
+                
+        self.stat = try_load_npy("stat.npy")
+        self._roi_response_raw = try_load_npy("F.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self._roi_response_neuropil = try_load_npy("Fneu.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self._roi_response_deconvolved = try_load_npy("spks.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self.iscell = try_load_npy("iscell.npy", mmap_mode="r")
+        self.ops = try_load_npy("ops.npy", fn_transform=lambda x: x.item())
 
         self._channel_names = [f"OpticalChannel{i}" for i in range(self.ops["nchannels"])]
         self._sampling_frequency = self.ops["fs"] * [2 if self.combined else 1][0]
@@ -70,7 +97,11 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
         self._image_mean = self._summary_image_read("meanImg")
 
     def _load_npy(self, filename, mmap_mode=None):
-        file_path = self.folder_path / f"plane{self.plane_no}" / filename
+        if self._search_plane_subdirectory:
+            file_path = self.folder_path / f"plane{self.plane_no}" / filename
+        else:
+            file_path = self.folder_path / filename
+        assert file_path.exists(), f"File {file_path} does not exist, but is required for this extractor."
         return np.load(file_path, mmap_mode=mmap_mode, allow_pickle=mmap_mode is None)
 
     def get_accepted_list(self):
@@ -115,7 +146,7 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
             pixel_mask.append(
                 np.vstack(
                     [
-                        self.ops["Ly"] - 1 - self.stat[i]["ypix"],
+                        self.stat[i]["ypix"],
                         self.stat[i]["xpix"],
                         self.stat[i]["lam"],
                     ]
@@ -173,7 +204,7 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
         for no, i in enumerate(stat):
             stat[no] = {
                 "med": roi_locs[no, :].tolist(),
-                "ypix": segmentation_object.get_image_size()[0] - 1 - pixel_masks[no][:, 0],
+                "ypix": pixel_masks[no][:, 0],
                 "xpix": pixel_masks[no][:, 1],
                 "lam": pixel_masks[no][:, 2],
             }
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ To update data later, `cd` into the test directory and run `gin get-content`
 
 ## Class descriptions:
 
-*  **SegmentationExtractor:** An abstract class that contains all the meta-data and output data from the ROI segmentation operation when applied to the pre-processed data. It also contains methods to read from and write to various data formats ouput from  the processing pipelines like SIMA, CaImAn, Suite2p, CNNM-E.
+*  **SegmentationExtractor:** An abstract class that contains all the meta-data and output data from the ROI segmentation operation when applied to the pre-processed data. It also contains methods to read from and write to various data formats ouput from  the processing pipelines like SIMA, CaImAn, Suite2p, CNMF-E.
 
 *  **NumpySegmentationExtractor:** Contains all data coming from a file format for which there is currently no support. To construct this, all data must be entered manually as arguments.
 


RichieHakim:
Transposing the NWB summary image may also be necessary. I can add that to this PR if you agree @CodyCBakerPhD 
Screenshot from 2023-06-18 17-29-13

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -56,12 +68,27 @@ def __init__(
...
+            This function allows for incomplete import of files.
+            """
+            try:
+                return fn_transform(self._load_npy(filename, mmap_mode=mmap_mode))
+            except FileNotFoundError:
+                if allow_incomplete_import:
+                    warnings.warn(f"File {filename} not found.") if warn_missing_files else None
+                    return None
+                else:
+                    raise FileNotFoundError(f"File {filename} not found.")
```
CodyCBakerPhD, line 84: So, if I understand your issue correctly, the problem that you encounter trying to use our reader is that your example of Suite2P output does not have all of these files present in each subfolder (by plane)?

If so, then I think it's fine to refactor this to only load the subset of available traces, and no need to error, warn, or have it be flag controllable (and just always load everything it can)

This would simplify the code here a lot; I'll suggest a simple ternary structure below for this

RichieHakim: > your example of Suite2P output does not have all of these files present in each subfolder (by plane)?
Correct. I and others often work with only stat.npy and/or ops.npy files for doing ROI tracking across sessions.

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -70,7 +97,11 @@ def __init__(
         self._image_mean = self._summary_image_read("meanImg")
 
     def _load_npy(self, filename, mmap_mode=None):
-        file_path = self.folder_path / f"plane{self.plane_no}" / filename
+        if self._search_plane_subdirectory:
+            file_path = self.folder_path / f"plane{self.plane_no}" / filename
+        else:
+            file_path = self.folder_path / filename
+        assert file_path.exists(), f"File {file_path} does not exist, but is required for this extractor."
         return np.load(file_path, mmap_mode=mmap_mode, allow_pickle=mmap_mode is None)
```
CodyCBakerPhD, line 105:
```suggestion
        file_path = self.folder_path / f"plane{self.plane_no}" / filename
        if not file_path.exists():
            return
        return np.load(file_path, mmap_mode=mmap_mode, allow_pickle=mmap_mode is None)
```

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -70,7 +97,11 @@ def __init__(
         self._image_mean = self._summary_image_read("meanImg")
 
     def _load_npy(self, filename, mmap_mode=None):
```
CodyCBakerPhD, line 99: Though with my suggested modification, not a bad idea to improve this private functions name+annotation+docstring to reflect new behavior

```suggestion
    def _attempt_load_npy(self, filename, mmap_mode=None) -> np.ndarray | None:
    """Attempt to load the .npy file at the in the current `plane_no` subfolder of this extractor; return None if file is missing."""
    
```

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
RichieHakim: they are also generally not in strict directory patterns like 'plane0' etc.

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -34,6 +38,14 @@ def __init__(
             if the plane is a combined plane as in the Suite2p pipeline
         plane_no: int
             the plane for which to extract segmentation for.
+        search_plane_subdirectory: bool
+            If True, will search for files in 'folder_path/plane{plane_no}',
+             else will search for files in 'folder_path'.
```
CodyCBakerPhD, line 43: After discussing it a bit, we don't think it's a good idea to support this level of implicit behavior as it makes the reader more complex than needed; in particular we find it simplest if reader classes take a single input that always behaves in a single explicit way

This minimizes the number of use cases, tests, and cross-argument conditions that need to be tested and supported

So within the current framework, we want to support the pattern where `Suite2P` as a format takes a (i) top-level `folder_path` that contains subfolders for each plane, and specifies which plane to load data for using the (ii) `plane_no` argument

Now, there's a lot we would like to improve on this in the long-term... namely, fully explicit control over which streams are detected (via a class method) and loaded (via user selection of streams from that detection) so that this kind of control is even more explicit for users; but until we get the bandwidth to do a wider refactor of that over the repo, we'll have to stay within the current paradigm

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
CodyCBakerPhD: > they are also generally not in strict directory patterns like 'plane0' etc.


Well, this is one part that is fairly important since we interpret all of this as being enforceable as a part of Suite2P format - are you saying that as you run Suite2P, it only saves the individual files and does not actually save a sessions worth of data using the `plane{integer}` subfolder format?

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -75,10 +76,16 @@ def _trace_extractor_read(self, field):
...
-    def _summary_image_read(self):
+    def _correlation_image_read(self):
         if self._dataset_file["estimates"].get("Cn"):
             return np.array(self._dataset_file["estimates"]["Cn"])
 
+    def _summary_image_read(self):
+        if self._dataset_file["estimates"].get("b"):
+            FOV_shape = self._dataset_file["params"]["data"]["dims"][()]
+            b_sum = self._dataset_file["estimates"]["b"][:].sum(axis=1)
+            return np.array(b_sum).reshape(FOV_shape, order="F")
```
CodyCBakerPhD, line 87: Are you able to confirm why the column order is Fortran-like? Is that documented somewhere?

CodyCBakerPhD:
@RichieHakim Sorry I haven't had the bandwidth to get to this until now - I'm personally spread across a number of projects and this corner of the ecosystem has often not gotten as much attention (unless we need to add a new format, such as the recent additions of Miniscope and Bruker)

To alleviate this, we'll be bringing in @pauladkisson who will become the key point person for reviewing, dev ops, and other miscellaneous admin responsibilities; I'll be bringing them up to speed on this repository in the coming weeks

In the meantime, I have a couple concerns about this PR being more complicated than it needs to be

We certainly want the `SegmentationExtractor` for Suite2P to not fail on some target output, and given that you indicate it is possible to have an absence of certain files in your output, I think the general request to enable safe loading of individual streams (potential trace files) is perfectly reasonable

We generally try to avoid `try/except` patterns as a style whenever possible however, and I'd say it is perfectly acceptable for the time being to simply skip the loading of the numpy file conditional on whether or not the file exists on the system; this reduces the code changes here by quite a bit while maintaining consistency of trace dictionary behavior in the current paradigm (though the very notion of the trace dictionary is something we'd eventually like to refactor out)

> 3. Potential bug fix in the orientation of roi_masks. It seems that the code was writen to flip the roi_masks in the y direction. 

What I can say in general is that  ROIExtractors follows image (height x width) shape convention similar to OpenCV and other image parsing libraries

Unfortunately, this often conflicts with the NWB equivalent, whose schema mandates usage of the `(width x height)` data arrays instead; hence, why the NWB extractor for any of these fields will also transpose the data from the source

However, given how long ago this class was implemented, and how the individual who originally implemented it is no longer active, I can't actually judge if it is correct or not

Now, if you're saying the flipping being done at https://github.com/catalystneuro/roiextractors/pull/227/files#diff-cdefaf6b12d83c6942b1dcdc3245b73ef75445fac2293e7ca046be26a77d720fL118 and https://github.com/catalystneuro/roiextractors/pull/227/files#diff-cdefaf6b12d83c6942b1dcdc3245b73ef75445fac2293e7ca046be26a77d720fL176-R207 is a bug as far as you can tell, can you do me a favor and raise a separate issue for that and remove the changes from this PR (so this PR can focus purely on the partial loading of streams +/- the Caiman summary improvement?); then we can do a deep dive using some source images as reference to confirm if it's intentional or not



> Typo fix in readme.

Thanks, if you're open to making the changes as I've suggested, can you also update the [CHANGELOG.md file](https://github.com/catalystneuro/roiextractors/blob/main/CHANGELOG.md) describing the changes of this PR?

> Add mean image extraction from caiman segmentation class.

Back to the question of testing; do you know if the example files on the GIN repo have these new fields that you're adding a functionality for? And if not can you share a version of your files (or a stubbed one that is < 10 MB) that you developed it for?

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
CodyCBakerPhD: If accepted, this would allow you to pretty much revert all the changes in the `__init__` that used the locally scoped function there (including all the new arguments)

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
RichieHakim: > are you saying that as you run Suite2P, it only saves the individual files and does not actually save a sessions worth of data using the plane{integer} subfolder format?

No. However, it is not uncommon for these directories to have their names changed or to copy only data that are necessary for subsequent steps. Given that stat.npy and ops.npy are very small compared to F.npy and other files, they are often moved without the others.

I'd say that this is not a huge deal and I think having strong opinions on directory structures is probably wise. Though, allowing incomplete file ingestion is pretty critical still.

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
RichieHakim: lgtm

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
RichieHakim: Sounds good

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
RichieHakim: trial and error with the dataset you provided plus the following:

https://github.com/flatironinstitute/CaImAn/blob/80e1681bbce8fb4a36b04c57d9e42ffb75a32d58/caiman/base/rois.py#L461
https://github.com/flatironinstitute/CaImAn/blob/80e1681bbce8fb4a36b04c57d9e42ffb75a32d58/caiman/utils/utils.py#L267
https://github.com/flatironinstitute/CaImAn/blob/80e1681bbce8fb4a36b04c57d9e42ffb75a32d58/caiman/utils/visualization.py#L106

RichieHakim:
@CodyCBakerPhD 

Thanks Cody! I appreciate that you all are maintaining a pretty large amount of code and projects so I'm thankful that you were able to address this PR. Your comments were good.

There are a few changes in here so I'll go one by one

Suite2p class:
- supporting incomplete ingestion: This is critical and it looks like your changes support this
- Non-strict directory structure for plane numbers: This is not critical and I'm cool with leaving it as is.

CaImAn class:
- Fortran ordering of reshaping: I'm 95% sure this is correct but I'd suggest some further testing with new datasets.

NWB class:
- transposition of image: I can't really conclude on the code links you sent. I'm not sure. This is not data I handle myself. I just plotted the results and it clearly showed that the image was transposed.

Changelog: Will do later today.

Data files:
> Back to the question of testing; do you know if the example files on the GIN repo have these new fields that you're adding a functionality for? And if not can you share a version of your files (or a stubbed one that is < 10 MB) that you developed it for?

I don't have access to personally extracted caiman files. The example caiman files on the GIN repo do not have a mean image. I think a recurring theme here is that we are unable to test changes effectively because we don't trust the GIN repo data. It might be a good idea to invest a little bit of time and rerun some example data through the different extraction pipelines (and maybe make a simple pipeline to run new tests as updates to those repos occur). 

I think it might be useful to provide a little more context to how I (and others) have been using this repo. 
Here is the ingestion class I'm wrapping roiextractors classes with: https://github.com/RichieHakim/ROICaT/blob/dev/roicat/data_importing.py#L1429
and here is a demo on how it is being used: https://github.com/RichieHakim/ROICaT/blob/main/notebooks/jupyter/other/demo_data_importing.ipynb

Let me know if there is anything else I can help with. Thanks again!

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
CodyCBakerPhD: OK thanks for thoroughly checking this!

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
CodyCBakerPhD: > I'd say that this is not a huge deal and I think having strong opinions on directory structures is probably wise. Though, allowing incomplete file ingestion is pretty critical still.

👍  OK sounds good, thanks for clarifying!

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -56,21 +64,38 @@ def __init__(
...
         self._image_mean = self._summary_image_read("meanImg")
 
-    def _load_npy(self, filename, mmap_mode=None):
+    def _attempt_load_npy(self, filename, mmap_mode=None) -> np.ndarray | None:
+        """Attempt to load the filename located in the current `plane_no` subfolder; return None if file is missing."""
+
         file_path = self.folder_path / f"plane{self.plane_no}" / filename
+        if not file_path.exists():
+            return
         return np.load(file_path, mmap_mode=mmap_mode, allow_pickle=mmap_mode is None)
```
CodyCBakerPhD, line 99:
```suggestion
    def _attempt_load_npy(self, filename, mmap_mode=None, transpose: bool = False) -> np.ndarray | None:
        """Attempt to load the filename located in the current `plane_no` subfolder; return None if file is missing."""

        file_path = self.folder_path / f"plane{self.plane_no}" / filename
        if not file_path.exists():
            return
        array = np.load(file_path, mmap_mode=mmap_mode, allow_pickle=mmap_mode is None)
        if transpose:
            return array.T
        return array
```

CodyCBakerPhD: small improvement for above code reduction

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -56,21 +64,38 @@ def __init__(
...
+                    return None
+                else:
+                    raise FileNotFoundError(f"File {filename} not found.")
+
+        self.stat = try_load_npy("stat.npy")
+        self._roi_response_raw = try_load_npy("F.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self._roi_response_neuropil = try_load_npy("Fneu.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self._roi_response_deconvolved = try_load_npy("spks.npy", mmap_mode="r", fn_transform=lambda x: x.T)
+        self.iscell = try_load_npy("iscell.npy", mmap_mode="r")
+        self.ops = try_load_npy("ops.npy", fn_transform=lambda x: x.item())
```
CodyCBakerPhD, line 85:
```suggestion
        self.stat = _attempt_load_npy(filename="stat.npy", mmap_mode="r")
        self._roi_response_raw = _attempt_load_npy(filename="F.npy", mmap_mode="r", transpose=True)
        self._roi_response_neuropil = _attempt_load_npy(filename="Fneu.npy", mmap_mode="r", transpose=True)
        self._roi_response_deconvolved = _attempt_load_npy(filename="spks.npy", mmap_mode="r", transpose=True)
        self.iscell = _attempt_load_npy(filename="iscell.npy", mmap_mode="r")
        self.ops = _attempt_load_npy(filename="ops.npy", mmap_mode="r")
        self.ops = self.ops.item() if self.ops is not None else self.ops
```

After the first comment regarding `attempt_load_npy` was accepted (and with extra simplification I just added), it allows us to remove a lot of this

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -34,6 +37,11 @@ def __init__(
             if the plane is a combined plane as in the Suite2p pipeline
         plane_no: int
             the plane for which to extract segmentation for.
+        allow_incomplete_import: bool
+            If True, will not raise an error if the file is incomplete.
+        warn_missing_files: bool
+            If True, will raise a warning if a file is incomplete and
+             allow_incomplete_import is True.
```
CodyCBakerPhD, line 44:
```suggestion
```

With cleanup below, optional warning no longer needed - incomplete import always enabled

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -22,6 +23,8 @@ def __init__(
         folder_path: Optional[PathType] = None,
         combined: bool = False,
         plane_no: IntType = 0,
+        allow_incomplete_import: bool = False,
+        warn_missing_files: bool = True,
```
CodyCBakerPhD, line 27:
```suggestion
```

With cleanup below, optional warning no longer needed - incomplete import always enabled

src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
```diff
@@ -1,6 +1,7 @@
 import shutil
 from pathlib import Path
 from typing import Optional
+import warnings
```
CodyCBakerPhD, line 4:
```suggestion
```

No need for warning either if the rest is cleaned up

CodyCBakerPhD:
> transposition of image: I can't really conclude on the code links you sent. I'm not sure. This is not data I handle myself. I just plotted the results and it clearly showed that the image was transposed.

I see; thanks for the report - given this PR already has a couple other things going for it, I say let's do a follow-up that specifically checks it for consistency across our repo

As you say, generating some up-to-date examples from each pipeline using some standard base data could be very helpful in this endeavor so I'll task @pauladkisson (who will also be improving the documentation for our project) with handling this in the coming weeks

RichieHakim:
What is the status of this PR? Are we waiting on anything else before merging? Happy to help if you point me in the right direction.

CodyCBakerPhD:
> What is the status of this PR? 

Thanks for checking in - just waiting on the cleanup of the unneeded lines pointed out in my suggestions above

CodyCBakerPhD:
@weiglszonja Can you confirm that most of the approved features from this request were merged in with #234? If so, this can likely be closed

weiglszonja:
These are addressed in #242
- missing traces: 
https://github.com/catalystneuro/roiextractors/blob/138a8e345ab5b73fc5a350951c8b2c4759f82ccb/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py#L202-L203
- roi_masks orientation:
https://github.com/catalystneuro/roiextractors/blob/138a8e345ab5b73fc5a350951c8b2c4759f82ccb/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py#L254

Not added in #242:
- readme fix (CNNM-E)
- Add mean image extraction from caiman segmentation class.

CodyCBakerPhD:
@weiglszonja Thanks for summarizing

> Not added in https://github.com/catalystneuro/roiextractors/pull/242:
> 
> readme fix (CNNM-E)
> Add mean image extraction from caiman segmentation class.


Looks like the README thing is no longer a problem (fixed at some point on `main`)

@alessandratrapani Can you take a crack at implementing isolated fixes to the Caiman images as seen from this PR?

pauladkisson (approved):

src/roiextractors/extractors/caiman/caimansegmentationextractor.py
h-mayorquin: Hey guys, I am thinking on reverting this:

https://github.com/catalystneuro/roiextractors/issues/456

I could not find a discussion on why `b` as assigned as a image_mean so it would be good to get your feedback if you remember

PR #230, opened 2023-06-30
Update `BrukerTiffImagingExtractor`

weiglszonja:
Update `BrukerTiffImagingExtractor` to handle the additional types of Bruker data (volumetric, multi color, stimulation)
WIP

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,0 +1,75 @@
+name: Full Tests
+on:
+  schedule:
+    - cron: "0 0 * * *"  # daily
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  run:
+    name: Full tests on ${{ matrix.os }} with Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [3.8, 3.9, "3.10", 3.11]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+      - uses: s-weigand/setup-conda@v1
+      - uses: actions/checkout@v2
+      - run: git fetch --prune --unshallow --tags
+      - name: Setup Python ${{ matrix.python-version }}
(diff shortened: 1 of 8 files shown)


CodyCBakerPhD:
Thanks @weiglszonja 

Going to enlist the help of @pauladkisson to help review this PR to get them some more experience with the workflow of managing this repository

pauladkisson (commented):
Changes look pretty reasonable to me.  It looks like CI checks are failing due to file/path issues, could you clear that up? Or if you're using a file that isn't in the ophys testing repo, we should probably add it.

weiglszonja:
Thanks @pauladkisson, the PR for uploading the test data is [here](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/14), once it's merged hopefully the tests will pass.
I'd still prefer to keep this in draft until I receive the multichannel example.

weiglszonja:
**Notes for Bruker version** (can be checked in the XML file)
- files that were created with version `5.6.64` are individual .ome.tif files (dual plane example, single color example)
- files that were created with version `5.8.64` are multipage .ome.tif files (dual color example [here](https://drive.google.com/drive/u/0/folders/1RrTfkEKYHC1WjxVub6VQchGmh9jXxxen))
The number of files for the multipage tiffs can depend on the number of channels, but I also think there can be more files per channel depending on the chunk size that was used (e.g. ~500MB). Also possible that even with the new version the user can choose to create individual tiffs, but that is being confirmed [here](https://github.com/Tyelab/bruker_pipeline/issues/9#issuecomment-1628790257).

CodyCBakerPhD:
Unfortunately it looks like @h-mayorquin's data, though it was originally Bruker output, was converted into another format (nifty) after acquisition (and source was not kept) so the example data we have here is from Pinto lab and the NWB helpdesk (best volumetric example so far)

CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM! Thanks for all the hard work on this

CodyCBakerPhD (approved):

PR #235, opened 2023-08-31
Streamline Docstrings

pauladkisson:
# Significant Changes
- Adds pydocstyle to pre-commit to ensure consistent docstring formatting.
- Adds docstrings to all members (modules, classes, methods, functions) without them and makes sure all docstrings are compliant with pydocstyle
- Some minor typo fixes and clarifications here and there


src/roiextractors/example_datasets/toy_example.py
```diff
@@ -7,10 +14,39 @@
...
+    size_x: int
+        Size of x dimension (pixels)
+    size_y: int
+        Size of y dimension (pixels)
+    roi_size: int
+        Siz of ROI in x and y dimension (pixels)
+    min_dist: int
+        Minimum distance between ROI centers (pixels)
+    mode: str
+        'uniform' or 'gaussian'.
```
CodyCBakerPhD, line 37: Stuff like this can - and in the Python 3.8 world - should become `typing.Literal`, e.g., `Literal["uniform", "gaussian"]`

I'm fine with this PR being the 'write things down as they are not as they should be', but that provides content for a next-step

src/roiextractors/example_datasets/toy_example.py
```diff
@@ -7,10 +14,39 @@
...
+    ----------
+    num_units: int
+        Number of ROIs
+    size_x: int
+        Size of x dimension (pixels)
+    size_y: int
+        Size of y dimension (pixels)
+    roi_size: int
+        Siz of ROI in x and y dimension (pixels)
+    min_dist: int
```
CodyCBakerPhD, line 36: I know @bendichter has a strong preference towards the `numpydoc` style guide for describing default values (and/or optionality) like so https://github.com/catalystneuro/neuroconv/pull/260#discussion_r1055557321

Reference: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

I'll grant you, the ongoing PEP 727 discussions (courtesy @h-mayorquin for point it out) emphasize these are pseudo-standards at best, which makes linting/parsing very difficult: https://discuss.python.org/t/pep-727-documentation-metadata-in-typing/32566/14

So similarly to my other comment, I'm fine with this being a 'first step' but we should reach agreement beyond pydocstyle about these other things too

src/roiextractors/example_datasets/toy_example.py
pauladkisson: Agree. Added a TODO so we don't forget.

src/roiextractors/example_datasets/toy_example.py
CodyCBakerPhD: I think Ben's threshold, for which another set of Python tests could be added similar to your inheritance tests, is whether or not the numpy docstring parser library can successfully scrape the docstring to essentially reproduce the method signature

src/roiextractors/example_datasets/toy_example.py
```diff
@@ -87,7 +124,7 @@ def toy_example(
     size_y: int
         Size of y dimension (pixels)
     roi_size: int
-        Siz of ROI in x and y dimension (pixels)
+        Size of ROI in x and y dimension (pixels)
```
CodyCBakerPhD, line 127: @pauladkisson Could you add Codespell to the CI/pre-commit in a follow-up PR?

Example on NeuroConv: https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/codespell.yml

src/roiextractors/extraction_tools.py
```diff
@@ -461,6 +591,20 @@ def write_to_h5_dataset_format(
...
+    Parameters
+    ----------
+    imaging: ImagingExtractor
+        The imaging extractor object to be saved in the .h5 file
+    ax: matplotlib axis
+        Axis to plot the video. If None, a new axis is created.
+
+    Returns
+    -------
+    anim: matplotlib.animation.FuncAnimation
```
CodyCBakerPhD, line 605: I find it interesting (again, pseudo-standard stuff) that the name of the return variable is a part of the docstring

Not against it, since it can be a 'suggestion' used to give downstream implementations a more standard naming convention

But it also syntactically doesn't matter what name the variable was at time of return; only the expected type matters

src/roiextractors/extraction_tools.py
```diff
@@ -487,9 +631,25 @@ def animate_func(i, imaging, im, ax):
...
+
+    Parameters
+    ----------
+    dict: dict
+        Dictionary to check.
+
+    Returns
+    -------
+    dict: dict
+        Dictionary with mat-objects converted to nested dictionaries.
```
CodyCBakerPhD, line 647: I notice with these a slight inconsistency; sometimes a one-line description of a field ends in a period (forms a complete sentence) but other times it does not, such as https://github.com/catalystneuro/roiextractors/pull/235/files#diff-9db21a3b993bb8f4a57b59186e0b23167702413272b3f7aa302a7bf5d0414d8bR46

src/roiextractors/example_datasets/toy_example.py
CodyCBakerPhD: Though I see for the other docstrings you did get into the habit of writing it in that way - maybe it's just these earlier once that didn't follow that convention?

src/roiextractors/example_datasets/toy_example.py
pauladkisson: See #236

src/roiextractors/extraction_tools.py
pauladkisson: Yeah, technically, it doesn't matter in the code (in the sense that you can return a named variable or some expression), but it is just easier to describe/document the return when it has a name.  I also prefer named returns in general (with the exception of very short functions) because they are more self-commenting than return expressions (although I couldn't find any authoritative expression of this preference ex. StackOverflow).

src/roiextractors/extraction_tools.py
CodyCBakerPhD: Similar to https://github.com/catalystneuro/roiextractors/pull/235#discussion_r1317709879, this seems to mostly be an inconsistency with some of the earlier docstrings, the more recent all end in periods

src/roiextractors/example_datasets/toy_example.py
CodyCBakerPhD: Though a part of Ben's opinion on this kind of thing also comes down to the spacing of the colon, such as

```python
     roi_size : int, optional
           # ^ ^
        Size of ROI in x and y dimension (pixels).
```

src/roiextractors/extraction_tools.py
CodyCBakerPhD: > but it is just easier to describe/document the return when it has a name.

I would agree with you if and only if the name of the return variable is nicely readable

In this case, a return name of `anim` is not verbose enough; on first glance I thought it was short for `animal`, and only upon reading the annotation type was it clear to be short for `animation`

Maybe another follow-up could be to choose better variable names for such things?

src/roiextractors/extraction_tools.py
pauladkisson: Looks like all [numpydoc examples](https://numpydoc.readthedocs.io/en/latest/format.html) of parameter/return descriptions end in periods.  But pydocstyle doesn't check this (unlike the first line which it checks for a period).

CodyCBakerPhD: From the conversation on https://github.com/catalystneuro/roiextractors/pull/235#discussion_r1317784019, is it useful at all in this case to give the return variable the same name as its built-in type?

pauladkisson: I wasn't too careful about it in this PR tbh.

pauladkisson: Of course not.  But that's a problem with the code -- not the docstring.  It's questionable to overwrite the name `dict` at all since it's a built-in, never mind the fact that it is not descriptive at all really.

pauladkisson: This sort of stuff could be part of a big refactor for clarity/readability, but I would want some clear actionable objectives since readability can be somewhat subjective.

CodyCBakerPhD: >  It's questionable to overwrite the name dict at all since it's a built-in

A good point


> This sort of stuff could be part of a big refactor for clarity/readability, but I would want some specific actionable objectives since readability can be somewhat subjective.

Can be many small refactors over time, starting with things like not overriding built-in names, and using full words to describe an object

src/roiextractors/extraction_tools.py
pauladkisson: See #238.

src/roiextractors/extraction_tools.py
pauladkisson: See #238.

src/roiextractors/example_datasets/toy_example.py
CodyCBakerPhD: As per conversation: will look into other linters that are more specific than `pydocstyle`

CodyCBakerPhD:
So, this looks good to me now as-is

@pauladkisson Would you like to merge this in directly (and follow-up PR perhaps enhances them to be even more stringent using a better linter) or wait and find a better linter in this PR?

pauladkisson:
We should merge this directly and follow-up.  I still think the [numpydoc validation pre-commit](https://numpydoc.readthedocs.io/en/latest/validation.html) is going to be our best bet, but getting it to work properly has proved more complicated than I originally thought.  Plus, this PR is already pretty substantial.

CodyCBakerPhD (approved):

PR #241, opened 2023-09-14
Multi-Plane and Multi-Channel Support for ScanImage

### LINKED ISSUE #240: [Feature]: Multi-channel and Multi-plane support for ScanImage
### What would you like to see added to ROI Extractors?

ScanImage currently only supports single-plane and single-channel .tif files from ScanImage.  But mouseV1 data contains both multi-plane and multi-channel data.  So, ROIExtractors should support this data.

Yes.

### Code of Conduct

pauladkisson:
Fixes #240.


h-mayorquin:
So for Bruker we have an extractor for the single channel - single plane, then another for single plane and then one that gets everything. But here, you are changing the single extractor that we have and adding al the functionality, right?

Is this because the tiff files sometimes come with only (time, rows columns) but other times come with everything included (time, row, columns, depth, channel)? Looking at your code it seems that this still only reads one file so probably yes?

pauladkisson:
> So for Bruker we have an extractor for the single channel - single plane, then another for single plane and then one that gets everything. But here, you are changing the single extractor that we have and adding al the functionality, right?

Yes, that is how it's currently implemented.  But, I am still playing around with the structure to see how it can best represent the data (still in draft).  I am still planning on incorporating the concept of streams, but also toying around with the idea of a base `MultiPlaneImagingExtractor` to unify the API and define some volumetric-specific fields. 


> Is this because the tiff files sometimes come with only (time, rows columns) but other times come with everything included (time, row, columns, depth, channel)? Looking at your code it seems that this still only reads one file so probably yes?

The .tiff files are always structured (time, rows, columns) BUT if they contain multi-plane and/or multi-channel data, they cycle round-robin in the 'time' dimension ex.
```
[channel_1_plane_1_frame_1, channel_2_plane_1_frame_1, ..., channel_C_plane_1_frame_1,
channel_1_plane_2_frame_1, ..., channel_C_plane_2_frame_1, ..., channel_C_plane_P_frame_1,
channel_1_plane_1_frame_2, ..., channel_C_plane_P_frame_F]
```
Whether or not they contain multiple planes or multiple channels is fully specified by the file metadata, so I figured I could read everything with one class.

h-mayorquin:
Thanks for your explanation, @pauladkisson.

pauladkisson:
@CodyCBakerPhD, This is ready for some feedback.

I plan on moving the MultiPlaneImagingExtractor to a different file and a different PR, but I included it to give a sense of how I see the ScanImageExtractors working with the inheritance structure. 

I also played around with incorporating support for the old `ScanImageTiffImagingExtractor`, but the metadata for the file we have to test is completely different from the current metadata (the testing file is pretty old). So, I think the cleanest way to deal with it is just keep it around as a legacy version and develop new tests for the new ScanImage stuff.

Last but not least, I still need to set up a testing structure, but I figured it would be good to get some feedback first.

CodyCBakerPhD:
First pass looks good!

I'll dig in deeper over next few days while you add tests

> I also played around with incorporating support for the old ScanImageTiffImagingExtractor, but the metadata for the file we have to test is completely different from the current metadata (the testing file is pretty old). So, I think the cleanest way to deal with it is just keep it around as a legacy version and develop new tests for the new ScanImage stuff.

Sounds good for retiring the class itself - can you add a `DeprecationWarning` then with a comment `#TODO: remove on or after <three months from now>` for a smooth cycle on that?

That said, supporting multiple versions of metadata is actually something we should generally strive to support for any example files we have on the testing repo - you never know what version people are using on older (and still valuable) data or rigs that just haven't been updated. Perhaps a routing function to multiple metadata extraction methods could lessen the complexity of that step? Granted, ROIExtractors doesn't concern itself too much with metadata, more of a problem for NeuroConv side

pauladkisson:
> Sounds good for retiring the class itself - can you add a `DeprecationWarning` then with a comment `#TODO: remove on or after <three months from now>` for a smooth cycle on that?

Done!

> That said, supporting multiple versions of metadata is actually something we should generally strive to support for any example files we have on the testing repo - you never know what version people are using on older (and still valuable) data or rigs that just haven't been updated. Perhaps a routing function to multiple metadata extraction methods could lessen the complexity of that step?

I also added a legacy metadata parsing function for the old metadata but it is less comprehensive and isn't actually used by the legacy version of `ScanImageTiffImagingExtractor`

pauladkisson:
Tests are failing until the new scanimage test file gets added to the gin repo -- See [Issue # 16](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/issues/16)

CodyCBakerPhD:
> I also added a legacy metadata parsing function for the old metadata but it is less comprehensive and isn't actually used by the legacy version of ScanImageTiffImagingExtractor

Right, I suppose that would be more of a thing for the NeuroConv side

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
     ).ScanImageTiffReader
 
 
-class ScanImageTiffImagingExtractor(ImagingExtractor):
+def extract_extra_metadata(
+    file_path,
+) -> dict:  # TODO: Refactor neuroconv to reference this implementation to avoid duplication
+    """Extract metadata from a ScanImage TIFF file.
```
CodyCBakerPhD, line 30: I would suggest a method name and docstring addition specifying what versions this is known to support for reading, similar to the v3_8 below

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
...
+            for x in metadata_string.replace("\n", "\r").split("\r")
+            if "=" in x
+        }
+        extra_metadata = dict(**extra_metadata, **metadata_dict)
+    return extra_metadata
+
+
+def parse_metadata(metadata):
+    """Parse metadata dictionary to extract relevant information and store it standard keys for ImagingExtractors.
+
```
CodyCBakerPhD, line 57: I would suggest a method name and docstring addition specifying what versions this is known to support for mapping onto ROIExtractor values, similar to the v3_8 below

Always possible that a future version, or user-requested past version has different variable names

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
     ).ScanImageTiffReader
 
 
-class ScanImageTiffImagingExtractor(ImagingExtractor):
+def extract_extra_metadata(
```
CodyCBakerPhD, line 27: It's a personal opinion, but I'm usually of the view that for code suites like this, it is cleaner to move these types of helper functions (particularly for metadata parsing) to an adjacent utils file such as `scanimage_utils.py` to free up visual space in the main file for the primary extractors

What do you think about that approach?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
...
+            list of imaging extractor objects
+        """
+        super().__init__()
+        assert isinstance(imaging_extractors, list), "Enter a list of ImagingExtractor objects as argument"
+        assert all(isinstance(imaging_extractor, ImagingExtractor) for imaging_extractor in imaging_extractors)
+        self._check_consistency_between_imaging_extractors(imaging_extractors)
+        self._imaging_extractors = imaging_extractors
+        self._num_planes = len(imaging_extractors)
+
+    def _check_consistency_between_imaging_extractors(self, imaging_extractors: List[ImagingExtractor]):
```
CodyCBakerPhD, line 148: Just something to be aware of, from prior experience...

While this check is very valuable from a perspective of safety of operation support on our end, for a very large number of relatively small files (may or may not be the case depending on the buffering size here) this check can be an extremely inefficient slowdown since it has to open/read a bit of info for every single I/O

In some previous conversions I built in an override via an boolean argument to disable the check for speedup, basically saying 'I trust the data to be consistent'

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
...
+    num_planes = int(metadata["state.acq.numberOfZSlices"])
+    metadata_parsed = dict(
+        sampling_frequency=sampling_frequency,
+        num_channels=num_channels,
+        num_planes=num_planes,
+    )
+    return metadata_parsed
+
+
+class MultiPlaneImagingExtractor(ImagingExtractor):
```
CodyCBakerPhD, line 126: I would say move this to the outer level next to the base classes

CodyCBakerPhD: And maybe add a TODO to rename the previous `MultiFrameImagingExtractor`?

CodyCBakerPhD: To help disambiguate

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -22,13 +24,487 @@ def _get_scanimage_reader() -> type:
...
 
     extractor_name = "ScanImageTiffImaging"
     is_writable = True
     mode = "file"
 
+    def __init__(
+        self,
+        file_path: PathType,
+        channel: Optional[int] = 0,
+        plane: Optional[int] = 0,
```
CodyCBakerPhD, line 309: What is supposed to happen when `channel` or `plane` is `None`? It looks like it would cause an error at the inequality assertions below

CodyCBakerPhD: Also as per discussion in #245, I'm in the group that thinks these ought to be strings to be as flexible and universal as possible

I know for certain formats, the `channel` name is entirely determined by the user through the acquisition interface and it doesn't need any ordinality or even a number in it at all (especially common to see them called simply by the approximate color they filter, e.g., 'blue_channel', 'green_channel', etc.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: Sure! That seems reasonable.  Maybe we could put them in the `extraction_tools` module and rename them to be specific to scanimage i.e. `extract_scanimage_metadata` and `parse_scanimage_metadata`?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: I don't understand why this would be slow.  All of the getters just return pre-stored attributes ex. 
```
def get_sampling_frequency(self) -> float:
        return self._sampling_frequency
```
So there shouldn't be any I/O during the consistency check -- all of it should happen during the construction of `SinglePlaneImagingExtractors`.  Maybe the other IE's you were working with were designed differently?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: > I plan on moving the MultiPlaneImagingExtractor to a different file and a different PR, but I included it to give a sense of how I see the ScanImageExtractors working with the inheritance structure.

Yes.

> And maybe add a TODO to rename the previous MultiFrameImagingExtractor?

Yeah I was thinking about that, but didn't want to break old code that used `MultiImagingExtractor`... maybe we can deprecate it and remove in 3mo's like `ScanImageTiffImagingExtractor`.  Also, MultiFrame isn't quite right bc single-file IE's can have multiple frames but not need `MultiImagingExtractor`.  Maybe MultiFile or MultiChunk?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: > What is supposed to happen when channel or plane is None? It looks like it would cause an error at the inequality assertions below

`plane` and `channel` default to `0` -- they should never be `None`.

> Also as per discussion in https://github.com/catalystneuro/roiextractors/issues/245, I'm in the group that thinks these ought to be strings to be as flexible and universal as possible
> 
> I know for certain formats, the channel name is entirely determined by the user through the acquisition interface and it doesn't need any ordinality or even a number in it at all (especially common to see them called simply by the approximate color they filter, e.g., 'blue_channel', 'green_channel', etc.

Ok, then we can specify `channel_name` instead of `channel` (index).  I still think `plane` should be an index though, since for large numbers of depth planes there won't be a unique name for each plane.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: I would still keep them scoped locally to the `extractors` subfolder (though I'd also advocate for this to be in a further subfolder `scanimage` distinct from generic `tiff` reader, but that's maybe beyond this PR) to make it obvious they are not to be used more generally

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: It's not MultiFile either, you might want to concatenate reshaped contents from a single file. Chunk is an overloaded term everywhere anyway so I'd avoid using that anywhere

MultiTimeImagingExtractor? TimeConcatenatedImagingExtractor?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Sorry for not being specific ; the skip I built in was to the `__init__` of the extractors so that they did not fetch/store the attribute to being with. But they were mostly doing that already in order to pass the consistency check. So the skip override does both, (i) does not fetch the attributes at time of `__init__` (only on call to retrieve each specific attribute) and (ii) skip the consistency check that would have run on those

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: > since for large numbers of depth planes there won't be a unique name for each plane.

Why would there not be a unique name? And why would it not in principle be user specifiable on any given acquisition system? The only reason I'd see to keep it a numeric int would be if you wanted to assert something about the ordinality, but it also ought to be possible to specify an acquisition order that is not contiguous over a Z direction

Maybe @alessandratrapani can shed light on this; for something like Bruker, can a user design/specify a cycle or alternation of cycles that are not strictly increasing/decreasing with respect the Z plane?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: > I would still keep them scoped locally to the extractors subfolder

Ok!

> I'd also advocate for this to be in a further subfolder scanimage distinct from generic tiff reader

What do you think about flattening the tiff extractors in general?  They don't actually share any common utils, so maybe we should just have separate folders for scanimage, bruker, tiff, and micromanager?

CodyCBakerPhD: What do you think about flattening the tiff extractors in general?

Sounds great


> They don't actually share any common utils, so maybe we should just have separate folders for scanimage, bruker, tiff, and micromanager?

Exactly, that's how we treat them in NeuroConv anyway, IDK why it never happened here

BTW this can be a follow-up PR

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: I feel like this is an architectural change/option that we would want to propagate to all the IE's, maybe accompanied by changing the private attributes to true properties.

pauladkisson: Riffing a bit: We could add an option to the base `ImagingExtractor` class for `precomputed_properties=True` and then if you pre-compute the properties, I/O happens during `__init__` and they get stored as attributes.  But if you choose not to (like in the case you're suggesting) then I/O happens lazily and you can skip the consistency checks for MultiImagingExtractors.

pauladkisson: Regardless, I think it's outside of the scope of this PR.

CodyCBakerPhD: > I see bc if you have a multi-file setup, you only need the first IE to actually load the properties, so all of the other I/O calls are just for checking consistency -- right?

Exactly - which is a good thing in general when data volumes are small enough to not matter much for performance at runtime. There's probably a threshold of number of files vs. size of each file + time taken to parse metadata in each


> I feel like this is an architectural change/option that we would want to propagate to all the IE's, maybe accompanied by changing the private attributes to true properties.
> Regardless, I think it's outside of the scope of this PR.

Yes, I just wanted you to be aware of it in general

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: I think that Paul claim here is really reasonable and your demands too strong. You are a kind of asking him prove a negative (are YOU really sure that no system could decide to use names for their frames?). Can you elaborate on where are you coming from? I really have a hard time imaging how it would make sense to put names to planes.

Even if the numbers are not ordered (cycles mostly are though) numbers are still seem like the best way to refer to planes. You really need to have meanigful names for strings to make sense and if that is the case we should use the concept of stream to refer to it.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: Ok, after thinking a bit about this, I think the best approach is to rename the `MultiPlaneImagingExtractor` to `VolumetricImagingExtractor` or something like that to avoid making the comparison with `MultiImagingExtractor` since they're not quite parallel.  Then we can leave `MultiImagingExtractor` as is or rename it `MultiFrameImagingExtractor` to make it extra clear that we are concatenating along the frames axis.

h-mayorquin: I dislike the profileration of specific classess to cut through channels and planes but `VolumetricImagingExtractor` is a great name. 

I think that a useful heuristic is that classes names should reflect to users what they are for and not implementation details. In this case, `MultiImagingExtractor` seems to be kind of doing the latter as we are leaking to the user that we are building this extractor out of multiple ones but whey they should care about this?

If give you my format, you give me an object that allows me to extract my data ergonomically and efficiently with a common API. I don't need to know **how** you do it.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: > Why would there not be a unique name?

I just mean that the plane wouldn't be an inherently *named* quantity in the way that channels are ('red channel', 'green channel', etc.).  But the planes are an inherently *ordered* quantity (plane 0, plane 1, ..., plane N from the bottom/top).

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: `VolumetricImagingExtractor` sounds great 👍

I'm fine with either `MultiImagingExtractor` or `MultiFrameImagingExtractor` in conjunction with that

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: > I really have a hard time imaging how it would make sense to put names to planes.

`_planeA_channelRed`
`_planeB_channelBlue`

`_cycle_{x}coord_{y}coord_{z}coord`

`_{uuid()}_chan0` (hash for each plane); e.g., `_a937315b-2c1b-4cd7-abfb-d5efc562deac_chan0`

I'm saying the fact I _can_ imagine these cases doesn't mean any microscope company ever _would_ or even _has_, which is why I appeal to someone with extensive experience to definitely say 'every microscope company in the world would save plane ID information as an integer?'. That seems too specific a claim to possibly be true, so why should we assume it here?

Whereas 'can every plane ID generated with any device be mapped as a string?' seems much safer to me. Also tautological in the case where the information is available in the file names (whose stems are string) or the metadata stored in a text-based header, because that's literally how it had to have gotten transcoded as bytes anyhow. And we already have cases of that seen witth Bruker



> I think that Paul claim here is really reasonable and your demands too strong.

Sorry if I phrased too broad of a question

I'm only after the most generic solution for a universal API that will not require us to go back and re-implement this when we see it occur in a future format, and to me even if you do nothing more than `int_as_str = str(input_you_expect_to_be_an_int)` in this case I'd be happy; e.g., `plane: str = "0"`.

No matter what, you can always map `int` -> `str` easily, and the latter is much more extensible. The only thing you lose is perhaps ordinal comparison, but that's why I also wonder if it's always necessary that the order in which cycles are captured is necessarily related to another ordinal property such as space (or can planes go back and forth non-linearly?)



> Even if the numbers are not ordered (cycles mostly are though) numbers still seem like the best way to refer to planes (see the cyclic group common representation).

> If give you my format, you give me an object that allows me to extract my data ergonomically and efficiently with a common API. I don't need to know how you do it.

As you say, the _how_ of implementing plane identification and referencing within our code might very well map/order the available plane selections onto a cyclic group since that is how the device would some way or another go about capturing them, and regardless of the identifiers of those planes we can always map them to a finite set of integer indices internally

But as you also say,  the users don't need to know how _we_ do it under the hood.

Let's say a user acquiring multiple planes does not care about this implicit order - why would we force them, via the top level usage of our common API, to know what that integer-sensitive indexing is?

As opposed to just fetching for them the available, unordered set of string options, and allowing them to select the one they want to extract?

Also, if you feel strongly in favor of duck typing, you could always make everyone happy and do both 😄

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: See #246

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
CodyCBakerPhD: Actually, there's already a practical case and we don't have to speculate about what might be possible.

Multi-plane support for reading from NWB files

A good user following best practices might apply an inferrable ordinal labeling of each imaging plane, e.g., `ImagingPlane0`, `ImagingPlane1`, etc.

Or they might have given it something richer and more identifiable.

`ImagingPlaneCortex`

`ImagingPlaneV1`

`ImagingPlaneHippocampus`

Sure we could arbitrarily cast these together into a list (order probably determined by the order in which they were written to the file) and reference to each element by their index in that list; but it wouldn't be as intuitive, or echo that richness of annotation.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: See #247

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: Moving this to a dedicated PR #248

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: Moving this to a dedicated PR #248

pauladkisson: > I'm only after the most generic solution for a universal API that will not require us to go back and re-implement this when we see it occur in a future format, and to me even if you do nothing more than int_as_str = str(input_you_expect_to_be_an_int) in this case I'd be happy; e.g., plane: str = "0".

Ok, I think this is reasonable.  For ScanImage (and I suspect for a lot of formats) using an integer for each plane is the most parsimonious solution.  But, as you pointed out, there may be some cases where strings are more appropriate esp. when analyzing depth planes separately rather than in a volumetric fashion.  More importantly, it is very easy to take `plane_name = '0'` instead of `plane=0` but much harder to go the other way.

h-mayorquin: Those are some good points, @CodyCBakerPhD, thanks for elaborating as I requested. Specially the example of NWB is very illustrative. I will integrate this in my reasoning about https://github.com/catalystneuro/roiextractors/issues/245.

Spikeinterface does have int | str type for channel_ids and I think that that bothers me more than overgenerality.

alessandratrapani: > Maybe @alessandratrapani can shed light on this; for something like Bruker, can a user design/specify a cycle or alternation of cycles that are not strictly increasing/decreasing with respect the Z plane?
I am sorry for coming late for this discussion. 
I was reasoning on edge cases where the plane acquisition is not strictly increasing/decreasing along depth, but I couldn't see how and why it should happen. 
However, this stands just for volumetric imaging acquisition, where you want to record the activity of a specific brain area without caring (a priori) how the neurons are distributed in the volume. 
Imagine the case where you want to record pyramidal cells in a cortical column, but you want to do it in a specific order that follows functional constraints along the layers distribution, not necessarily depth. In this very specific case, you could define a cycle acquisition that acquires planes following a functional order. Therefore, you might also want to name the plane according to the specific region/layer they are referred to. (Note that this is just a speculation of what might come in the future; I have never encountered such cases.)
This is why I endorse the solution of defining `plane_name` as a string that can be just an int_as_str for most of the cases.

pauladkisson:
@CodyCBakerPhD See gin PR [#18](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/18)

pauladkisson:
Continued on #253.

Issue #156, opened 2022-06-21
Add a documentation section about the standard format for `get_frames`


h-mayorquin:
As described in the title. When finish agreeing over final details in  #155.

Quoting @CodyCBakerPhD 
> For example, this tutorial for EXTRACT https://github.com/schnitzer-lab/EXTRACT-public/blob/master/Learning%20materials/Lecture%20Tutorials/Tutorial%201%20-%20Starting%20code/example_tutorial.m

> indicates the data storage as height x width, which we as rows x cols (https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py#L131, though not terribly obvious to read from that code, also not very educational since the example data is square...) but then this will get transposed to NWB standard, which is width x height... It can get very confusing and it would be great to have a nicely documented showcase for it

pauladkisson:
Fixed by #244

PR #244, opened 2023-09-15
Add note about `get_video` convention to `ImagingExtractor`

h-mayorquin:
Related to https://github.com/catalystneuro/roiextractors/issues/156

The ideal will be to add this in a section of the read the docs but we are far from it. The standard should be written somewhere and this is a place.

### DIFF
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -112,22 +112,17 @@ class ImagingExtractor(ABC):
 
         Notes:
         ------
-        This is the central extraction function for the ImagingExtractor instances.
-        It should fetch the requested frames from the instance format.
-
         Importantly, we follow the convention that the dimensions of the array are returned in their matrix order,
         More specifically:
+        (time, height, width)
 
-        (frames or time, height or rows, width or columns, depth or volume)
+        Which is equivalent to:
+        (samples, rows, columns)
 
         Note that this does not match the cartesian convention:
-
-        (t, x, y, z)
+        (t, x, y)
 
         Where x is the columns width or and y is the rows or height.
-
-        It is important to check for each format what is the convention used and to transpose the data to the
-        roi convention if needed.
         """
         pass
 
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -109,6 +109,25 @@ class ImagingExtractor(ABC):
         -------
         video: numpy.ndarray
             The video frames.
+            
+        Notes:
+        ------
+        This is the central extraction function for the ImagingExtractor instances.
+        It should fetch the requested frames from the instance format. 
+        
+        Importantly, we follow the convention that the dimensions of the array are returned in their matrix order,
+        More specifically:
+        
+        (frames or time, height or rows, width or columns, depth or volume)
+        
+        Note that this does not match the cartesian convention:
+        
+        (t, x, y, z)
+        
+        Where x is the columns width or and y is the rows or height.
+        
+        It is important to check for each format what is the convention used and to transpose the data to the
+        roi convention if needed.
         """
         pass
 
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -109,23 +109,23 @@ class ImagingExtractor(ABC):
         -------
         video: numpy.ndarray
             The video frames.
-            
+
         Notes:
         ------
         This is the central extraction function for the ImagingExtractor instances.
-        It should fetch the requested frames from the instance format. 
-        
+        It should fetch the requested frames from the instance format.
+
         Importantly, we follow the convention that the dimensions of the array are returned in their matrix order,
         More specifically:
-        
+
         (frames or time, height or rows, width or columns, depth or volume)
-        
+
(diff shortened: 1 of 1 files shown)


pauladkisson (changes requested):
In general, `get_video` only returns 3D numpy ndarrays (frames, height, width).  Depth is handled only for some extractors, so they should override their `get_video` docstring to specify how it is treated.

h-mayorquin:
Every method needs to override the whole function and therefore also the docstring as this is an abstract method. In other words, this is not user facing documentation misguiding the users but documentation for developers with instructions of a common pitfall when they overide the method. In that sense, I think it is better served if it has the more general case, more information is better than less in this case.

pauladkisson:
> Every method needs to override the whole function and therefore also the docstring as this is an abstract method.

Most of the current `ImagingExtractors` return video exactly as it's described in the base class, so they can inherit the docstring without needing to rewrite it.  I believe that is one of the pros of defining an abstract `get_video` function.

> In that sense, I think it is better served if it has the more general case, more information is better than less in this case.

More information can be helpful for devs, but let's at least mention that the return `video` can be 3D or 4D depending on whether the extractor defines depth.

h-mayorquin:
> > Every method needs to override the whole function and therefore also the docstring as this is an abstract method.
> 
> Most of the current `ImagingExtractors` return video exactly as it's described in the base class, so they can inherit the docstring without needing to rewrite it. I believe that is one of the pros of defining an abstract `get_video` function.

Maybe you have a plan that I don't understand. How are you going to inherit the docstring of the abstract method? 

This seems like a non-standard trick, is that something that you guys have discussed, @CodyCBakerPhD ?

pauladkisson:
> How are you going to inherit the docstring of the abstract method?

The child class inherits the docstrings of the parent methods automatically even if it overrides the body of the method -- as long as it doesn't define its own docstring.

h-mayorquin:
It does not seem that this is true to me. If it is, what's missing in the code below?

```python
from abc import ABC, abstractmethod

class AbstractClass(ABC):
    """
    This is the AbstractClass's docstring.
    """
    
    @abstractmethod
    def abstract_method(self):
        """
        This is the abstract_method's docstring.
        """
        pass
    
    @abstract_method
    def another_abstract_method(self):
        """
        This is the another_abstract_method's docstring.
        """
        pass
    
    def a_normal_method(self):
        """Parent docstring"""

class ChildClass(AbstractClass):
    
    def abstract_method(self):
        return "Implemented in ChildClass"

    def another_abstract_method(self):
        """This one has a docstring"""
        return "Implemented in ChildClass"
    
    def a_normal_method(self):
        return "Do I override the parent's docstring?"
# Test
child = ChildClass()
print(child.abstract_method())
print(child.abstract_method.__doc__)
print(child.another_abstract_method.__doc__)
print(child.a_normal_method())
print(child.a_normal_method.__doc__)```
Output:
```
Implemented in ChildClass
None
This one has a docstring
Do I override the parent's docstring?
None
```

**Comment by h-mayorquin** [comment]
It would be nice to get a response for this, @pauladkisson. I think that you are wrong in your claim above.

**Comment by CodyCBakerPhD** [comment]
@h-mayorquin Note that directly accessing the `__doc`__ is not how we generally instruct users to view docstrings on instances of objects

Check out our recommended ways

```
child.abstract_method?
help(child.abstract_method)
```

This discrepancy is still good for all of us to take note of

**Comment by h-mayorquin** [comment]
That seems convicing to me:

image

I will implement the requested changes.

**Comment by pauladkisson** [comment]
@h-mayorquin, sorry for the delayed response.  Basically, as Cody pointed out, dynamic docstring look-up via `help` finds inherited docstrings, even though `__doc__` doesn't.  We are taking advantage of that functionality to only write docstrings for the base class methods (`get_num_channels`, `get_image_size`, etc.) in the base class implementation UNLESS a child class specifically changes its inputs/outputs, in which case it would need to write its own docstring for the altered method.

**Comment by h-mayorquin** [comment]
That's fine, this was useful to me. I was not aware that python performs attribute lookup even for attributes of the class' methods when using the help builtin. At the beginning I thought that this was ipython / jupyter behavior but I also tried on the python REPL and is the same behavior.

**Comment by h-mayorquin** [comment]
Changed this.

File: src/roiextractors/imagingextractor.py [hunk 1]
```diff
@@ -109,6 +109,20 @@ def get_video(
         -------
         video: numpy.ndarray
             The video frames.
+
+        Notes:
+        ------
```
pauladkisson, line 114: pydocstyle is complaining about formatting -- this should fix it.
```suggestion
        Notes
        -----
```

pauladkisson (changes requested):
One formatting change otherwise LGTM

src/roiextractors/imagingextractor.py
h-mayorquin: Thanks, do you test this locally usually? any advice on how to go about this?

pauladkisson: Yes, if you install pre-commit locally then it will check all of your local commits so that you can fix as you go.  That's how I've been using it.

pauladkisson (approved):

PR #248, opened 2023-09-20
VolumetricImagingExtractor

### LINKED ISSUE #245: Unify Volumetric/MultiPlane Imaging API
# Current Behavior

Right now, the only imaging extractor that supports 3D imaging is BrukerTiff.  It does so by taking a list of `SinglePlaneImagingExtractors` and combines them a la `MultiImagingExtractor`.  Note that each `SinglePlaneImagingExtractor` is itself a `MultiImagingExtractor` that combines multiple 'chunked' files corresponding to a single depth plane.  Each `SinglePlaneImagingExtractor` instance only supports 1 channel and 1 plane, which are passed via `stream_name` during the `__init__`.  Each `MultiPlaneImagingExtractor` instance supports 1 channel (but all planes) which is passed via `stream_name` during the `__init__`.

# Issues

- 'streams' are a low-cohesion concept to represent optical channels and depth planes.
  - They are referred to by string names, which can be different in the `MultiPlaneImagingExtractor` (which only needs to specify the channel) and the `SinglePlaneImagingExtractor` (which needs to specify some formatted combination of channel and plane).
  - They are supposed to be analogous to 'AP', 'LF', etc. streams in spikeinterface, but those streams refer to collections of channels that undergo different types of pre-processing.  In contrast, each stream as they are implemented in BrukerTiff just refer to 1 channel and 1 plane.  There may be a relevant parallel in ophys (ex. functional vs anatomical streams), but specifying a plane/channel combo is not that parallel.
  - optical channels and depth planes are fully independent concepts (1 plane + many channels, 1 channel + many planes, many channels + many planes are all possible) with rich associated metadata that can guide the API: channel name, channel wavelength, plane depth, etc.  By lumping these two axes together as 'streams' we lose the ability to interact with them separately.
- Combining multiple `SinglePlaneImagingExtractors` with `MultiImagingExtractor` is a bit hacky and confusing.
  - `MultiImagingExtractor` was designed to concatenate `ImagingExtractors` along the `frames` axis, but here we want to combine multiple `IEs` along a separate depth axis.
  - Using `MultiImagingExtractor` for this purpose leads to wonky stuff like `start_frames = [0, 0, ..., 0]`.
- BrukerTiff specifies channel during the `__init__` but all other IE's specify channel during the call to `get_video`.
  - Not sure which is better, but we should be consistent.

# Proposed Solutions

- Specify channel and plane separately by index (or ID?) and then the extractor handles the logic of how to load that channel/plane combo in their file structure (Bruker, ScanImage, etc.)
- Add a `MultiPlaneImagingExtractor` base class to handle the combination of IE's along the depth axis.  Could come with extra metadata properties like `get_plane_depth`.
- We need to specify plane at `__init__` time so that SinglePlaneImagingExtractors can be appropriately combined into MultiPlaneImagingExtractors.  So, to keep it consistent we should specify channel also at `__init__` time.  This will require refactoring of all the existing IE's but I think it is the most convenient implementation agnostic to file structure. It also simplifies `get_video` to only take `start_frame` and `end_frame`, which I think is a good thing.

Yes.

### Code of Conduct

pauladkisson:
Fixes #245 

See also discussion in #241


src/roiextractors/volumetricimagingextractor.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class VolumetricImagingExtractor(ImagingExtractor):
+    """Class to combine multiple ImagingExtractor objects by depth plane."""
+
+    extractor_name = "VolumetricImaging"
+    installed = True
+    installatiuon_mesage = ""
+
+    def __init__(self, imaging_extractors: List[ImagingExtractor]):
```
h-mayorquin, line 15: Wiill all volumetric imaging extractors be a list of imaging extractors. We are dealing with a volumetric extractor that is not like that in the current conversion: two files for two channels and each has volumetric data.

pauladkisson: Yeah, so the idea would be to make a separate ImagingExtractor for each plane and then combine them with the VolumetricImagingExtractor.

CodyCBakerPhD:
I will recruit @alessandratrapani to help review this PR as well - maybe we'll go over it at the retreat?

alessandratrapani (commented):
Looks good to me, just a minor correction

src/roiextractors/volumetricimagingextractor.py
```diff
@@ -0,0 +1,167 @@
...
+        ----------
+        start_frame: int, optional
+            Start frame index (inclusive).
+        end_frame: int, optional
+            End frame index (exclusive).
+
+        Returns
+        -------
+        video: numpy.ndarray
+            The 3D video frames (num_rows, num_columns, num_planes).
```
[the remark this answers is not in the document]
pauladkisson: Thanks for the catch!

CodyCBakerPhD (approved):

PR #251, opened 2023-09-28
Fixed Github Workflows

pauladkisson:
Updated AWSCLI version to fix github workflows.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,10 +4,6 @@ on:
     - cron: "0 0 * * *"  # daily
   pull_request:
   workflow_dispatch:
-  workflow_run:
-    workflows: ["Update testing data"]
-    types:
-      - completed
 
 jobs:
   run:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,9 +1,10 @@
 name: Full Tests
 on:
-  schedule:
-    - cron: "0 0 * * *"  # daily
   pull_request:
   workflow_dispatch:
+  workflow_run:
+    workflows: [update-testing-data]
+    types: [completed]
 
 jobs:
   run:
@@ -33,13 +34,12 @@ jobs:
 
       - name: Install roiextractors with minimal requirements
         run: pip install .[test]
+
       - name: Run minimal tests
         run: pytest tests/test_internals -n auto --dist loadscope
 
       - name: Test full installation (-e needed for codecov report)
         run: pip install -e .[full]
-
-
 
       - name: Get ophys_testing_data current head hash
         id: ophys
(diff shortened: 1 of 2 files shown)


.github/workflows/run-tests.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       - if: steps.cache-ophys-datasets.outputs.cache-hit == false
         name: Install and configure AWS CLI
         run: |
-          pip install awscli==1.25.27
+          pip install awscli==1.29.56
```
CodyCBakerPhD, line 56: My suggestion was to completely remove the step from this workflow so you don't run into this problem again and to keep the steps more modular in general

pauladkisson:
@CodyCBakerPhD, FYI one can run a workflow based on the conclusion of another workflow without invoking complicated meta-workflows (see [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow).  But, since just updating the awscli version fixed the issue, I don't think we need to even make that change.

CodyCBakerPhD:
> @CodyCBakerPhD, FYI one can run a workflow based on the conclusion of another workflow without invoking complicated meta-workflows (see [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow). But, since just updating the awscli version fixed the issue, I don't think we need to even make that change.

Cool, thanks for that - good to know they added that capability

.github/workflows/run-tests.yml
pauladkisson: Ok, I must be misuderstanding something bc when I remove the awscli call from the run-tests workflow, I get an error: `Failed: No folder found in location: ophys_testing_data!`

See [Action #230](https://github.com/catalystneuro/roiextractors/actions/runs/6342979572)

pauladkisson: What if we just removed `update-testing-data.yml` since it's redundant?

pauladkisson: Or is the whole point of that workflow for debugging?

CodyCBakerPhD: > Ok, I must be misuderstanding something bc when I remove the awscli call from the run-tests workflow, I get an error:

Because you didn't just remove the AWS CLI part, but also the caching (which is the thing that has the folder): https://github.com/catalystneuro/roiextractors/blob/ce624f8dfc07bd0e89dbeb2a59e4ffc0125e082e/.github/workflows/run-tests.yml#L48-L65

> What if we just removed update-testing-data.yml since it's redundant?

Other way around

> Or is the whole point of that workflow for debugging?

Not for debugging, but as the primary way of updating the caches for all platform/python versions from a common source while also keeping their environments (downloading data vs. running tests) completed isolated to avoid stuff like this forever into the future

CodyCBakerPhD: It's only these lines: https://github.com/catalystneuro/roiextractors/blob/ce624f8dfc07bd0e89dbeb2a59e4ffc0125e082e/.github/workflows/run-tests.yml#L57-L65

That you would have to remove; and perhaps shift that CRON trigger on the update job when relying on the scheduling

pauladkisson: Ahh, I see thanks -- I didn't catch that the cache action grabs the files if they've already been downloaded.

.github/workflows/run-tests.yml
```diff
@@ -1,9 +1,10 @@
 name: Full Tests
 on:
-  schedule:
-    - cron: "0 0 * * *"  # daily
   pull_request:
   workflow_dispatch:
+  workflow_run:
+    workflows: [update-testing-data]
+    types: [completed]
```
CodyCBakerPhD, line 7: If going this route, you'll probably want to add the 'on success' logic to allow early exit (otherwise will proceed to run tests and fail when it doesn't find a certain file/folder); I wonder if `success` is an allowed type for the trigger itself?

pauladkisson: Yeah that seems reasonable.  Apparently `on: workflow_run` only works on the default branch ([see Issue](https://github.com/orgs/community/discussions/25288)), so I can't test it before merging? Kind of annoying.

CodyCBakerPhD: > Yeah that seems reasonable. Apparently on: workflow_run only works on the default branch ([see Issue](https://github.com/orgs/community/discussions/25288)), so I can't test it before merging? Kind of annoying.

This kind of thing is typical of GitHub Action devops stuff - it's the one situation where committing to main to debug stuff is OK IMO

I'll approve and merge this and let you continue messing with it

CodyCBakerPhD (approved):

PR #252, opened 2023-09-28
added success/failure logic to Full Tests workflow

pauladkisson:
Continues from #251

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -12,7 +12,9 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     steps:
-      - run: echo 'The triggering workflow failed.'
+      - run: |
+           echo 'The triggering workflow failed.'
+           0
   on-success:
     name: Full tests on ${{ matrix.os }} with Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


pauladkisson:
Looks like this works, needs to merge into main to properly test the triggering tho.

.github/workflows/run-tests.yml
```diff
@@ -7,7 +7,13 @@ on:
     types: [completed]
 
 jobs:
-  run:
+  on-failure:
+    name: Notify on failure
+    runs-on: ${{ matrix.os }}
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - run: echo 'The triggering workflow failed.'
```
CodyCBakerPhD, line 15:
```suggestion
    steps:
      - run: |
           echo 'The triggering workflow failed.'
           0
```

You'll also want to return status 0 to the CI in order for it to not deliver a false positive (check mark as 'passed' if it didn't even run)

Example: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/.github/workflows/deploy-tests.yml#L25-L27

CodyCBakerPhD: Though the indenting in my suggestion may need to be corrected (two spaces from the start of the 'run' string)

CodyCBakerPhD (approved):

PR #253, opened 2023-09-28
Multi-Plane and Multi-Channel Support for ScanImage

### LINKED ISSUE #240: [Feature]: Multi-channel and Multi-plane support for ScanImage

pauladkisson:
Fixes #240
Continues from #241 rebased from #248


pauladkisson:
The one thing that I never really figured out is how to deal with the times for the volumetric imaging.  Still not really sure how that should be represented.  Right now the times are accessible in each SinglePlaneImagingExtractor.  Other than that this should be good to go.

CodyCBakerPhD:
> The one thing that I never really figured out is how to deal with the times for the volumetric imaging. Still not really sure how that should be represented. 

It looks like our base classes themselves still need to be updated to the latest SI standard (`get_times`) so fine to leave that as a follow-up

I know we talked about having an array of times to indicate the shifts for each plane, but for consistency with the other methods like `get_video`, I think the simplest return for such a function would just be the timestamps of the first plane capture in each volume, for each frame

Then we can talk about a separate method for volumes that returns a more detailed structure and get feedback from the rest of the team on that

CodyCBakerPhD:
This LGTM, very well tested!

@alessandratrapani How does it look to you?

alessandratrapani (approved):
Looks good to me too, we were discussing about it just few minutes ago and I think this should work properly!

Issue #14, opened 2020-08-14
get_num_rois vs no_rois


bendichter:
`SegmentationExtractor.get_num_rois` and `SegmentationExtractor.no_rois` appears to be redundant. I would prefer `SegmentationExtractor.get_num_rois`.

Saksham20:
fixed here: c25127149547da5c8266d6021d9de7238b9f1997

PR #256, opened 2023-10-23
Docstring tests

pauladkisson:
Implements docstring checks from catalystneuro/.github instead of locally on the repo.  See [.github #14](https://github.com/catalystneuro/.github/pull/14).


.github/workflows/check-docstrings.yaml
```diff
@@ -0,0 +1,12 @@
+name: Check Docstrings
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  check-docstrings:
+    uses: catalystneuro/.github/.github/workflows/check_docstrings.yaml@docstring_tests
```
CodyCBakerPhD, line 8:
```suggestion
    uses: catalystneuro/.github/workflows/check_docstrings.yaml@main
```

I'd say in terms of file organization, if you could arrange things to work as above it would be preferred

pauladkisson: I'd rather the workflows live in .github/workflows so that we can run them from the .github repo directly (very useful for development and debugging).

CodyCBakerPhD (approved):

PR #268, opened 2023-12-05
Two small improvements to `BrukerTiffMultiPlaneImagingExtractor`

h-mayorquin:
1) Avoid calculating calling `determine_imaging_is_volumetric` which parses the whole xml file twice. The code to get the streams already calculates this and if returns sucesfully the streams have `plane_streams`:

https://github.com/catalystneuro/roiextractors/blob/467bbb00c20d79ecdd383b39aa1ced2734fcb33e/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py#L114-L116

2) When stream_name is not passed and there are multiple channels we can make the life of the user easier by already displaying the options which we have already available.


CodyCBakerPhD:
Leaving this one to @alessandratrapani as well

alessandratrapani (approved):
Way better!

PR #270, opened 2023-12-19
Extract roi indices from redcell.npy in Suite2pExtractor

alessandratrapani:
In the case of two-channel acquisition, where one of the channels is a "red" channel holding anatomical information, suite2p outputs also a .npy file named redcell.npy. Similarly to iscell.npy, redcell.npy it's a binary array in which ones represent rois which actually correspond to a cell, identified in the imaging data stream acquired with second channel, and matched the cell identified in the imaging data stream acquired with the first channel. 

I just add few lines of code to extract the accepted/rejected rois ids from redcell.npy instead iscell.npy.

### DIFF
--- a/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
+++ b/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py
@@ -166,7 +166,13 @@ class Suite2pSegmentationExtractor(SegmentationExtractor):
             self._load_npy(file_name="spks.npy", mmap_mode="r", transpose=True) if channel_name == "chan1" else None
         )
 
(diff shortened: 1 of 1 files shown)


alessandratrapani:
> @alessandratrapani You confirmed that this `redcell` file is in fact generated by the Suite2P software? And is an official part of the Suite2P file standard

Sorry for the late reply. Yes I confirm it is generated from Suite2p: 
https://github.com/MouseLand/suite2p/blob/193e7f1f656bfbd1c100eb51411737c80f54ac3c/suite2p/gui/io.py#L400-L405

CodyCBakerPhD (approved):

PR #272, opened 2024-01-22
Add function to extract ROIs metadata from ScanImage tiff

alessandratrapani:
In the ScanImage .tif metadata there is a dictonary saved as text that the function `extract_extra_metadata` cannot parse, where information on the ROIs are stored. I implemented a new function call `extract_roi_metadata` (https://github.com/catalystneuro/roiextractors/pull/272) that return this dict.
```
"RoiGroups":{
    "imagingRoiGroup":{# general imaging space metadata (only 2D)
        ...
        "centerXY": [-1.776356839e-16,0.1],
        "sizeXY": [13.5,13.5], # size of the FOV in micrometers
        "rotationDegrees": 0,
        "enable": 1,
        "pixelResolutionXY": [512,512],
        "pixelToRefTransform": [[0.0263671875,0,-6.763183594],
                                [0,0.263671875,-6.663183594],
                                [0,0,1]],
        "affine":  [[13.5,0,-6.75],
                    [0,13.5,-6.65],
                    [0,0,1]]
        ...
    },
    "integrationRoiGroup":{
        ...
        "rois": [
        {
          "ver": 1,
          "classname": "scanimage.mroi.Roi",
          "name": "ROI 1",
            ...
          "zs": 0,
          "scanfields": {
            ...
            "centerXY": [2.518066406,1.062402344],
            "sizeXY": [0.2373046875,0.2373046875],
            "rotationDegrees": 0,
            "enable": 1,
            "threshold": 100,
            "channel": 1,
            "processor": "cpu",
            "mask": [roi_pixel_mask 9x9 binary matrix],
            "affine": [[0.2373046875,0,2.518066406],
                       [0,0.2373046875,1.062402344],
                       [0,0,1]]
          },
          "discretePlaneMode": 1,
          "powers": null,
          "pzAdjust": null,
          "Lzs": null,
          "interlaceDecimation": null,
          "interlaceOffset": null,
          "enable": 1
        },
        ... # all the 368 ROIs
        ]
    } # imaging space metadata for each ROI
}
```
From `imagingRoiGroup` we can extract info on the dimension of the field of view (XY) from `"sizeXY"` and divide by the dimension to get the grid_spacing (z dimension missed)

In `integrationRoiGroup` we can extract information about the position of the rois: inv(`pixelToRefTransform`)*`centerXY`


pauladkisson (changes requested):
Thank you for catching this. It looks like there's a whole swath of metadata stored in a json-style string that we haven't been extracting. 

However, I would prefer if this were integrated into the existing `extract_extra_metadata` function. That way the metadata function would return all the metadata in the 'system_metadata_string' AND the 'extra_metadata_string' (as you called them). Then you can add a section to grab the ROI groups in the `parse_metadata` function.

pauladkisson (approved):
LGTM! Thanks for implementing the changes.

PR #273, opened 2024-01-22
clean:

bendichter:
typos
fix docstrings
rmv unused import


src/roiextractors/extraction_tools.py
```diff
@@ -325,7 +325,7 @@ def _image_mask_extractor(pixel_mask, _roi_ids, image_shape):
     return image_mask
 
 
-def get_video_shape(video):
+def get_video_shape(video) -> Tuple[int, int, int, int]:
```
pauladkisson, line 328: This function is only used in the `NumpyImagingExtractor`, so it might make sense to move it there rather than in extraction_tools.py.

Also, the order of the shape is incorrect based on usage in `NumpyImagingExtractor`: It should be `num_frames, size_x, size_y, num_channels`.

pauladkisson (changes requested):
Everything looks good except an error in the `get_video_shape` function.

pauladkisson (approved):

PR #276, opened 2024-01-23
add inscopix imaging extractor

bendichter:


.github/workflows/run-tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11]
+        python-version: ["3.9", "3.10", "3.11"]
```
CodyCBakerPhD, line 24: Why the strong drop of 3.8 10 months before Python stops supporting it?

Even if `isx` doesn't support it, you can have conditional requirements that say that `roiextractors` + `isx` support requires `py>3.8`, but the rest of `roiextractors` is cool with it

bendichter: Let's follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), not the [Python release cycle](https://devguide.python.org/versions/). According to NEP 29, Python 3.8 was dropped Apr 14, 2023.

CodyCBakerPhD: We've been following the Python release cycle everywhere else - maybe discuss in the next team meeting?

bendichter: The latest version of numpy, 1.26.0, requires Python 3.9+. If we want to use any new numpy features (or any features that rely on them) we will need to drop support for Python 3.8.

CodyCBakerPhD: What new features would those be, and why would ROIExtractors, NeuroConv, or the GUIDE minimally require them instead of simply letting it be up to the user and their condition installations?

You can also specify `numpy>=1.26.0` conditionally for various Python versions and optional requirements without making the strong decision to retire an entire Python version

bendichter: > This NEP is being put forward by maintainers of Matplotlib, scikit-learn, IPython, Jupyter, yt, SciPy, NumPy, and scikit-image.

Which means that the latest version of all of those libraries is intended to be 3.9+. See for example https://pypi.org/project/scikit-learn/

bendichter: This has been put forth by all of those projects as a best practice for ecosystem support. I don't really want to create any unnecessarily packaging complexity to extend beyond this window

h-mayorquin: @bendichter 
I was not aware of NEP 29. Nice.

I also think that there are minimal costs to dropping support of python 3.8. I think our user base has a lot of freedom over their environments as conda is available on most scientific contexts. As Ben said, I think the cost of increasing packacing complexity is a greater burden for us. If some user comes forward with an issue maybe we can carve out a solution for them but I think it would be easier for us to stick to the scientific python standards.

bendichter:
blocked by https://github.com/inscopix/py_isx/issues/48

pauladkisson:
Looks like this is unblocked as of v1.0.4 (https://github.com/inscopix/py_isx/issues/48#issuecomment-2033057806)

requirements-full.txt
```diff
@@ -2,3 +2,4 @@ tifffile>=2018.10.18
 scanimage-tiff-reader==1.4.1
 neuroconv[video]>=0.4.6  # Uses the VideoCaptureContext class
 natsort>=8.3.1
+isx
```
CodyCBakerPhD, line 5:
```suggestion
isx>=1.0.4
```

src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py
```diff
@@ -0,0 +1,54 @@
+"""
+Inscopix Imaging Extractor
+"""
```
CodyCBakerPhD, line 3:
```suggestion
"""Inscopix Imaging Extractor."""
```

src/roiextractors/extractors/inscopixextractors/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""
+Defines extactors for Inscopix data.
+"""
```
CodyCBakerPhD, line 3:
```suggestion
"""Defines extactors for Inscopix data."""
```

src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py
```diff
@@ -0,0 +1,52 @@
...
+
+    extractor_name = "InscopixImaging"
+
+    def __init__(self, file_path: PathType):
+        """
+        Parameters
+        ----------
+        file_path : PathType
+            Path to the Inscopix file.
+        """
```
CodyCBakerPhD, line 23:
```suggestion
        """
        Create an InscopixImagingExtractor instance from a single .isx file.

        Parameters
        ----------
        file_path : PathType
            Path to the Inscopix file.
        """
```

CodyCBakerPhD:
Looks like the test cases just need to be updated now - also a CHANGELOG entry

CHANGELOG.md
```diff
@@ -2,11 +2,17 @@
 
 ### Features
 
-* Updated testing workflows to include python 3.12, m1/intel macos, and dev tests to check neuroconv: [PR #317](https://github.com/catalystneuro/roiextractors/pull/317)
+* Add InscopixSegmentationExtractor [#276](https://github.com/catalystneuro/roiextractors/pull/276)
```
CodyCBakerPhD, line 5:
```suggestion
* Add InscopixImagingExtractor [#276](https://github.com/catalystneuro/roiextractors/pull/276)
```

bendichter:
This is finally working. Though I think dropping 3.8 should probably be a different PR

CodyCBakerPhD:
I'll defer to @pauladkisson for PR review but mostly looks good to me

> Though I think dropping 3.8 should probably be a different PR

For the global drop of 3.8 I would agree, and then we can do it synchronously throughout the ecosystem across the next few weeks (I think we have a couple of months still)

However, you would still need various skips of `isx` stuff for 3.8 (such as in the CI installation step too) for this PR so maybe too much bother here?

Could do the 3.8 drop in separate PR, give that PR precedence, then come back to this?

pauladkisson:
Happy to review, but will have to wait until 5/20

h-mayorquin:
https://github.com/catalystneuro/neuroconv/pull/853

bendichter:
depends on https://github.com/catalystneuro/roiextractors/pull/325

tests/test_inscopiximagingextractor.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.mark.skipif(sys.version_info <= (3, 8), reason="Requires Python 3.9 or higher")
+def test_inscopiximagingextractor_movie_128x128x100_part1():
+    file_path = OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "movie_128x128x100_part1.isxd"
+    extractor = InscopixImagingExtractor(file_path=file_path)
+
+    assert extractor.get_num_frames() == 100
+    assert extractor.get_image_size() == (128, 128)
+    assert extractor.get_dtype() == dtype("float32")
```
h-mayorquin, line 20: This is geting the method from the base class which makes a call to a read operation to get the dtype. 

To avoid this read operation everytime that someeone calls dtype maybe we can determine this at `__init__`

I was hoping that this information was available on the footer and there is a dtype but I am not sure how to map it to numpy dtypes. Check out the footer for this test:

```
print(extractor.movie.footer)
{'dataType': 1, 'extraProperties': None, 'fileVersion': 1, 'hasFrameHeaderFooter': False, 'producer': {'name': 'isxcore', 'version': [1, 8, 0]}, 'spacingInfo': {'numPixels': {'x': 128, 'y': 128}, 'pixelSize': {'x': {'den': 1, 'num': 3}, 'y': {'den': 1, 'num': 3}}, 'topLeft': {'x': {'den': 1, 'num': 0}, 'y': {'den': 1, 'num': 0}}}, 'timingInfo': {'blank': [], 'cropped': [], 'dropped': [], 'numTimes': 100, 'period': {'den': 1000, 'num': 100}, 'start': {'secsSinceEpoch': {'den': 1, 'num': 0}, 'utcOffset': 0}}, 'type': 0}
```

h-mayorquin (commented):
This looks good, my only request is that we avoid the read operation every time dtype is requested.

tests/test_inscopiximagingextractor.py
h-mayorquin: Ah, this can be extracted from `self.movie.data_type` see:

https://inscopix.github.io/py_isx/how-to/read-movie-data-type.html

tests/test_inscopiximagingextractor.py
```diff
@@ -0,0 +1,52 @@
...
+    assert extractor.get_num_frames() == 1248
+    assert extractor.get_image_size() == (33, 29)
+    assert extractor.get_dtype() == dtype("uint16")
+    np.testing.assert_almost_equal(extractor.get_sampling_frequency(), 5.5563890139076415)
+    assert extractor.get_channel_names() == ["channel_0"]
+    assert extractor.get_num_channels() == 1
+    assert extractor.get_video().shape == (1248, 33, 29)
+
+
+@pytest.mark.skipif(sys.version_info <= (3, 8), reason="Requires Python 3.9 or higher")
```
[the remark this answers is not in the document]
bendichter: I'm on the fence. I was keeping them on in case users run 3.8 tests. On the other hand, maybe it's best for them to fail to inform the user that inscopix is not supported on their system

bendichter:
@h-mayorquin I added a method as you suggested. I also added a test to ensure that the dtype of the data indeed matches this dtype extacted from the metadata (this did not match in a previous version of the isx package)

h-mayorquin (approved):
All right.

PR #282, opened 2024-02-05
Update Docs

pauladkisson:
Update documentation with new info previously found in the readme, and some style/CI changes.


CodyCBakerPhD:
OK goal here is to get the docs to build successfully on the PR with the pydata theme

pauladkisson:
Is there a way for me to get access to the readthedocs account associated with roiextractors?

CodyCBakerPhD:
> Is there a way for me to get access to the readthedocs account associated with roiextractors?

I'd need your RTD username

pauladkisson:
> I'd need your RTD username

It's pauladkisson

pauladkisson:
> Out of curiosity, are you unable to see the details from the log build without admin access? https://readthedocs.org/projects/roiextractors/builds/23351775/

Yeah, turns out I could see pretty much all the details without full access, I just wanted to be sure I wasn't missing anything.

pauladkisson:
This is now successfully building the docs but throws out a bunch of (non-breaking) import errors when I build locally, so the API documentation seems like it still needs some work. Nevertheless, I think I would like to merge this PR here, and later swing back around to fix the API documentation.

CodyCBakerPhD:
Indeed looks as  if the API docs will take some work to cleanup; could you at least suppress them here so they don't show up that way on `main`, then polish in follow-up?

pauladkisson:
> Indeed looks as if the API docs will take some work to cleanup; could you at least suppress them here so they don't show up that way on `main`, then polish in follow-up?

Done

CodyCBakerPhD:
It might be kind of petty, but I'd prefer to have the section numbering removed - I don't think it really adds much here; what do you think?

CodyCBakerPhD:
Also on main page, what are STK and FLI formats? Have I forgotten about those extractors?

CodyCBakerPhD:
Also on main page, we don't save to NWB format with this package (which is only for reading) so I'd remove that or potentially add a cross link to NeuroConv to let people know they can use that to write to NWB

alessandratrapani:
> Ahhh, just noticed this as well on the front page
> 
> > ROI
> > Stands for Region Of Interest, which is the region in a set of acquired fluorescence images which the segmentation software has determined as a neuron.
> 
> Not entirely true; perhaps change to
> 
> "which is the set of pixels from acquired fluorescence images which the segmentation software has determined to follow a particular cellular structure." or something similar? Maybe @alessandratrapani can improve description even more
> 
> We've seen cases in last years of ROIs being (i) glia, not neurons, (ii) dendritic branches or axon trajectories, (iii) segmentation by cell type based on morphology, (iv) edge case of considering neurotransmitter densities as the ROI of fibre photometry

I agree with you. I think the description it's already very good.

CodyCBakerPhD (approved):

PR #286, opened 2024-02-20
Update Version

pauladkisson:
This PR manually updates the version of ROI extractors to accommodate latest release.

It also sets up a github workflow to automatically increment the patch version number upon every subsequent release. But, the workflow needs to be in the default branch before it can be deployed and tested. So, the workflow file will likely need a  follow up PR for debugging.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ if not gin_config_file_local.exists():
 
 setup(
     name="roiextractors",
-    version="0.5.6",
+    version="0.5.7",
     author="Heberto Mayorquin, Szonja Weigl, Cody Baker, Ben Dichter, Alessio Buccino",
     author_email="ben.dichter@gmail.com",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -22,7 +22,7 @@
 
 setup(
     name="roiextractors",
-    version="0.5.6",
+    version="0.5.7",
     author="Heberto Mayorquin, Szonja Weigl, Cody Baker, Ben Dichter, Alessio Buccino",
```
CodyCBakerPhD, line 26: You should add yourself here 😉

CodyCBakerPhD (approved):

CodyCBakerPhD:
We'll see then

CodyCBakerPhD:
I suspect you'll have to do some `git` operations to add, commit, push the modified file

PR #290, opened 2024-02-20
Version workflow

pauladkisson:
Added automatic version update workflow file that will run after publishing a new release to pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ if not gin_config_file_local.exists():
 
 setup(
     name="roiextractors",
-    version="0.5.7",
+    version="0.5.6",
     author="Heberto Mayorquin, Szonja Weigl, Cody Baker, Ben Dichter, Alessio Buccino, Paul Adkisson",
     author_email="ben.dichter@gmail.com",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -22,7 +22,7 @@
 
 setup(
     name="roiextractors",
-    version="0.5.7",
+    version="0.5.6",
```
CodyCBakerPhD, line 25: Mmmm shouldn't it be `0.5.7` though? There's also no harm at all in jumping one as a test to `0.5.8`

pauladkisson: Ok, I was going to trigger the workflow one more time (0.5.6-->0.5.7) just to be extra sure it works, but I guess it's even better just to cut a new release and check that everything goes automatically

CodyCBakerPhD:
> better just to cut a new release and check that everything goes automatically

Don't even need to release; I trust that trigger condition (and can be debugged next time you do a release if not), since there's a manual dispatch can just do it that way

CodyCBakerPhD (approved):

Issue #287, opened 2024-02-20
Alternate name than 'neuropil' when saving Caiman denoised calcium traces?


EricThomson:
First, thanks a lot for making neuroconv and roiextractors. I just started today and it was incredibly  easy to use -- I got things working really quickly with it.  I have an annoying question about nomenclature, as I'm relatively new to this I'm probably just missing something.

I'm planning to rebuild Caiman's save nwb functionality using neuroconv. My main question is about the nomenclature. If I understand correctly, the three time series being saved are  `Neuropil`, `Deconvolved`, and `DfOverF`. These correspond to Caiman's `estimates.C` (denoised calcium traces), `estimates.S` (spike counts from deconvolution), and `estimates.F_dff`. 

 My main concern/question is whether there might be another name available within the nwb ophys universe for Caiman's denoised calcium traces in C than `Neuropil` (my concern is that 'neuropil' typically refers to background noise that nobody wants to look at)? E.g., is there some taxonomy of time series in nwb ophys that includes other types of calcium time series?

I apologize for bringing up a terminological rather than substantive concern here -- it is only because everything else is working so well!

CodyCBakerPhD:
> I apologize for bringing up a terminological rather than substantive concern here -- it is only because everything else is working so well!

We love these kinds of discussions. I will let @alessandratrapani explain our current understanding as we've been discussing exactly these types of things for the past ~4 months now

alessandratrapani:
Hi @EricThomson! 
I see the problem here, which is also related to this [issue](https://github.com/catalystneuro/roiextractors/issues/249#issuecomment-1955711025).

- `estimates.C` (denoised calcium traces) should be stored in `Fluorescence` as a `ROIResponseSeries` (`_roi_response_raw` in the `SegmentationExtractor`), 
- `estimates.S` (spike counts from deconvolution) should also be stored in `Fluorescence` as a `ROIResponseSeries` named `Deconvolved` (`_roi_response_deconvolved` in the `SegmentationExtractor`)
- `estimates.F_dff` --> in `DfOverF` as a `ROIResponseSeries` named `DfOverF`  (`_roi_response_dff` in the `SegmentationExtractor`), 
- `estimates.f` (temporal background components) should be stored in Fluorescence as a `ROIResponseSeries` named either `Neuropil` or `Background` (`_roi_response_neuropil` in the `SegmentationExtractor`). Here, we should also add the spatial background components `estimates.b` and create the appropriate plane segmentation.

[...]

PR #291, opened 2024-02-21
Add _roi_response_denoised to store denoised traces 

### LINKED ISSUE #249: [Bug]: Background Components are mixed up in CaimanSegmentationExtractor
# Expected Behavior

According to the [CaImAn docs](https://caiman.readthedocs.io/en/master/Getting_Started.html#basic-structure), `estimates.C` refers to the 'temporal components', which should be stored as `_roi_response_raw` in the `SegmentationExtractor` and `estimates.f` refers to the 'temporal background components', which should be stored as `_roi_response_neuropil` in the `SegmentationExtractor`.

# Current Behavior

Instead, in the `CaimanSegmentationExtractor`,
```
self._roi_response_neuropil = self._trace_extractor_read("C")
```
and `estimates.f` is ignored.

### Code of Conduct

alessandratrapani:
Denoised calcium traces should be distinguished from raw calcium traces.
See discussion in https://github.com/catalystneuro/roiextractors/issues/287#issuecomment-1956837952

This implementation should also fix #249.

TODO
- [x] Add background spatial components
- [x] Add raw traces as C+Yr
- [x] Add testing

### DIFF
--- a/src/roiextractors/extractors/caiman/caimansegmentationextractor.py
+++ b/src/roiextractors/extractors/caiman/caimansegmentationextractor.py
@@ -56,7 +56,8 @@ class CaimanSegmentationExtractor(SegmentationExtractor):
         self.file_path = file_path
         self._dataset_file = self._file_extractor_read()
         self._roi_response_dff = self._trace_extractor_read("F_dff")
-        self._roi_response_neuropil = self._trace_extractor_read("C")
+        self._roi_response_denoised = self._trace_extractor_read("C")
+        self._roi_response_neuropil = self._trace_extractor_read("b")
         self._roi_response_deconvolved = self._trace_extractor_read("S")
         self._image_correlation = self._summary_image_read()
         self._sampling_frequency = self._dataset_file["params"]["data"]["fr"][()]
@@ -177,8 +178,10 @@ class CaimanSegmentationExtractor(SegmentationExtractor):
             estimates = f.create_group("estimates")
             params = f.create_group("params")
             # adding to estimates:
+            if segmentation_object.get_traces(name="denoised") is not None:
+                estimates.create_dataset("C", data=segmentation_object.get_traces(name="denoised"))
             if segmentation_object.get_traces(name="neuropil") is not None:
-                estimates.create_dataset("C", data=segmentation_object.get_traces(name="neuropil"))
+                estimates.create_dataset("b", data=segmentation_object.get_traces(name="neuropil"))
             if segmentation_object.get_traces(name="dff") is not None:
                 estimates.create_dataset("F_dff", data=segmentation_object.get_traces(name="dff"))
             if segmentation_object.get_traces(name="deconvolved") is not None:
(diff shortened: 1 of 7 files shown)


src/roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -56,7 +56,8 @@ def __init__(self, file_path: PathType):
         self.file_path = file_path
         self._dataset_file = self._file_extractor_read()
         self._roi_response_dff = self._trace_extractor_read("F_dff")
-        self._roi_response_neuropil = self._trace_extractor_read("C")
+        self._roi_response_denoised = self._trace_extractor_read("C")
+        self._roi_response_neuropil = self._trace_extractor_read("b")
```
EricThomson, line 60: Hi I'm just in here poking around your PR 😄 It looks great 🚀 

I think you probably want `f` here for neuropil for the `trace_extractor()` method (that's the background fluctuations). `b` is the background *spatial* component(s) that would correspond for instance to `A` for the neural components). Forgive me if I've misunderstood how things work I'm pretty new to NWB. 

I'm thinking a typo (our variable names are awful, `b`/`f` what do those mean? 😆 ), as `f` is what you noted [in the original issue](https://github.com/catalystneuro/roiextractors/issues/287#issuecomment-1956276420).

alessandratrapani: Thank you so much @EricThomson. I am now trying include the spatial component "b" as well.

pauladkisson:
For neuroconv [#783](https://github.com/catalystneuro/neuroconv/pull/783), I need
- [ ] `get_background_ids()` following the example of `get_roi_ids`
- [ ] `get_background_image_masks()` following the example of `get_roi_image_masks()`
- [ ] `get_background_pixel_masks()` following the example of `get_roi_pixels_masks()`

pauladkisson (commented):
Looks good on first viewing -- I'll try it out in neuroconv and see if I run into any issues.

pauladkisson:
I noticed that CaimanSegmentationExtractor doesn't have any dedicated unit tests -- maybe this would be a good time to put that together?

I'm still pretty torn about whether we should go for a rolling refactor or an intensive refactor for the testing. @CodyCBakerPhD maybe you can weigh in?

Also, perhaps it is time to add a deprecation warning to the `write_segmentation` method since we are moving the api away from write support.

EricThomson:
Note away on vacation until end of next week but I should be able to start looking more then. Very exciting stuff!

CodyCBakerPhD:
> I noticed that CaimanSegmentationExtractor doesn't have any dedicated unit tests -- maybe this would be a good time to put that together?

Looks like it's a part of the general roundtrip tests: https://github.com/catalystneuro/roiextractors/blob/main/tests/test_io.py

From that perspective, dedicated tests would only be for specific assertions of specific structures unique to Caiman that can't be easily included as params in the generic testing framework

> I'm still pretty torn about whether we should go for a rolling refactor or an intensive refactor for the testing. @CodyCBakerPhD maybe you can weigh in?

Rolling in small iterative steps is always appreciated for review - for tests as well, there's nothing wrong with keeping the current infrastructure, adding a new one, then remove the old one after the new one is completely done

I'd be in favor of a refactor to something akin to the [mixin style of NeuroConv interface tests](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/testing/data_interface_mixins.py#L34), which allow you to define format-specific tests pretty easily

> Also, perhaps it is time to add a deprecation warning to the write_segmentation method since we are moving the api away from write support.

Yeah, I share the philosophy of SpikeInterface that supporting format-specific writing methods is beyond our scope in this package

pauladkisson (approved):
Works well in neuroconv, I think this is good to go.

PR #310, opened 2024-04-16
Fix Cron Bug with Assess-file-changes

pauladkisson:
Based on the failed daily actions, it looks like the assess-file-changes.yml shows up as changed --> excluding it from the checked files should prevent daily test failures.

Also excluded setup.py so that Update dev version workflow doesn't fail (see #311)


CodyCBakerPhD:
We can try

It's mostly about how the assessment of file changes shouldn't even trigger on a daily run (since not relevant)

CodyCBakerPhD (approved):

PR #317, opened 2024-04-29
Update Workflows

pauladkisson:
- [x] remove conda
- [x] x64
- [x] arm64
- [x] python 3.12
- [x] dev tests

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11]
+        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
         os: [ubuntu-latest, windows-latest, macos-13]
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -38,9 +38,8 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.8, 3.9, "3.10", 3.11]
-        os: [ubuntu-latest, windows-latest, macos-12]
+        os: [ubuntu-latest, windows-latest, macos-13]
     steps:
-      - uses: s-weigand/setup-conda@v1
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow --tags
       - name: Setup Python ${{ matrix.python-version }}
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         os: [ubuntu-latest, windows-latest, macos-13]
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -78,7 +78,7 @@ jobs:
         run: pytest -n auto --dist loadscope --cov=./ --cov-report xml:./codecov.xml
       - if: ${{ matrix.python-version == '3.9' && matrix.os == 'ubuntu-latest'}}
         name: Upload full coverage to Codecov
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v4
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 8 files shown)


.github/workflows/run-tests.yml
```diff
@@ -37,10 +37,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10", 3.11]
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: [3.8, 3.9, "3.10", 3.11, 3.12]
```
CodyCBakerPhD, line 40: It's petty, but what about having all these python versions in quotes so they look consistent?

pauladkisson:
Looks like [scanimage-tiff-reader](https://pypi.org/project/scanimage-tiff-reader/) doesn't support arm...I guess we wait for them?

CodyCBakerPhD:
> Looks like [scanimage-tiff-reader](https://pypi.org/project/scanimage-tiff-reader/) doesn't support arm...I guess we wait for them?

Can you ping them about that?

Also you can just add a skip for that on relevant tests, `platform.architecture` + `platform.platform` or `sys.platform` should give you something to use to distnguish skip logic

pauladkisson:
> > Looks like [scanimage-tiff-reader](https://pypi.org/project/scanimage-tiff-reader/) doesn't support arm...I guess we wait for them?
> 
> Can you ping them about that?

Done: https://gitlab.com/vidriotech/scanimagetiffreader-python/-/issues/31

.github/workflows/run-tests.yml
```diff
@@ -75,7 +74,16 @@ jobs:
...
           key: ophys-datasets-042023-${{ matrix.os }}-${{ steps.ophys.outputs.HASH_OPHYS_DATASET }}
 
-      - name: Run full pytest with coverage
+      - name: Run full pytest (macos-latest)
+        if: ${{ matrix.os == 'macos-latest' }}
+        run: |
+          pytest -n auto --dist loadscope \
+            --ignore=tests/test_scan_image_tiff.py \
+            --ignore=tests/test_scanimage_utils.py \
+            --ignore=tests/test_scanimagetiffimagingextractor.py \
```
CodyCBakerPhD, line 83: Were you not able to get a skip to work well in the test suite itself? (like `pytest.mark.skipif`)?

tests/test_scan_image_tiff.py
```diff
@@ -5,7 +5,12 @@
 
 from hdmf.testing import TestCase
 from numpy.testing import assert_array_equal
-from ScanImageTiffReader import ScanImageTiffReader
+import platform
+
+if (
+    platform.processor() != "arm"
```
CodyCBakerPhD, line 11: Are you sure it's ALL arm that they don't support? I thought it was just mac + arm

pauladkisson: I don't have a windows-arm machine handy, so I have no idea if it's mac + arm or just arm.  But for the sake of CI, does it really matter?

.github/workflows/run-tests.yml
pauladkisson: The cli-level ignore solution is more dry.  I am not particularly interested in adding pytest.mark.skipif to every test in tests/test_scanimagetiffimagingextractor.py (and then removing them if/when scanimage adds m1 mac support).

CodyCBakerPhD: It's just as many lines to type and remove, and is easier to catch from source code in tests than to notice the split in CI

tests/test_scan_image_tiff.py
CodyCBakerPhD: Well, if you wanted to specify JUST "CI" then you'd also fetch the CI environment variable from GitHub actions. The logic you have here affects every platform and device that might want to run the test suite, whereas we want a pretty narrow exclusion here

pauladkisson:
wait, I'm in the middle of this fix too

CodyCBakerPhD:
> @pauladkisson
[lowercase darwin](https://github.com/catalystneuro/roiextractors/pull/317/commits/a7c63278cf7688110724e352e528e8d4620fa38a)

I believe `platform.system()` is upper-case, `sys.platform` is lower case

You'll just also need a conditional add for https://github.com/catalystneuro/roiextractors/blob/workflow/tests/test_io.py#L48 to the list of IO to test

pauladkisson:
> You'll just also need a conditional add for https://github.com/catalystneuro/roiextractors/blob/workflow/tests/test_io.py#L48 to the list of IO to test

Ah, right, forgot about that one.

CodyCBakerPhD:
Looks great, thanks!

CodyCBakerPhD (approved):

PR #318, opened 2024-04-29
Update-Testing-Data fix

pauladkisson:
Looks like the caching doesn't work properly on the workflow branch, so will need to merge this in before I can properly test the testing workflow.


pauladkisson:
Tests fail bc they are still broken, but this workflow runs properly -- see [action](https://github.com/catalystneuro/roiextractors/actions/runs/8884036257)

CodyCBakerPhD (approved):

PR #324, opened 2024-05-10
Update testing.py types

bendichter:


src/roiextractors/testing.py
```diff
@@ -52,9 +52,9 @@ def generate_dummy_imaging_extractor(
     num_rows: int = 10,
     num_columns: int = 10,
     num_channels: int = 1,
-    sampling_frequency: float = 30,
+    sampling_frequency: float = 30.0,
     dtype: DtypeType = "uint16",
-    channel_names: Optional[list] = None,
+    channel_names: Optional[List[str]] = None,
```
h-mayorquin, line 57: FYI, python 3.9 already supports using the constructor as typing
https://peps.python.org/pep-0585/

And the new documentation marks the use of `Typing.List` as deprecated:
https://docs.python.org/3/library/typing.html#typing.List

And favors using the generic constructors.

Tbh, I kind of prefer the disctinction between list and List but it seems that the community moved in the other direction.

bendichter: ok this should be updated once we drop 3.8

h-mayorquin (approved):

PR #326, opened 2024-05-13
fix inscopix dtype

bendichter:
Apparently `np.uint16` is different from `np.dtype(np.uint16)`

### DIFF
--- a/src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py
+++ b/src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py
@@ -53,5 +53,5 @@ class InscopixImagingExtractor(ImagingExtractor):
         end_frame = end_frame or self.get_num_frames()
         return np.array([self.movie.get_frame_data(i) for i in range(start_frame, end_frame)])
 
-    def get_dtype(self):
+    def get_dtype(self) -> np.dtype:
         return np.dtype(self.movie.data_type)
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Specifically seems that if you ever see `numpy.uint16` it's from taking the `type` instead of the `.dtype` of an array element or numpy scalar;

```python
array = np.array([1,2,3], dtype="uint16")
type(array[0])
> np.uint16

array[0].dtype
> dtype('uint16')
```

CodyCBakerPhD:
Is there a test to add or fix (doesn't look like changing this line in source altered the outcome of a test?) for this

bendichter:
Unfortunately,

```python
np.uint16 == np.dtype(np.uint16)
```

```python
True
```

however

```python
np.uint16.itemsize()
```

```python
TypeError: 'getset_descriptor' object is not callable
```

while

```python
np.dtype(np.uint16).itemsize
```

```python
2
```

I suppose I could test asking for the itemsize

CodyCBakerPhD:
> Unfortunately,
> 
> np.uint16 == np.dtype(np.uint16)

I believe this might be intended since equality is somewhat looser than saying each side of the expression should be of the same object type (with same methods/attributes, etc)

I think the correct assertion would instead be

```python
assert extractor.get_dtype() is np.dtype("uint16")
```

bendichter:
@CodyCBakerPhD good point. I changed it to `is` and it works. You are right, `is` should be more restrictive

src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py
```diff
@@ -54,4 +54,4 @@ def get_video(
         return np.array([self.movie.get_frame_data(i) for i in range(start_frame, end_frame)])
 
     def get_dtype(self):
```
CodyCBakerPhD, line 56:
```suggestion
    def get_dtype(self) -> np.dtype:
```

For good measure

CodyCBakerPhD (approved):

PR #327, opened 2024-05-15
BaseSegmentation class improvements

CodyCBakerPhD:
Writing a custom segmentation extractor, noticed a couple of things

i) `self._image_mask` is used in later methods but never specific in the base `__init__`; not a huge deal, but figured I'd include it there to at least demonstrate how to include it during file read

ii) moved an `abstractmethod` towards the top of the class to make it easier to spot (since it needs to be defined in any child)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,21 +2,27 @@
 
 ### Back-compatibility break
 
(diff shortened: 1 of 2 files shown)


pauladkisson (approved):
lgtm

PR #330, opened 2024-05-18
suite2p require essential files

### LINKED ISSUE #329: suite2p allowing for options=None
https://github.com/catalystneuro/roiextractors/blob/22ccdad19697284efa5b39db09d72e9d598eac81/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py#L153-L157

I'm not really sure what the intention of these lines is. `self.options = options.item() if options is not None else options` will set `self.options` to `None` if `options is None`. But then if `None` is set as the value, the next line, `self._sampling_frequency = self.options["fs"]` will fail because you are trying to use `getattr` on a `None`.

bendichter:
fix #329

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,22 +1,27 @@
-# Upcoming (v0.6.0)
+# Upcoming (v0.5.9)
 
-### Back-compatibility break
+### Deprecations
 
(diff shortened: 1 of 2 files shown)


weiglszonja (dismissed):
Thank you @bendichter for this fix, looks good.

CodyCBakerPhD (approved):

PR #335, opened 2024-05-18
unshadow dict

bendichter:


src/roiextractors/extraction_tools.py
```diff
@@ -593,15 +593,15 @@ def animate_func(i, imaging, im, ax):
     return anim
 
 
-def check_keys(dict):
+def check_keys(dict_) -> dict:
```
CodyCBakerPhD, line 596:
```suggestion
def check_keys(dict_: dict) -> dict:
```

This makes the conflation against the builtin a bit clearer

CodyCBakerPhD: (already in docstring)

PR #337, opened 2024-05-22
API docs

### LINKED ISSUE #184: [Documentation]: Add API docs
### What would you like changed or added to the documentation and why?

Add auto-generated API docs

### Do you have any interest in helping write or edit the documentation?

No.

### Code of Conduct

pauladkisson:
Fixes #184

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -5,4 +5,4 @@ API documentation
    :maxdepth: 0
    :glob:
 
-   build_api/*
+   my_build_api/*
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -31,5 +31,6 @@ With this package, a user can:
    usage
    usecase
    contribute
+   my_api/index
    licence
    contact
--- a/docs/source/my_api/base_imagingextractors.rst
+++ b/docs/source/my_api/base_imagingextractors.rst
@@ -1,0 +1,7 @@
+Base ImagingExtractors
+-------------------------------------------------
+.. automodule:: roiextractors.imagingextractor
+
+.. automodule:: roiextractors.multiimagingextractor
+
+.. automodule:: roiextractors.volumetricimagingextractor
--- a/docs/source/my_api/base_segmentationextractors.rst
+++ b/docs/source/my_api/base_segmentationextractors.rst
@@ -1,0 +1,5 @@
+Base SegmentationExtractors
+-------------------------------------------------
+.. automodule:: roiextractors.segmentationextractor
+
+.. automodule:: roiextractors.multisegmentationextractor
--- a/docs/source/my_api/index.rst
+++ b/docs/source/my_api/index.rst
@@ -1,0 +1,8 @@
+API documentation
+=================
+
+.. toctree::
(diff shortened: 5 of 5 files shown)


src/roiextractors/example_datasets/__init__.py
```diff
@@ -1,14 +1 @@
-"""Toy example ImagingExtractor and SegmentationExtractor for testing.
```
CodyCBakerPhD, line 1: Why go to all the effort on this PR for removing all of these? Now if I ask `roiextractors.example_datasets?` and the like I'm back to `<no docstring>`

CodyCBakerPhD:
https://github.com/catalystneuro/roiextractors/actions/runs/9198701854/job/25301972879?pr=337

src/roiextractors/example_datasets/__init__.py
pauladkisson: I removed them bc a lot of times they were a more or less exact replication of the class docstrings, so I wanted to remove extra maintenance.  I also wanted to remove the Classes section of some of the module docstrings bc it makes the api docs less intuitive. 

But I can roll this back and just keep 1-liner docstrings for every module since it helps the navigability in interactive python settings.

CodyCBakerPhD: > I removed them bc a lot of times they were a more or less exact replication of the class docstrings

Easy solution there is to make them more informative; e.g., describing module contents, how to import and use contents, etc.

SKlearn import style (separation into private vs. public modules) would also reduce the number of publically exposed modules that deserve such top-level documentation



> so I wanted to remove extra maintenance

But taking the time to remove them all is already more maintenance than you've had to do in many months on this 😆


> I also wanted to remove the Classes section of some of the module docstrings bc it makes the api docs less intuitive.

Surely a configurable setting for `automodule` to limit rendering of headers? Can always just use `autoclass` as well since it's usually one module per class

CodyCBakerPhD: IMO even if you want to change them I'd do it in a follow-up PR, not here

Goal of this PR is to just get something up and running for API docs for what we currently have, right?

CodyCBakerPhD:
Huh, any clue as to why Inscopix isn't rendering anything? https://roiextractors--337.org.readthedocs.build/en/337/api/imaging_extractors/inscopiximagingextractor.html

CodyCBakerPhD:
That's the only one with an issue - otherwise everything else looks fine

pauladkisson:
> Huh, any clue as to why Inscopix isn't rendering anything? https://roiextractors--337.org.readthedocs.build/en/337/api/imaging_extractors/inscopiximagingextractor.html

Renders fine locally...
so thats a little strange...

CodyCBakerPhD:
@pauladkisson Another thing to look into that I just noticed, in addition to Inscopix on RTD - the API docs seem to be grabbing and displaying private methods - but do we really want that behavior?  Sphinx is supposed to hide those by default apparently (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-members)

Issue #341, opened 2024-06-05
[Bug]: Incorrect bruker metadata parsing for single plane imaging


mhturner:
### What happened?

Using the Bruker->NWB converter in neuroconv (as [here](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/imaging/brukertiff.html#)), I noticed that my single-plane Bruker imaging data was being flagged as multiplane.

Tracked this down to `brukertiffimagingextractor._determine_imaging_is_volumetric()`, which assumes that the metadata flag 'zDevice' == 1 means this is a volumetric scan, which is a sensible thing to assume. However my single plane image data as well as my multiplane image data both have this value as 1.

In my experience, a reliable way to tell what the scan type is from the .xml metadata is the Sequence type. This can be things like:
'TSeries Timed Element' - XYT image data
'TSeries ZSeries Element' - XYZT image data
'ZSeries' - XYZ data (not a time series)
'Single' - XY image

and probably other things too that I don't acquire.

### Steps to Reproduce

```python
from roiextractors.extractors.tiffimagingextractors.brukertiffimagingextractor import _determine_imaging_is_volumetric

print(_determine_imaging_is_volumetric(bruker_file_dir))
```

### Traceback

```python
True
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.10

### Code of Conduct

Yes

### Duplicated Issue Check

No

CodyCBakerPhD:
Thanks for finding this and including all the details!

attn @weiglszonja @alessandratrapani @pauladkisson

weiglszonja:
Thank you for opening this issue @mhturner! I'm thinking that since it is not obvious to decide whether it is a volumetric type, how about we just don't assume and separate the classes to single and multiplane, and then the user decides what is the more appropriate extractor/interface for the data. I'm thinking a refactor similar to ScanImage. What do you guys think? @pauladkisson @alessandratrapani @CodyCBakerPhD

h-mayorquin:
@mhturner would you be so kind to share the corresponding xml of your data? 

The volumetric data that I have seen has a specification that looks like this:

```xml
      <PVStateShard>
        <PVStateValue key="framePeriod" value="0.048474237" />
        <PVStateValue key="laserPower">
          <IndexedValue index="0" value="650.390625" description="920 Axon" />
          <IndexedValue index="1" value="0" description="1064 Axon" />
          <IndexedValue index="2" value="0" description="Monaco" />
[...]

PR #343, opened 2024-06-10
Fix bug in `BrukerTiffSinglePlaneImagingExtractor`

h-mayorquin:
The current code assumes that the channel name (the stream) should be in the filename. This is not the case for some data that I have. Here is a frame element in the correspoding xml

```xml
    <Frame relativeTime="0.00949076" absoluteTime="41.2114907600011" index="3" parameterSet="CurrentSettings">
      <File channel="1" channelName="Red" filename="TSeries-20240527-001_Cycle00001_Ch1_000003.ome.tif" />
      <File channel="2" channelName="Green" filename="TSeries-20240527-001_Cycle00001_Ch2_000003.ome.tif" />
```

As you can see the current asumption is not valid and therefore the code is a bug. The fix proposed in this PR yses the channelName in the xml element to filter the file paths that correspond to the channel.


h-mayorquin:
Ah, OK, this is a case for a problem when using a single argument to specify plane and channel. 

Here:
https://github.com/catalystneuro/roiextractors/blob/10aea18c8abe2e70fb9ed163ba435293f66fde5a/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py#L226-L230

We pass plane_stream (a string identifying the plane) as a stream to `BrukerTiffSinglePlaneImagingExtractor`.  But in some other places we pass a channel. The current implementation is relying on the fact that both are included in the filename and that this code will work. However, as I argued above the channelName is not always in the filename. 

I think at least at this level we should seprate them, it would make the implementation easier.

weiglszonja:
Thank you @h-mayorquin for looking into this issue. I still think that (following [this issue](https://github.com/catalystneuro/roiextractors/issues/341#issuecomment-2159359023)):

> We refactor the code so the classes are clearly separated as you guys propose.

would be the way to go, it would follow how @pauladkisson refactored the extractor for ScanImage, I believe we don't have a routing class there either. We didn't know it was a bug because the examples that we've seen for Bruker always followed the convention we built on, I think it would reasonable to have designated classes instead of a router.

h-mayorquin:
Ok, @weiglszonja I found a way of solving this without requiring more changes. I think that the changes we have discussed are more robust and would make the code clear but this should be a intermediate fix. Note that this requires https://github.com/catalystneuro/roiextractors/pull/344 before merging.

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -372,25 +372,33 @@ def __init__(self, folder_path: PathType, stream_name: Optional[str] = None):
...
+                f.attrib["filename"] for f in file_elements if f.attrib["channelName"] == stream_name
+            ]
+
+        else:  # This is the case for when stream_name is a plane_stream
+            file_names = [file.attrib["filename"] for file in file_elements]
+            file_names_for_stream = [file for file in file_names if self.stream_name in file]
+            if file_names_for_stream == []:
+                raise ValueError(
+                    f"The selected stream '{self.stream_name}' is not in the available channel_streams '{streams['channel_streams']}'!"
+                )
```
weiglszonja, line 401: Ah I see, this can work.

h-mayorquin: Yes, this is the string ducktyping (changing the workflow depending on the value/shape/content of the string) that I did not want to do but at least it solves the problem. I think it would be better if we had another argument plane_stream or just pass file_paths somewhere as discussed. But I think that discussion is more dififcult and this already solves the bug so...

weiglszonja: I agree with you, we can revisit this once we have the bandwidth to do the `file_paths` approach.

weiglszonja: Feel free to merge #346 into this once finished and let me know when this is ready for final review.

h-mayorquin: This is ready for final review. 

#346 is an optimization on top of this but this PR is the bug fix. 

I would prefer to merge this first and then add the optimization as another PR if you are OK with that.

weiglszonja (approved):

PR #346, opened 2024-06-12
Use lxml for Bruker xml parsing

h-mayorquin:
I am having troubles testing if the recent changes in https://github.com/catalystneuro/roiextractors/pull/343 work well with the data that I got from Clandinin.  The problem is that the data takes too long too load (around 20 minutes). I profiled and the bottleneck is xml parsing. 

This PR uses xpath from lxml instead of the standard parser from the library. **There is a speed up of around ~100 times**

The firs run in the screenshot is using this PR and the other one is using the base branch here:

image


h-mayorquin:
Note, adapt get stream to use lxml

h-mayorquin:
This is ready to go, I also changed is_volumetric to use lxml

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -414,8 +435,8 @@ def __init__(self, folder_path: PathType, stream_name: Optional[str] = None):
         self._sampling_frequency = frame_rate
         self._channel_names = [self.stream_name.split("_")[0]]
 
-        # count the number of occurrences of each file path and their names
-        # files that contain stacks of images (multi-page tiffs) will appear repeated (number of repetition is the number of frames in the tif file)
+        # # count the number of occurrences of each file path and their names
+        # # files that contain stacks of images (multi-page tiffs) will appear repeated (number of repetition is the number of frames in the tif file)
```
weiglszonja, line 439:
```suggestion
        # count the number of occurrences of each file path and their names
        # files that contain stacks of images (multi-page tiffs) will appear repeated (number of repetition is the number of frames in the tif file)
```

weiglszonja (approved):

PR #348, opened 2024-06-13
Remove wheel requirement and remove testing dependencies from the CI.

h-mayorquin:
Three things in the PR:

* The latest version of scan image supports modern python build systems so installing wheel is no longer necessary:
https://gitlab.com/vidriotech/scanimagetiffreader-python/-/merge_requests/2/diffs
* Moving the dependencies to testing instead of having the implicit in the CI. Companion to https://github.com/catalystneuro/neuroconv/pull/913
* Remove the editable installation because that mode is unstable in setuptools https://setuptools.pypa.io/en/latest/userguide/development_mode.html

### DIFF
--- a/.github/workflows/dev-testing.yaml
+++ b/.github/workflows/dev-testing.yaml
@@ -22,15 +22,12 @@ jobs:
       - name: Global Setup
         run: |
             pip install -U pip
-            pip install pytest-xdist
             git config --global user.email "CI@example.com"
             git config --global user.name "CI Almighty"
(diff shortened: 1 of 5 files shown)


h-mayorquin:
Also related to the tests added in https://github.com/catalystneuro/roiextractors/pull/347#pullrequestreview-2117028131

pauladkisson (approved):
LGTM. Thanks @h-mayorquin!

PR #349, opened 2024-06-13
Bye bye spikeextractors

h-mayorquin:
Been meaning to this for a while @CodyCBakerPhD  : )

We can still avoid this dependency if we just define an in-house function that gets spike trains which should not be tood hard. But I really don't have time right now. So I am doing this just because at some point spikeextractors will stop working so I am doing it now before I forget.


pauladkisson (approved):
LGTM

PR #357, opened 2024-09-11
updated requirements to include cv2 and update dev testing locations for neuroconv

pauladkisson:
Fixes failing dailies

Fixes failing dev tests

### DIFF
--- a/requirements-full.txt
+++ b/requirements-full.txt
@@ -1,5 +1,8 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader>=1.4.1.4
 neuroconv[video]>=0.4.6  # Uses the VideoCaptureContext class
+opencv-python-headless>=4.5.1.48
+opencv-python-headless>=4.5.1.48,<4.7.0.72; sys_platform == 'darwin'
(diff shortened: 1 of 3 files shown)


requirements-full.txt
```diff
@@ -1,5 +1,8 @@
 tifffile>=2018.10.18
 scanimage-tiff-reader>=1.4.1.4
 neuroconv[video]>=0.4.6  # Uses the VideoCaptureContext class
+opencv-python-headless>=4.5.1.48
```
h-mayorquin, line 4: Isn't this included in neuroconv video?

It might be a good idea to make it explicit though, do other extractors require it?

h-mayorquin (approved):
LGTM

requirements-full.txt
pauladkisson: I thought it would've been, but the tests were failing bc opencv wasn't installed.

h-mayorquin: I would be concerned if this is related to the recent changes to pyproject.toml, when did the test start to fail?

h-mayorquin: Ok, then it sounds unlikely. Still strange. Probably not worth finding out if this is working.

PR #368, opened 2024-10-07
Add Minian segmentation extractor

alessandratrapani:
Add segmentation extractor for Minian output.
Supported output format: Zarr (for now)
Similarly to Caiman uses CNMF to perform cell identification.

Output structure (example):

    minian/
     ├── A.zarr 
     │   ├── A (851, 608, 608) float64
     │   ├── animal () <U7
     │   ├── height (608,) int64
     │   ├── session_id () <U9
     │   ├── unit_id (851,) int64
     │   └── width (608,) int64
     ├── C.zarr
     │   ├── C (851, 8931) float64
     │   ├── animal () <U7
     │   ├── frame (8931,) int64
     │   ├── session () <U14
     │   └── session_id () <U9
     │   └── unit_id (851,) int64
     ├── b.zarr
     │   ├── animal () <U7
     │   ├── b (608, 608) float64
     │   ├── height (608,) int64
     │   ├── session () <U14
     │   ├── session_id () <U9
     │   ├── unit_id () int64
     │   └── width (608,) int64
     ├── b0.zarr
     │   ├── animal () <U7
     │   ├── b0 (851, 8931) float64
     │   ├── frame (8931,) int64
     │   ├── session () <U14
     │   ├── session_id () <U9
     │   └── unit_id (851,) int64
     ├── c0.zarr
     │   ├── animal () <U7
     │   ├── c0 (851, 8931) float64
     │   ├── frame (8931,) int64
     │   └── session () <U14
     ├── f.zarr
     │   ├── animal () <U7
     │   ├── f (8931,) float64
     │   └── frame (8931,) int64
     ├── S.zarr
     │   ├── S (851, 8931) float64
     │   ├── animal () <U7
     │   ├── frame (8931,) int64
     │   ├── session () <U14
     │   └── session_id () <U9
     │   └── unit_id (851,) int64
     ├── max_proj.zarr
     │   ├── animal () <U7
     │   ├── max_proj (608, 608) float64
     │   ├── height (608,) int64
     │   ├── session () <U14
     │   ├── session_id () <U9
     │   └── width (608,) int64

- A: Spatial footprints of cells. Should have dimensions (“unit_id”, “height”, “width”). --> `image_masks`
- C: Temporal components of cells. Should have dimensions “frame” and “unit_id”. --> `roi_response_denoised`
- b: Spatial footprint of background. Should have dimensions (“height”, “width”). --> `background_image_masks`
- f: Temporal dynamic of background. Should have dimension “frame”. --> `roi_response_neuropil`
- b0: Baseline fluorescence for each cell. Should have dimensions (“frame”, “unit_id”) and same shape as C --> `roi_response_baseline`
- c0: Initial calcium decay, in theory triggered by calcium events happened before the recording starts. Should have dimensions (“frame”, “unit_id”) and same shape as C
- S: Deconvolved spikes for each cell. Should have dimensions (“frame”, “unit_id”) and same shape as C  --> `roi_response_deconvolved`
- max_proj: the maximum projection --> `summary_image`

TODOs
- [x] add extractor
- [x] update CHANGELOG
- [x] implement tests


h-mayorquin:
Do tag me when this is ready for review.

alessandratrapani:
I am uploading on the s3 bucket the testing folder for the minian segmentation data

alessandratrapani:
@h-mayorquin ready for review

h-mayorquin (commented):
I took the first look. Looks good and simple to me.

1) Is the structured documented?  (the one that you shared on the PR) I am inferring the meaning of A, C, b, f, etc from the equation of CNMF here:

https://minian.readthedocs.io/en/stable/pipeline/notebook_5.html

But maybe there is other place in the documentation where this is stated more precisely, if so, we probably should link that in the docstring.

2) The example that you uploaded, can you add a readme and share it with me so that I can upload it to gin.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,189 @@
...
+        Parameters
+        ----------
+        folder_path: str
+            The location of the folder containing minian .zarr output.
+        """
+        SegmentationExtractor.__init__(self)
+        self.folder_path = folder_path
+        self._roi_response_denoised = self._trace_extractor_read(field="C")
+        self._roi_response_baseline = self._trace_extractor_read(field="b0")
+        self._roi_response_neuropil = self._trace_extractor_read(field="f")
```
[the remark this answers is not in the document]
alessandratrapani: Let's say first that the background can contain neuropil signal. The problem is that I wanted to map this "f" signal with something already in the SegmentationExtractor.

Now, I am not super happy with this formulation I think we should have two separate variables:

1. _roi_response_neuropil with the respective _neuropil_images_masks. This can be used to store the actual neuropil signals that sometimes refer to the area surrounding the ROIs
2. _roi_response_background with the respective _background_images_masks. This can be used more generally to store the signal in the field of view that excludes the areas in the identified ROIs (as in this case).

I will open an issue on roiextractor to discuss this with the others.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,189 @@
...
+        image_masks: numpy.ndarray
+            The image masks for each background components.
+        """
+        dataset = self._file_extractor_read("/b.zarr")
+        if dataset is None or "b" not in dataset:
+            return None
+        else:
+            return np.expand_dims(dataset["b"], axis=2)
+
+    def _trace_extractor_read(self, field):
```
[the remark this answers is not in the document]
alessandratrapani: Yeah sure, I can change it to `read_trace_from_zarr_field`, I was just mirroring the CaimanSegmentationExtractor

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,189 @@
...
+import pandas as pd
+
+from ...extraction_tools import PathType
+from ...segmentationextractor import SegmentationExtractor
+
+
+class MinianSegmentationExtractor(SegmentationExtractor):
+    """A SegmentationExtractor for Minian.
+
+    This class inherits from the SegmentationExtractor class, having all
```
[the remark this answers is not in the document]
alessandratrapani: Ok, sure! I just wanted to be consistent with all the other extractors, but I guess a more detailed docstring won't hurt.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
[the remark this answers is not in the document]
h-mayorquin: OK, makes sense.

h-mayorquin:
What's the status of this?

tests/test_miniansegmentationextractor.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_image_size(self):
+        self.assertEqual(self.extractor.get_image_size(), self.image_size)
+
+    def test_num_frames(self):
+        self.assertEqual(self.extractor.get_num_frames(), self.num_frames)
+
+    def test_frame_to_time(self):
+        self.assertEqual(self.extractor.frame_to_time(frames=[0]), 0.328)
+
+    def test_num_channels(self):
```
h-mayorquin, line 91: This method is deprecated maybe it should not be tested? Or maybe it is not deprecated on segmentation extractors? do we have multi channel segmentation extractors?

tests/test_miniansegmentationextractor.py
```diff
@@ -0,0 +1,114 @@
...
+                shutil.copy(src, dst)
+
+        extractor = MinianSegmentationExtractor(folder_path=self.test_dir)
+        traces_dict = extractor.get_traces_dict()
+        self.assertEqual(traces_dict["deconvolved"], None)
+
+    def test_image_size(self):
+        self.assertEqual(self.extractor.get_image_size(), self.image_size)
+
+    def test_num_frames(self):
```
h-mayorquin, line 85: same, did we not change to image or frame shape and num samples instead of frames and size?

tests/test_miniansegmentationextractor.py
```diff
@@ -0,0 +1,114 @@
...
+            "b0.zarr",
+            "b.zarr",
+            "f.zarr",
+            "max_proj.zarr",
+            ".zgroup",
+            "timeStamps.csv",
+        ]
+        self.test_dir.mkdir(exist_ok=True)
+
+        for folder in folders_to_copy:
```
h-mayorquin, line 70: To avoid all of this tear down logic I suggest using pytest and tmp_path fixtures as they work reliable cross OS:

https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/test_miniansegmentationextractor.py
```diff
@@ -0,0 +1,114 @@
...
+        dataset = zarr.open(folder_path + "/b0.zarr")
+        cls.baseline_traces = np.transpose(dataset["b0"])
+        # neuropil trace
+        dataset = zarr.open(folder_path + "/f.zarr")
+        cls.neuropil_trace = np.expand_dims(dataset["f"], axis=1)
+
+        # ROIs masks
+        dataset = zarr.open(folder_path + "/A.zarr")
+        cls.image_masks = np.transpose(dataset["A"], (1, 2, 0))
+        cls.image_size = (dataset["height"].shape[0], dataset["width"].shape[0])
```
h-mayorquin, line 42: I think in this and some tests below you compare that the internal representation of the extractor is equal to the API.

I think we should compare the API classes to fixed values.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,210 @@
...
+
+    Parameters
+    ----------
+    folder_path: str
+        Path to the folder containing Minian .zarr output files.
+
+    """
+
+    extractor_name = "MinianSegmentation"
+    is_writable = True
```
h-mayorquin, line 54: I don't know is_writable and the mode are used anymore, they are not needed.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,210 @@
...
+            List of rejected ROI ids.
+        """
+        return list()
+
+    def get_traces_dict(self) -> dict:
+        """Get traces as a dictionary with key as the name of the ROiResponseSeries.
+
+        Returns
+        -------
+        _roi_response_dict: dict
```
h-mayorquin, line 186: why leading _ on the return argument?

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,210 @@
...
+        dataset = self._read_zarr_group(f"/{field}.zarr")
+
+        if dataset is None or field not in dataset:
+            return None
+        elif dataset[field].ndim == 2:
+            return np.transpose(dataset[field])
+        elif dataset[field].ndim == 1:
+            return np.expand_dims(dataset[field], axis=1)
+
+    def _read_timestamps_from_csv(self):
```
h-mayorquin, line 140: How are we or could we test this?

h-mayorquin (changes requested):
I would like to get the data on Gin aligned with the bucket so we can move forward with this. Let's pin the PR once we have it.

Two main points:

The LLMs are heavily biased toward managing I/O gracefully. When they don't find a value, they return None and keep running (as you can see in the functions used to read data in your extractor). But is this the behavior we want? Should we proceed with the extraction if some values are missing? Should we throw an error in some cases or at least warn the user? Since we're focused on metadata and data provenance, I believe certain combinations of missing data should either trigger an error or be clearly communicated to the user with a warning. You know this format and field better than I but I think this is something to take into account.

Is there any metadata we can extract here? The session start time is crucial, but is there anything else related to segmentation metadata that is preserved?

src/roiextractors/extractors/minian/miniansegmentationextractor.py
alessandratrapani: added a test that checks th first timestamps match the expected value from the csv

tests/test_miniansegmentationextractor.py
alessandratrapani: right

tests/test_miniansegmentationextractor.py
alessandratrapani: Is it better like this? The test uses `unittest.TestCase`, pytest fixtures aren't compatible

alessandratrapani:
(side note)test are failing because I updated the test to work with the updated timeStamp.csv file in https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/30

alessandratrapani:
> The LLMs are heavily biased toward managing I/O gracefully. When they don't find a value, they return None and keep running (as you can see in the functions used to read data in your extractor). But is this the behavior we want? Should we proceed with the extraction if some values are missing? Should we throw an error in some cases or at least warn the user? Since we're focused on metadata and data provenance, I believe certain combinations of missing data should either trigger an error or be clearly communicated to the user with a warning. You know this format and field better than I but I think this is something to take into account.

Ok, I'll add warnings in the __init__

alessandratrapani:
> Is there any metadata we can extract here? The session start time is crucial, but is there anything else related to segmentation metadata that is preserved?

No, but I discussed with the authors the possibility of including the segmentation parameters in the NWBFile (maybe with an ad hoc extension). However, the minian software does not save these parameters currently.

tests/test_miniansegmentationextractor.py
h-mayorquin: I guess just change the naming here now that #443 is merged

h-mayorquin:
The data on gin is ready.

> Ok, I'll add warnings in the init

I was asking if there is a combination of input (or lack of) that is inacceptable. If the answer is no, and any combination is really possible then the warnings are not needed. 

> No, but I discussed with the authors the possibility of including the segmentation parameters in the NWBFile (maybe with an ad hoc extension). However, the minian software does not save these parameters currently.

I see, no `session_start_time` or anything else then.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
h-mayorquin: Can you add a cache so the timestamps are not read twice if somebody calls the function again?

h-mayorquin: Yes, as in, this should connect to `frame_to_times`, right? or how and where are we calling this?

tests/test_miniansegmentationextractor.py
```diff
@@ -0,0 +1,118 @@
...
+from numpy.testing import assert_array_equal
+
+from roiextractors import MinianSegmentationExtractor
+from tests.setup_paths import OPHYS_DATA_PATH
+
+
+class TestMinianSegmentationExtractor(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        folder_path = str(OPHYS_DATA_PATH / "segmentation_datasets" / "minian")
```
h-mayorquin, line 17: Hey, you have a different structure in the s3 bucket and in the gin file. It took me like half an hour to make sense on why the test was failing for me.

Guys, we should be careful, we are using a double repo structure to make it easier but then this is the type of error that slips @pauladkisson @weiglszonja

alessandratrapani: I fixed the s3 bucket repo

tests/test_miniansegmentationextractor.py
alessandratrapani: Sorry I misunbderstood the comment. Is it ok now?

alessandratrapani:
> Also, from the look of the zarr file that you shared it seems that there is some metadata that we can access from here and propagate to neuroconv? sesion_id, unit_id,, animal?

Yeah, I did not add them in the ad hoc extractor for Cai Lab because I preferred managing the session_id and subject_id in the conversion script, because there was an overlap between the two string: session_id = "Ca_EEG3-4" and animal = "Ca_EEG3" while the session_id should have been related to the specific task of the session as in the original dataset structure, e.g. FC (for Fear Conditioning), and the subject_id should have been "Ca_EEG3-4" that is stored in the session_id attribute in the zarr group (I know it is a bit confusing).
However, I guess I could add function to return this to the mininian interface and then let the user decide if it is worth to override them or not in the conversion script. What is your opinion?

> Can we get the ids from the zarr instead of determined implicitly and positionally the way that we are doing it here?

Nope. That was my first implementation, and it failed because the unit_id contains yes indexes, but they do not always index the traces actually stored in the C.zarr group. 
I'll try to explain: the unit_id might look like this [0,1,2,4], because the unit "3" has been discarded. However, the shape C dataset (in the C.zarr group) will be (n_times, 4), because the unit 3 temporal component is not stored. Same for the spatial component A. 
So it is better to use our get_roi_ids function. I could add the roi_name column we **recently** discussed.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,260 @@
...
+    def _read_background_image_mask_from_zarr(self):
+        """Read the image masks from the zarr output.
+
+        Returns
+        -------
+        image_masks: numpy.ndarray
+            The image masks for each background components.
+        """
+        dataset = self._read_zarr_group("/b.zarr")
+        if dataset is None or "b" not in dataset:
```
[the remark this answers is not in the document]
h-mayorquin: I am suggestion adding this:

```
 ├── b.zarr
 │   ├── animal () <U7
 │   ├── b (608, 608) float64
 │   ├── height (608,) int64
 │   ├── session () <U14
 │   ├── session_id () <U9
 │   ├── unit_id () int64
 │   └── width (608,) in
```

To the docstring.

This why, someone that does not know the format will understand why you are testing for "b" in the dataset

src/roiextractors/extractors/minian/miniansegmentationextractor.py
h-mayorquin: Hi, you marked this as resolved as well but I don't see changes.

h-mayorquin:
> However, I guess I could add function to return this to the mininian interface and then let the user decide if it is worth to override them or not in the conversion script. What is your opinion?

Yes, we will use this in neuroconv to fetch default metadata for the interface. If this is wrong, the user can always overwrite it. 

> I'll try to explain: the unit_id might look like this [0,1,2,4], because the unit "3" has been discarded. However, the shape C dataset (in the C.zarr group) will be (n_times, 4), because the unit 3 temporal component is not stored. Same for the spatial component A.

Why would this be a problem? The 4 response traces are there, it is just that their ids are not positional. or am I misunderstanding?

src/roiextractors/extractors/minian/miniansegmentationextractor.py
alessandratrapani: > Hi, you marked this as resolved as well but I don't see changes.

I mark this resolved when I added the test as the original request. Then you comment with a different request and I guess it stayed resolved.

> Yes, as in, this should connect to frame_to_times, right? or how and where are we calling this?
It is called in line 106:

alessandratrapani: Does this solution works?

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        if self._times is not None:
+            return self._times
+
+        csv_file = self.folder_path / "timeStamps.csv"
+        df = pd.read_csv(csv_file)
+        frame_numbers = self._read_zarr_group("/C.zarr/frame")
+        filtered_df = df[df["Frame Number"].isin(frame_numbers)] * 1e-3
+
+        return filtered_df["Time Stamp (ms)"].to_numpy()
```
alessandratrapani, line 203: multiply 1e-3 here for clarity

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,284 @@
...
+            )
+
+        if self._background_image_masks is None and self._roi_response_neuropil is not None:
+            warnings.warn(
+                "Background temporal component (f.zarr) is available but no background spatial component (b.zarr) is associated. "
+                "This means background temporal dynamics exist but without corresponding spatial masks.",
+                UserWarning,
+            )
+
+        self._times = self._read_timestamps_from_csv()
```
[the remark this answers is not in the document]
h-mayorquin: The discussion here is that we should not load the timestamps by default at extractor initialization because this makes the object slower to initialize and heavier.

On the other hand, But we should connect this with the default function to get the times,so this is called when someone uses `sample_indices_to_time` because that is the function that is called to the timestamps in neuroconv. 

I am writing this because I want to reference.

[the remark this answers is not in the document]
alessandratrapani: I went for get_original_timestamps() and sample_indexes_to_times() untill we agree on a unique solution.

alessandratrapani:
Ok, I hope I didn't forget anything, what I did with the last commits:

- [ ] remove self._times and add get_original_timestamps() and sample_indices_to_time()
- [ ] add get_roi_ids() to read the unit_id list directly
- [ ] add methods to retrieve session and subject ids
- [ ] replace warnings with value errors

alessandratrapani:
I think tests are failing because the get_roi_ids() return type can be string, but we haven't implemented yet (https://github.com/catalystneuro/neuroconv/pull/1390) to accept string type in neuroconv, correct?

h-mayorquin:
OK, let me know this is ready for another review.

alessandratrapani:
> OK, let me know this is ready for another review.

It should be ready for final review

h-mayorquin (changes requested):
Some more small request and his should be ready to go.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,361 @@
...
+            dictionary with key, values representing different types of Images used in segmentation:
+                Mean, Correlation image
+        """
+        return dict(
+            mean=self._image_mean,
+            correlation=self._image_correlation,
+            maximum_projection=self._image_maximum_projection,
+        )
+
+    def get_session_id(self) -> str:
```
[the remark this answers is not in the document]
alessandratrapani: sure!

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,360 @@
...
+
+        Returns
+        -------
+        np.ndarray
+            The original timestamps in seconds.
+        """
+
+        return self._read_timestamps_from_csv()
+
+    def sample_indices_to_time(self, sample_indices: Union[FloatType, np.ndarray]) -> Union[FloatType, np.ndarray]:
```
h-mayorquin, line 196:
```suggestion
    def sample_indices_to_time(self, sample_indices: FloatType |  np.ndarray) -> FloatType | np.ndarray:
```
I think that @pauladkisson prefers that

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -0,0 +1,361 @@
...
+        Returns
+        -------
+        image_masks: numpy.ndarray
+            The image masks for each ROI.
+        """
+        dataset = self._read_zarr_group("/A.zarr")
+        if dataset is None or "A" not in dataset:
+            raise ValueError("No image masks found in A.zarr dataset.")
+        else:
+            return np.transpose(dataset["A"], (1, 2, 0))
```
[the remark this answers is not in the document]
h-mayorquin: Thanks.

h-mayorquin (approved):
LGTM.

I am not sure whether it is right or desirable to extract the (estiamted) raw trace when the residuals are available the way we do in Caiman:

https://github.com/catalystneuro/roiextractors/blob/75afb5b4eedadf22a2a42cc22d8e5e2287a6e192/src/roiextractors/extractors/caiman/caimansegmentationextractor.py#L127-L128

But I think you are in a better position to decide this. Feel free to merge.

PR #373, opened 2024-10-09
Fixed bug with ScanImage's parse_metadata so that it works properly when hStackManager is disabled

### LINKED ISSUE #372: [Bug]: invalid literal for int() 'Inf' for SI.hStackManager.framesPerSlice #1108
See https://github.com/catalystneuro/neuroconv/issues/1108

pauladkisson:
Fixes #372

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 * Added specific error message for single-frame scanimage data [PR #360](https://github.com/catalystneuro/roiextractors/pull/360)
+* Fixed bug with ScanImage's parse_metadata so that it works properly when hStackManager is disabled [PR #373](https://github.com/catalystneuro/roiextractors/pull/373)
 
 ### Improvements
 * Removed unnecessary import checks for scipy, h5py, and zarr [PR #364](https://github.com/catalystneuro/roiextractors/pull/364)
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
@@ -114,8 +114,12 @@ def parse_metadata(metadata: dict) -> dict:
         where M is the number of channels (active or not).
     """
     sampling_frequency = float(metadata["SI.hRoiManager.scanFrameRate"])
-    num_planes = int(metadata["SI.hStackManager.numSlices"])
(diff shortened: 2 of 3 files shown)


pauladkisson:
@k1o0, This should fix the bug you found.  Please let me know if you run into any more problems.

h-mayorquin (commented):
This LGTM.  Two things:

* Any chance @k1o0 could test this just to be sure?
* Maybe we can stub the example file that he shared with us and add it to the examples? 

As we have many tiff based systems it might be a good investment of developer time to understand the format to the degree that we can stub so we can add this this and future examples to our testing suite. Probably this discussion deserves its own issue.

pauladkisson:
> Maybe we can stub the example file that he shared with us and add it to the examples?

Can be tricky to stub sometimes...
In this case one of the existing examples already has SI.hStackManager.enable = "false", so I think we have it covered.

h-mayorquin:
> > Maybe we can stub the example file that he shared with us and add it to the examples?
> 
> Can be tricky to stub sometimes... In this case one of the existing examples already has SI.hStackManager.enable = "false", so I think we have it covered.

So how come it was not failing?

pauladkisson:
> > > Maybe we can stub the example file that he shared with us and add it to the examples?
> > 
> > 
> > Can be tricky to stub sometimes... In this case one of the existing examples already has SI.hStackManager.enable = "false", so I think we have it covered.
> 
> So how come it was not failing?

It was only testing the metadata parsing function (not e.g. get_video), and instead of Inf, it had some (presumably inaccurate) ints for num_planes and frames_per_slice.

k1o0:
I was able to successfully extract the metadata and generate an NWB file from the tif. Happy for you to use this file as a test fixture.

h-mayorquin:
Thanks @k1o0 
we will be merging this and storing your file. As Paul says it might be tricky to stub (basically make a shorter version of it for continuous integration purposes) but I will give it a try as soon as I can.

Thanks for reporting the issue and following up.

One final question, can you give us a bit of provenance for the file? what system was used to acquire it and whatever you remember from when you make the recording?

h-mayorquin (approved):

h-mayorquin:
Ah, I see that you answered here:

https://github.com/catalystneuro/neuroconv/issues/1108#issuecomment-2407547884

Thanks a bunch.

PR #387, opened 2025-02-17
Deprecate channel argument in `get_frames`

### LINKED ISSUE #386: Deprecate `channel` in `get_frames` and `get_video`
I propose deprecating the channel argument from the `get_frames` method. The arguments are about accepting the conversion-centered nature of the library and also just enforcing certain implicit decisions that we already made:

* **Unused Parameter**: Currently, the channel argument is not utilized within the get_frames method. For instance, the Bruker extractor for multiple channels treats the channel as a stream rather than a parameter in get_frames. The parameter is then ignored which is inconsistent and breaks intent.

* **Aligned with Single-Channel NWB data types**: Both the `pynwb.phys` module and the new `ndx-microscopy` are designed around single-channel structures. Maintaining a one-to-one correspondence between an extractor and a PyNWB data scaffold simplifies the number of decisions that we have to make and reduces our burden. We can adapt the ophys pipeline in neuroconv to deal with multiple streams when there are channel streams present or continue with the converter approach but getting rid of the channel here means one less thing to deal with. Additionally, neuroconv lacks functionality for handling multiple channels through the channel argument in `get_frames`. We implicitly always convert the default channel and, as mentioned above, we rely on converters or other tricks to transform multiple channels. 

* **API Simplification**: Having the channel specified both during initialization and as an argument in `get_frames can lead to confusion, especially since the latter is not actively used. Removing this redundant parameter will streamline the API and reduce potential misunderstandings especially now in the era of agents where they rely on our implicit contract with abstract classes. I had this problem when developing an interface for Thor within the scope of the Cohen conversion.

h-mayorquin:
Courtesy of cline

Should fix #386

### DIFF
--- a/tests/test_numpymemmapextractor.py
+++ b/tests/test_numpymemmapextractor.py
@@ -107,49 +107,6 @@ class TestNumpyMemmapExtractor(unittest.TestCase):
         )
         check_imaging_equal(imaging_extractor1=extractor, imaging_extractor2=roundtrip_extractor)
 
-    def test_get_frames_indexing_with_single_channel(self):
-        # Build a video structure
-        num_rows = 10
-        num_columns = 20
-        num_channels = 1
-        frame_axis = 0
-        rows_axis = 1
-        columns_axis = 2
-        channels_axis = 3
-        dtype = "uint16"
-
-        video_structure = VideoStructure(
-            num_rows=num_rows,
-            num_columns=num_columns,
-            num_channels=num_channels,
-            rows_axis=rows_axis,
-            columns_axis=columns_axis,
-            channels_axis=channels_axis,
-            frame_axis=frame_axis,
-        )
-
-        # Build a random video
-        memmap_shape = video_structure.build_video_shape(self.num_frames)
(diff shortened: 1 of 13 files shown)


tests/test_numpymemmapextractor.py
```diff
@@ -107,49 +107,6 @@ def test_roundtrip(self, dtype, num_channels, num_rows, num_columns, buffer_size
         )
         check_imaging_equal(imaging_extractor1=extractor, imaging_extractor2=roundtrip_extractor)
 
-    def test_get_frames_indexing_with_single_channel(self):
```
pauladkisson, line 110: Do you know what's going on with these tests?

h-mayorquin: Cline removing them because there is a mention of channel.

pauladkisson: Ok, after giving it a think, I suppose we don't really need to replace these tests with anything.

pauladkisson (approved):

h-mayorquin:
Turns out that cline forgot to remove some of them : P

PR #365, opened 2024-09-24
Update ImagingExtractor

pauladkisson:
This PR is the first of a series that implement breaking changes to fully standardize the roiextractors api and update/streamline the testing suite.  Some of the main goals include
- removing channels argument to enforce 1 channel/extractor
- Switching to modularized mixins for testing and supporting pytest fixtures/parameterization
- Standard input validation, numpy indexing for 'get' methods, etc.

This PR focuses on `ImagingExtractor`, `FrameSliceImagingExtractor`, and `NumpyImagingExtractor`

PR #390, opened 2025-02-19
updated tifffile requirements

pauladkisson:
Fixes Failing Dailies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Features
 
 ### Fixes
+* Use tifffile.imwrite instead of tifffile.imsave for TiffImagingExtractor: [#390](https://github.com/catalystneuro/roiextractors/pull/390)
 
 ### Deprecations
 * The 'channel' parameter in get_frames() and get_video() methods is deprecated and will be removed in August 2025.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
 
 [project.optional-dependencies]
 full = [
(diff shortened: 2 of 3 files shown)


h-mayorquin:
I would vote against putting a ceiling on tifffile. Reasons:

1) Writing is not a central part of this library. It is nice to have but not central. Newer versions of tifffile offer new capabilities that I think we should not let it pass (what is a good word for it?).
2) We can use `imwrite` instead to accomplish the same goal. The specific function that we are using for writing was deprecated but the general writing functionality was not:

https://github.com/cgohlke/tifffile/blob/f1d053c1f9922a207b84df615eefa52c6fbae3bb/tifffile/tifffile.py#L1309

pauladkisson:
Fair enough. tifffile v2018.10.18 doesn't have imwrite, but v2018.11.6 and onward does.

h-mayorquin (approved):

h-mayorquin:
Yeah, I think we should probably drop the writing aspect as we don't use it and it is just one more thing to maintain.

pauladkisson:
> Yeah, I think we should probably drop the writing aspect as we don't use it and it is just one more thing to maintain.

Yes...one of the many things I fixed in my roiextractors 0.6.0 PRs...

h-mayorquin:
Oh, great, I did not know it was on the works. Are those PRs only in your system or is it here?

pauladkisson:
> Are those PRs only in your system or is it here?

Not exactly sure what you mean, but you can see I removed write_imaging in #365

h-mayorquin:
here means the repo so the answer is yes.  Thanks

PR #392, opened 2025-02-28
Deprecate `get_number_of_channels`

### LINKED ISSUE #391: Remove `get_number_of_channels`
Follow up on:
https://github.com/catalystneuro/roiextractors/issues/386

Now that we no longer have a channel as an argument of `get_frames` and we are sticking to the one extractor per channel argument I don't think it makes sense to have the `get_number_of_channels` method:

https://github.com/catalystneuro/roiextractors/blob/10e56eefe3862a98f69fe7b67f7f67245dabbe74/src/roiextractors/imagingextractor.py#L74-L83

I think we should remove it

h-mayorquin:
Should fix #391

This removes the method from the base class so new classes don't need to implement it and add deprecation warnings on the actual use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Deprecations
 * The 'channel' parameter in get_frames() and get_video() methods is deprecated and will be removed in August 2025.  [#388](https://github.com/catalystneuro/roiextractors/pull/388)
+* Removed get_num_channels from the ImagingExtractor abstract class. Implementations remain in concrete classes.
 
 ### Improvements
 * Use `pyproject.toml` for project metadata and installation requirements [#382](https://github.com/catalystneuro/roiextractors/pull/382)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Deprecations
 * The 'channel' parameter in get_frames() and get_video() methods is deprecated and will be removed in August 2025.  [#388](https://github.com/catalystneuro/roiextractors/pull/388)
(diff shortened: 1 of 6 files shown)


h-mayorquin:
Mmm this is an annoying behavior of the docstring checker, any way to avoid the check for some specific cases, @pauladkisson ?

pauladkisson:
> Mmm this is an annoying behavior of the docstring checker, any way to avoid the check for some specific cases?

Yes, it's an input called `names-to-skip`

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 
 ### Deprecations
 * The 'channel' parameter in get_frames() and get_video() methods is deprecated and will be removed in August 2025.  [#388](https://github.com/catalystneuro/roiextractors/pull/388)
+* Removed get_num_channels from the ImagingExtractor abstract class. Implementations remain in concrete classes. [#392](https://github.com/catalystneuro/roiextractors/pull/392)
```
h-mayorquin, line 22:
```suggestion
* Removed get_num_channels from the base ImagingExtractor as an abstract class. Implementations remain in concrete classes until deprecation on August 2025 [#392](https://github.com/catalystneuro/roiextractors/pull/392)
```

h-mayorquin:
Ok, this should be ready to go, @pauladkisson. We just need to remove the abstract condition of the function on the base. This allows us to not have to define them in the child classes and also avoids the need to write new docstrings. We can then remove the function when the deprecation period is due.

pauladkisson (approved):
Looks good, thanks @h-mayorquin!

PR #396, opened 2025-03-04
Add correct binary prefixes to text representation of ImaginExtractors

h-mayorquin:
I forgot to add the right prefix in #393. I also added clarifying comments on the docstring of the respective function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Improvements
 * Use `pyproject.toml` for project metadata and installation requirements [#382](https://github.com/catalystneuro/roiextractors/pull/382)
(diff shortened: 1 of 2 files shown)


pauladkisson (approved):

h-mayorquin:
Also, should build the docs be required for merging? I feel that we might want to skip situations like this.

pauladkisson:
> Build the docs is getting stuck, @pauladkisson. Any ideas on how to check the log for this one? It was happening before.

Looked at the buildthedocs [dashboard](https://readthedocs.org/projects/roiextractors/builds/), but I don't see any pending builds...

pauladkisson:
> Also, should build the docs be required for merging? I feel that we might want to skip situations like this.

I think it's fair to require it, just like the tests, we really don't want to merge with failing docs. But I can skip for this PR, since it's not doc-related.

PR #401, opened 2025-03-11
Improve scan image number of channel detection

h-mayorquin:
1. Updated channel detection logic in ScanImage TIFF processing:
  - Changed from using `SI.hChannels.channelsActive` to `SI.hChannels.channelSave` for determining saved channels. Kept the old logic for backup when the former key is not available.
  - Per Kim Lab: 
    > Actually, whether the channels are active or not doesn't matter to the experiment. When i do 1 channel imaging, I usually leave both channels on, but the shutter on the 2nd channel is not on so no laser is coming through and no images from channel 2 will be saved. (My other labmates sometimes like to leave the 2nd channel on, sometimes turn it off.)

2. Added missing `get_num_planes` method to `ScanImageTiffMultiPlaneMultiFileImagingExtractor` class.

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
@@ -121,6 +121,8 @@ def parse_metadata(metadata: dict) -> dict:
         num_planes = 1
         frames_per_slice = 1
 
+    # `channelSave` indicates whether the channel is saved. Note that a channel might not be saved even if it is active.
+    # We check `channelSave` first but keep the `channelsActive` check for backward compatibility.
     channel_availability_keys = ["SI.hChannels.channelSave", "SI.hChannels.channelsActive"]
     for channel_availability in channel_availability_keys:
         if channel_availability in metadata.keys():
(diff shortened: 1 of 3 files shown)


src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
```diff
@@ -120,8 +120,14 @@ def parse_metadata(metadata: dict) -> dict:
     else:
         num_planes = 1
         frames_per_slice = 1
-    active_channels = parse_matlab_vector(metadata["SI.hChannels.channelsActive"])
-    channel_indices = np.array(active_channels) - 1  # Account for MATLAB indexing
+
+    channel_availability_keys = ["SI.hChannels.channelSave", "SI.hChannels.channelsActive"]
```
pauladkisson, line 124: Can you add a note in the docstring of this function?

pauladkisson (changes requested):
Looks good just needs updated documentation.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
h-mayorquin: Good idea.

h-mayorquin: Done.

pauladkisson (approved):

PR #402, opened 2025-03-11
[Cline draft] MultiTiffMultiFile

h-mayorquin:
We did this together in the AI strategy meeting. 

I have not reviewed it, I will take a look first (in a while as this is not a priority) and then come back to you guys.

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/multitiffmultipageextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/multitiffmultipageextractor.py
@@ -1,0 +1,448 @@
+"""Extractor for multiple TIFF files, each with multiple pages.
+
+Classes
+-------
+MultiTIFFMultiPageExtractor
+    An extractor for handling multiple TIFF files, each with multiple pages, organized according to a specified dimension order.
+"""
+
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple, Union, NamedTuple
+from collections import defaultdict
+import warnings
+import numpy as np
+import glob
+
+from ...extraction_tools import PathType, FloatType, ArrayType, DtypeType, get_package
+from ...imagingextractor import ImagingExtractor
+
+
+class MultiTIFFMultiPageExtractor(ImagingExtractor):
+    """
+    An extractor for handling multiple TIFF files, each with multiple pages, organized according to a specified dimension order.
+
+    This extractor allows for lazy loading of frames from multiple TIFF files, where each file may contain multiple pages.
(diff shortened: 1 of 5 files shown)


tests/test_multitiffmultipageextractor.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+        # Check basic properties
+        self.assertEqual(extractor_xyztc.get_num_frames(), 2)
+
+        # Get frames and check they're accessible
+        frame_0 = extractor_xyztc.get_frames(0)
+        self.assertEqual(frame_0.shape, (10, 12, 3))
+
+    def test_comparison_with_scanimage(self):
```
h-mayorquin, line 169: Review this. This is the most critical part. Apparently the scan image example file used here should not be volumetric.

src/roiextractors/extractors/tiffimagingextractors/multitiffmultipageextractor.py
```diff
@@ -0,0 +1,448 @@
...
+        self._kwargs = {
+            "file_paths": [str(path) for path in self.file_paths],
+            "dimension_order": dimension_order,
+            "channel_size": channel_size,
+            "time_size": time_size,
+            "depth_size": depth_size,
+            "sampling_frequency": sampling_frequency,
+        }
+
+    def _create_frame_mapping(
```
h-mayorquin, line 148: This is the central part of the code where the mapping is done. Review this carefully. Check the same code on the ThorTiff extractor to see of a similar design pattern.

tests/test_multitiffmultipageextractor.py
```diff
@@ -0,0 +1,266 @@
...
+    assert extractor_ch1.get_num_samples() == num_timepoints
+
+    # Verify different channels have different data
+    series_ch0 = extractor_ch0.get_series()
+    series_ch1 = extractor_ch1.get_series()
+
+    assert not np.array_equal(series_ch0, series_ch1)
+
+
+def test_error_handling(tmp_path):
```
pauladkisson, line 239: I can't see the code coverage report for some reason. But I believe there's quite a few error cases that should probably be tested here, like the RuntimeError and the ValueError on invalid dimension sizes, etc. Also, I don't think there's any reason to bundle them up into one test. We can probably split them out. So test_invalid_file_path, test_invalid_dimension_order, test_invalid_tiff_file, and so on.

pauladkisson (changes requested):
Just a small request for more detailed tests on the error cases. But other than that, looks great.

PR #405, opened 2025-03-19
Fix `ScanImage` sampling rate for volumetric extractor

h-mayorquin:
Currently, `ScanImage` retrieves the same sampling rate for both volumetric and non-volumetric data. This is wrong and the sampling rate should be slower for volumetric data. I am fixing this here.

Additionally, I am extracting the metadata using `read_scanimage_metadata` from the corresponding `tifffile` library. This approach has several advantages:

- I spoke with Lawrence from the ScanImage team today, and he indicated that the `scanimage-tiff-reader` library should be considered unmaintained. We've already encountered problems with ARM Macs, so transitioning to this method is a step in the right direction.
- The `read_scanimage_metadata` function directly returns metadata parsed into Python objects. This means we no longer need to maintain our own parsing logic like parsing matlab vector strings. 
- It's significantly faster—approximately 40 times, based on my preliminary profiling.

image

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
@@ -163,6 +163,7 @@ def read_scanimage_metadata(file_path: PathType) -> dict:
             break
 
     available_channels = non_varying_frame_metadata[channel_availability]
+    available_channels = [available_channels] if not isinstance(available_channels, list) else available_channels
     channel_indices = np.array(available_channels) - 1  # Account for MATLAB indexing
     channel_names = non_varying_frame_metadata["SI.hChannels.channelName"]
     channel_names_available = [channel_names[i] for i in channel_indices]
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
@@ -120,8 +120,16 @@ def parse_metadata(metadata: dict) -> dict:
     else:
         num_planes = 1
         frames_per_slice = 1
-    active_channels = parse_matlab_vector(metadata["SI.hChannels.channelsActive"])
-    channel_indices = np.array(active_channels) - 1  # Account for MATLAB indexing
+
+    # `channelSave` indicates whether the channel is saved. Note that a channel might not be saved even if it is active.
+    # We check `channelSave` first but keep the `channelsActive` check for backward compatibility.
+    channel_availability_keys = ["SI.hChannels.channelSave", "SI.hChannels.channelsActive"]
+    for channel_availability in channel_availability_keys:
+        if channel_availability in metadata.keys():
+            break
+
+    available_channels = parse_matlab_vector(metadata[channel_availability])
+    channel_indices = np.array(available_channels) - 1  # Account for MATLAB indexing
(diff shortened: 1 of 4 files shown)


tests/test_scanimagetiffimagingextractor.py
```diff
@@ -31,7 +31,7 @@ def file_path():
 @pytest.fixture(scope="module")
 def expected_properties():
     return dict(
-        sampling_frequency=29.1248,
+        sampling_frequency=7.28119,
```
h-mayorquin, line 34: This is what is failing on the dev tests. They should fail because we were getting the frame rate instead of the volume rate.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiff_utils.py
```diff
@@ -86,6 +86,111 @@ def parse_matlab_vector(matlab_vector: str) -> list:
...
+    channel_indices = np.array(available_channels) - 1  # Account for MATLAB indexing
+    channel_names = non_varying_frame_metadata["SI.hChannels.channelName"]
+    channel_names_available = [channel_names[i] for i in channel_indices]
+    num_channels = len(channel_names_available)
+    if roi_group_metadata:
+        roi_metadata = roi_group_metadata["RoiGroups"]
+    else:
+        roi_metadata = None
+
+    metadata_parsed = dict(
```
h-mayorquin, line 176: I am adding this for backwards compatibility but probably not strictly necessary.

tests/test_scanimagetiffimagingextractor.py
pauladkisson: Ok, can you link an issue or a PR on neuroconv for this?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -71,6 +74,9 @@ def __init__(
         self._num_planes = imaging_extractors[0].get_num_planes()
         super().__init__(imaging_extractors=imaging_extractors)
 
+    def get_sampling_frequency(self) -> float:
+        return self._imaging_extractors[0].get_sampling_frequency()
+
```
pauladkisson, line 79: This is already defined in MultiImagingExtractor

pauladkisson (changes requested):
read_scanimage_metadata looks good, but the get_sampling_frequency method is redundant and should be removed.

tests/test_scanimagetiffimagingextractor.py
h-mayorquin: How could that work?

The released extractor gets the wrong sampling rate. We need to release this first to fix it there I feel.

How you have handled this things on the past? I did not realize that you had downstream dev tests here until now.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Yes, that's a mistake.

tests/test_scanimagetiffimagingextractor.py
pauladkisson: > The released extractor gets the wrong sampling rate. We need to release this first to fix it there I feel.

Yes, that's right. I just find it easier to coordinate these changes to minimize down-time for neuroconv.

pauladkisson (approved):

Issue #406, opened 2025-03-19
Frame terminology is confusing for volumetric data


h-mayorquin:
This is something that @alessandratrapani pointed out in the last conversion meeting and is related to the discussion in #394.

The language of the abstract `ImagingExtractor` is very planar-microscopy-oriented (for good reasons). However, this means some of the terminology is unclear or confusing. Examples:

- `get_image_size`: It's unclear if this refers to the shape of the volume or the shape of the frame.
- `get_num_frames`: It's unclear if this refers to the number of samples (equivalent to the time dimension size) or the number of images/frames in the series.
- `get_frames`: This returns samples, but it's unclear if it should return volumes or frames in the volumetric case.

### Proposal

Make a clear distinction by consistently using:

- **frame** for referring to (x, y). This is what `ScanImage` does which is arguably the most popular software of acquisition. 
- **volume** for referring to (x, y, depth).
- **samples** explicitly for sample data that works in both cases. So a sample is (x, y) if the extractor is planar but (x, y, z) if it is volumetric.

We should also structure the important abstract classes around the NWB-conversion-centered nature of this library. From that perspective, what matters most are samples.

An example of the proposed convention would be:

- `get_image_shape`: returns `(x, y)` and `get_volume_shape` that returns (x, y, z)
- `get_num_samples`: returns the number of frames in the planar case and the number of volumes in the volumetric case.
- Rename `get_frames` to `get_samples`, clearly indicating it returns frames in the planar case and volumes in the volumetric case. That is `get_samples` returns (x, y) on planar and (x, y, z) on volumetric. This will also allows us to have the method `get_frames` (or alternatively get_images` to return the frames even in the volumetric case.

Don't really know what to do with get_video though as I think that is very planar oriented.

What do you think?

alessandratrapani:
I've been thinking about this, and I'm totally on board with the proposed changes. Love the idea of clearly defining **frame** vs **volume** and making **samples** more flexible.

One thing I'd add - I would not use the term frame if we are referring to a slice of a volume. Thus, the `get_frames()` function would be only applicable to the planar extractor, but it would be redundant since we already have `get_samples()`. 
If we really need a function that returns 2D slices from the volumetric imaging data, I would opt for something like `get_slices_at_z()`.

I think that in this context, `get_video()` has become pretty useless.

The suggested naming conventions make way more sense and will definitely make the code more intuitive. 
Especially, digging the distinction between `get_image_shape()` and `get_volume_shape()`. 

[...]

PR #408, opened 2025-03-24
Standardize parameter name in `get_frames()` method from `frame_idxs` to `frames` for consistency 

### LINKED ISSUE #202: Unify terminology for frames in `frame_to_time` and  `get_frame`.
Right now we have `frames` on the functions for getting time:
https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/imagingextractor.py#L69

And `frame_idxs` in the function for getting the frames:
https://github.com/catalystneuro/roiextractors/blob/b71f17db4d3dadeaa4ae02c806c1e269f03d3d9b/src/roiextractors/imagingextractor.py#L62-L68
But both of them are referring to the same. We should use either `frames`, `frame_idxs` or `frame_indices` for the two of them.  Numpy uses `indices` for their indexing routines so maybe that's a good choice:
https://numpy.org/doc/stable/reference/arrays.indexing.html#routines-indexing

But I don't think it matters that much.

daphnedequatrebarbes:
fix #202 

### Summary of Changes  
I've completed the standardization of parameter name in the `roiextractors` library. Below is an overview of the modifications and potential considerations:  

#### **Changes Made**  
-Standardized Parameter Names: Renamed the `frame_idxs` parameter to `frames` in the `get_frames()` method across extractors.
-Maintained Backward Compatibility: Ensured that previous implementations remain functional and modify the corresponding tests. 

### **Testing & Platform-Specific Considerations**  

- **All accessible tests pass** with the standardized parameter names.  
- **Platform-Specific Test Limitations**: Some local tests couldn't be run due to macOS compatibility issues:  
  - `test_scanimagetiffimagingextractor.py` – Platform compatibility 
  - `test_scan_image_tiff.py` – Platform compatibility
  - `test_thortiffimagingextractor.py` – Memory limitations (process killed)  

I am still working on addressing platform-specific compatibility and memory-related test limitations, as my GitHub push is not passing the checks for similar reasons. I would appreciate any feedback on resolving these issues and any insights on whether I might have misunderstood the underlying causes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,15 +14,15 @@
 ## [Unreleased]
 
 ### Changed
-- Standardized parameter name in `get_frames()` method across all extractors from 
-  `frame_idxs` to `frames` for better consistency. This maintains backward compatibility 
+- Standardized parameter name in `get_frames()` method across all extractors from
+  `frame_idxs` to `frames` for better consistency. This maintains backward compatibility
   while aligning with existing test code and other extractors in the library.
-  
+
   For example:
   ```python
   # Old style (still supported but will be deprecated)
   extractor.get_frames(frame_idxs=[0, 1, 2])
-  
+
   # New standardized style
   extractor.get_frames(frames=[0, 1, 2])
   ```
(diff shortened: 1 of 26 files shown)


h-mayorquin:
See #406 for more recent discussion.

PR #410, opened 2025-03-27
Remove `installed` attribute

h-mayorquin:
I think this was left from the SpikeExtractors time and it does not have any use whatsoever currently.


weiglszonja (approved):
Yeah, I don't it is used anymore. LGTM

PR #412, opened 2025-04-03
Add scan imaging extractor

h-mayorquin:
[Draft] will explain in issue.

PR #417, opened 2025-04-22
Add support for multiple samples per frame to `ScanImage`

h-mayorquin:
Should come #412.

@pauladkisson 
This is ready as well and is mostly testing.


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -651,6 +682,31 @@ def get_plane_extractor(self, plane_index: int) -> ImagingExtractor:
 
         return sliced_extractor
 
+    @staticmethod
+    def get_slices_per_sample(file_path: PathType) -> int:
```
pauladkisson, line 686: Couple questions:
1. Why are you defining this here as a staticmethod if it only gets used in the tests?
2. What is the point of introducing a new term (slices per sample) if it's just frames per slice?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -72,6 +75,10 @@ def __init__(
             file detection heuristics. Use this if automatic detection does not work correctly and you know
             exactly which files should be included.  The file paths should be provided in an order that
             reflects the temporal order of the frames in the dataset.
+        slice_sample : int, optional
```
pauladkisson, line 78: What if a user wants to include multiple frames per slice as in https://github.com/catalystneuro/neuroconv/issues/1260#issuecomment-2738055264?

tests/test_scanimage_extractor.py
```diff
@@ -107,6 +107,278 @@ def test_planar_multi_channnel_multi_file(self):
...
+        This test does not have enough samples to do a full slice extraction. It has 20 slices
+        and 8 frames per slice but only 20 pages of data. This means that only data from the first three
+        depths are included. Skipping this test for now.
+        """
+        return
+
+    def test_volumetric_data_multi_channel_single_file(self):
+        """Test with volumetric data acquired in grab mode with multiple frames per depth.
+
+        File: scanimage_20220923_noroi.tif
```
pauladkisson, line 155: These file description blocks are nice. I bet they will come in handy later.

pauladkisson (commented):
Tests look good but not sure about the implementation. I think we should have a way to include all the frames in a single extractor without having to specify only one slice_sample.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Ah, thanks for pointing out the bad term. I mixed that up badly. This is me being tired and dumb : P. I will correct.

The static method is to enable the workflow B here:

https://github.com/catalystneuro/neuroconv/issues/1260#issue-2921052096

1) First you find out the number of frames_per_slice. 
2) You instantiate one extractor per sample.
3) You write each of them as a separate Series.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Not implemented yet. I need to think more carefully about the timing in that case and the resources for the Kim Conversion are not enough for me to implement at the moment.

Let me implement a draft so you get an idea of what would look though.

h-mayorquin: Here you go:

https://github.com/catalystneuro/roiextractors/pull/420

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: I see, so it's analogous to the `get_available_channels` and `get_available_planes` methods from the older extractors?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -651,6 +682,33 @@ def get_plane_extractor(self, plane_index: int) -> ImagingExtractor:
 
         return sliced_extractor
 
+    @staticmethod
+    def get_frames_per_slice(file_path: PathType) -> int:
+        """
+        Get the number of slices per sample from a ScanImage TIFF file.
```
pauladkisson, line 688:
```suggestion
        Get the number of frames per slice from a ScanImage TIFF file.
```

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -651,6 +682,33 @@ def get_plane_extractor(self, plane_index: int) -> ImagingExtractor:
...
+
+        Parameters
+        ----------
+        file_path : PathType
+            Path to the ScanImage TIFF file.
+
+        Returns
+        -------
+        int
+            Number of slices per sample.
```
pauladkisson, line 700:
```suggestion
            Number of frames per slice.
```

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Thanks.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: Thanks for the draft implementation, that was very helpful.

Based on the issues I flagged there, I think a bit more care is needed with frames_per_slice than just treating it as an artificial depth. Probably, the frame_per_slice_index needs to be another mapped entry in the ifd table.

h-mayorquin: I think the PR you suggested is an improvement, and the readability and clarity are more important than generalization in this caso. I also think it's good that you took a closer look at the IFD mapping approach. Thanks for taking the time to review this. Feel free to merge your PR here.

pauladkisson (approved):
This should be good to go now

PR #418, opened 2025-04-23
Add  `get_volume_shape`

### LINKED ISSUE #406: Frame terminology is confusing for volumetric data

h-mayorquin:
This should close #406  

This is the final addition to complete the program that @alessandratrapani proposed here:

https://github.com/catalystneuro/roiextractors/issues/406#issuecomment-2760922865

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Features
 * New `read_scanimage_metadata` for reading scanimage metadata from a file directly as a python dict [#405](https://github.com/catalystneuro/roiextractors/pull/401)
+* Added volumetric imaging support with `is_volumetric` flag, `get_frame_shape`, `get_num_planes`, and `get_video_shape` methods [PR #TBD](https://github.com/catalystneuro/roiextractors/pull/XXX)
 
 
 ### Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Features
 * New `read_scanimage_metadata` for reading scanimage metadata from a file directly as a python dict [#405](https://github.com/catalystneuro/roiextractors/pull/401)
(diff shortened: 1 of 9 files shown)


h-mayorquin:
Let's add some tests.

Issue #413, opened 2025-04-03
Improve `ScanImage` data extraction


h-mayorquin:
Hey, while working with `ScanImage` with the Kim Lab from the Cohen conversion, I have been reading the metadata in more detail and also talking with Lawrence Niu from MBF (who are behind ScanImage).

Currently, if the user wants to extract data from ScanImage, they need to choose the appropriate extractor with different options depending on planar vs. volumetric and single file vs. multi file. So we end up having `ScanImageTiffMultiPlaneMultiFileImagingExtractor`, `ScanImageTiffSinglePlaneMultiFileImagingExtractor`, `ScanImageTiffMultiPlaneImagingExtractor`, and `ScanImageTiffSinglePlaneImagingExtractor`, which depend on each other, with the `ScanImageTiffSinglePlaneImagingExtractor` being the place where the extraction logic is implemented.

In [this opened PR](https://github.com/catalystneuro/roiextractors/pull/412), I am proposing a simplification of this, where we have a single extractor that takes as input the `file_path` and determines automatically whether the data is volumetric or planar (see [here](https://github.com/catalystneuro/roiextractors/blob/8ce234a86f28b793097222ff38eea4c2707cf7a8/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L100-L103)), whether it is single file or multi file (see [here](https://github.com/catalystneuro/roiextractors/blob/8ce234a86f28b793097222ff38eea4c2707cf7a8/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L223-L273)), and can handle the multi-channel nature of ScanImage (see [here](https://github.com/catalystneuro/roiextractors/blob/8ce234a86f28b793097222ff38eea4c2707cf7a8/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L129-L140)). This is a simpler API, with less code and fewer interactions between classes, which should be easier to maintain and reduce complexity.

[...]

PR #419, opened 2025-04-23
Add flyback frames support for `ScanImageImagingExtractor`

h-mayorquin:
Should come after https://github.com/catalystneuro/roiextractors/issues/413 and has it as a base.


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -235,6 +230,7 @@ def _create_frame_to_ifd_table(
         num_acquisition_cycles: int,
         ifds_per_file: list[int],
         num_frames_per_slice: int = 1,
+        flyback_frames: int = 0,
```
pauladkisson, line 233: Nit: I think it should be called `num_flyback_frames` (more explicit and matches the other args).

pauladkisson:
> Unfortunately, we don't have testing data for this yet.

Any chance we can get some from Lawrence Niu?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: I agree with this.

h-mayorquin: Changed.

h-mayorquin:
> > Unfortunately, we don't have testing data for this yet.
> 
> Any chance we can get some from Lawrence Niu?

Here from the Kim Lab.
https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/26

h-mayorquin:
OK, this is ready @pauladkisson

pauladkisson (approved):
Looks great!

For my own understanding, why does scanimage record flyback frames at all? It seems like they wouldn't be very useful for scientific purposes. And if they are useful, maybe we shouldn't just throw them away?

h-mayorquin:
Looks great!

> For my own understanding, why does scanimage record flyback frames at all? 

I am not sure why those are not dropped directly from the acquisition system. The contact person from the Kim Lab claims they have no value and are byproduct of the acquisition system. I agree and I can't find a reason why she is wrong. We could put the effort to write just the flyback frames as a series of their own as Alessandra suggests here:

https://github.com/catalystneuro/neuroconv/issues/1265

But I don't have the hours right now to do that effort and I would like to focus on the data.

PR #420, opened 2025-04-23
`ScanImageImagingExtractor` get full series as in the order of acquisition

h-mayorquin:
From the discussion on  #417 which is the base of this.

This showcases the idea on how to solve this problem. The testing needs to be done more carefully paying attention to the times and the sampling frequency.


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -79,11 +80,30 @@ def __init__(
             When frames_per_slice > 1 (multiple frames per slice), this parameter specifies which frame to use
             for each slice. Must be between 0 and frames_per_slice-1. If None and frames_per_slice > 1,
             a ValueError will be raised.
+        load_series_as_contigious_samples : bool
+            False by default.
+
+            When true, all the samples are loaded as they appear in the acquisition cycle:
+
+            depth_1_frame_sample_1
+            depth_1_frame_sample_1
```
h-mayorquin, line 89:
```suggestion
            depth_1_frame_sample_2
```

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -79,11 +80,30 @@ def __init__(
...
+            depth_1_frame_sample_3
+            depth_2_frame_sample_1
+            depth_2_frame_sample_1
```
h-mayorquin, line 92:
```suggestion
            depth_2_frame_sample_2
```

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -235,6 +257,12 @@ def __init__(
             )
 
             self._frames_to_ifd_table = slice_frames_to_ifd_table
+        elif has_multiple_slices and self._slice_sample is None and load_series_as_contigious_samples:
+            # Adjust depth index to true value so it can be sliced but not drop frames
+            channel_frames_to_ifd_table["depth_index"] = (
```
pauladkisson, line 262: If you slice this out, wouldn't the ifd table essentially have duplicates? Like would there be any way to distinguish between the first frame and second frame of a 2-frame-per-slice acquisition cycle?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -79,11 +80,30 @@ def __init__(
             When frames_per_slice > 1 (multiple frames per slice), this parameter specifies which frame to use
             for each slice. Must be between 0 and frames_per_slice-1. If None and frames_per_slice > 1,
             a ValueError will be raised.
+        load_series_as_contigious_samples : bool
+            False by default.
+
+            When true, all the samples are loaded as they appear in the acquisition cycle:
+
+            depth_1_frame_sample_1
```
pauladkisson, line 88: This isn't exactly what I had in mind. depth should still be a distinct dimension, it's just that the frames_per_slice are concatenated in time instead of being treated as different dimensions.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -401,6 +429,13 @@ def get_series(self, start_sample: Optional[int] = None, end_sample: Optional[in
             For a volumetric dataset with 5 planes and 512x512 frames, requesting 3 samples
             would return an array with shape (3, 512, 512, 5).
         """
+        if self.load_series_as_contigious_samples:
+            error_msg = (
```
pauladkisson, line 433: This shouldn't be an error case. get_series should work fine even if you haven't sliced out the different planes yet.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -625,6 +660,13 @@ def get_num_planes(self) -> int:
         int
             Number of depth planes.
         """
+        if self.load_series_as_contigious_samples:
+            error_msg = (
```
pauladkisson, line 664: Similarly, this shouldn't be a problem.

pauladkisson (commented):
I understand that this is just a first pass implementation, but I'm highlighting some high-level issues I think might be relevant for #417

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Yeah, depth is still a different dimension it is just that your volumes are larger than they actually are. I re-wrote this explanation let me know if this matches what you had in mind.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: I'm not sure I follow you here, since there's no slicing involved. But just to add some context, here's the table with the artificial depth:

image

And here's what it looks like after `depth_index` has been corrected:

image

You can tell the slice samples apart because they're now in the correct order within the same depth and acquisition index. But ... why would you even want to do that? This is just a proxy object meant to provide access to planar extractors for each plane in your proposal C right? That’s the only purpose of this table, and it's why I added errors for cases where someone tries to call `get_series` or `get_planes` here.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Hmm, I don't think using `get_series` like this makes sense for writing to NWB. The acquisition metadata won't align—for example, the grid space—and it won't be compatible with the artificially large volumes. 

Also, this approach doesn’t match scenario A, B, or your proposed C from this discussion:

https://github.com/catalystneuro/neuroconv/issues/1260

What do you mean when you say it should work fine?

pauladkisson: Nevermind, you're right that it wouldn't work. I keep forgetting that multiple frames-per-slice can't be ordered in time when depth dimensions are all taken together.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Wait, I pushed the wrong one, let me correct.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: But it means that this object is non-sensical and basically useless before get_plane_extractor is applied.  I think it would be much better if we had a way to automatically apply plane-wise slicing so that you get usable extractor(s) right from the start.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Actually, easier to discuss it here:

image

This is a representation of the frames as acquired -with artificial volume- that can be used to implement the proposed workflow C here:

https://github.com/catalystneuro/neuroconv/issues/1260

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Yeah, I agree with you. I also don't like adding yet another parameter to this class.

I am thinking on a static method that returns a dictionary with an extractor per plane and can be used directly to implement the workflow C.

I just wanted to showcase how we would need to slice the table and then get planar extractors.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
pauladkisson: I mean that after artificial depth has been corrected, once you apply get_plane_extractor, you'll have a new ifd_table where all the depth_index values are the same. In this table, the first two entries would be identical except for the IFD_index.

Then, when you call get_series(start_sample=1, end_sample=2) that would find the frame with acquisition_cycle_index=1 (right?) so there would no way to get the second frame with acquisition_cycle_index=0 (I think?).

h-mayorquin: `get_series` works positionally, so it returns the frames in the correct order:

https://github.com/catalystneuro/roiextractors/blob/eec59e2b1b6a70ce164c7afa9a39ff17b033d5a9/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L382-L394

For example, if you call:

```python
planar_extractor = recording.get_plane_extractor(plane_index=0)
frames = planar_extractor.get_series(start_frame=0, end_frame=6)
```

The frames will come from this order:

```
[acquisition_cycle_1]
frame_depth_0_sample_0 [IFD 0] - sample 0
frame_depth_0_sample_1 [IFD 2] - sample 1

[acquisition_cycle_2]
frame_depth_0_sample_0 [IFD 8] - sample 2
frame_depth_0_sample_1 [IFD 10] - sample 3

[acquisition_cycle_3]
frame_depth_0_sample_0 [IFD 16] - sample 4
frame_depth_0_sample_1 [IFD 18] - sample 5
```

pauladkisson: I see, so for get_series you really only need file_index and IFD_index in the right order in the ifd_table. The other columns (channel_index, etc.) are just for convenient slicing?

h-mayorquin: Right. The goal is to get a positional map of the (IFD_index, file_index) tuples.

I thought about adding more elements according to the situation (adding a flyback frame marker for example) but I would like to keep a consistent format that does not depend on the context as common routines will keep working. Moreover, this is also a bit more memory efficient for large datasets.

pauladkisson: In that case there's probably nothing functionally wrong with treating frames_per_slice as artificial depth, but I think stylistically it's still probably better to be explicit about how frames_per_slice affects the ifd_table, and I think it will be less confusing if we need to modify stuff later on.

h-mayorquin: >  it's still probably better to be explicit about how frames_per_slice affects the ifd_table,

What do you propose exactly?

pauladkisson: > Are you talking about explicitly adding `slice_sample_index` (or something similar) to the table?

At least, adding a `num_frames_per_slice` argument to the _create_frame_to_ifd_table method. Maybe `slice_sample_index` as well if it makes subsetting the table easier.

> The reason I'm hesitant is that I believe most TIFF-based formats can be handled using a common abstraction via this table: Bruker, Thorlabs, the generic TIFF extractor proposed in #402, and a future [OME-TIFF extractor](https://docs.openmicroscopy.org/ome-model/5.6.3/ome-tiff/), that we should implement at some point. Having a shared format would let us rely on a common set of abstractions, which I think is valuable and would make maintenance easier. That's why I'm reluctant to add ScanImage specific properties to the table.
> 
> That said, I do agree with you that it would improve readability and probably make debugging easier, which is also very important. I guess it's a trade-off between generalization and clarity.

These are fair points but I think frames_per_slice might well generalize outside ScanImage at some point, and it is very easily incorporated in the default case (num_frames_per_slice=1).

pauladkisson: > > it's still probably better to be explicit about how frames_per_slice affects the ifd_table,
> 
> What do you propose exactly?

I can PR into #417 to make it concrete.

h-mayorquin: Yeah, thanks, but I would rather have you do another PR who has that as its base.

h-mayorquin: >  frames_per_slice might well generalize outside ScanImage at some point

I know those other formats and they don't have that feature.

h-mayorquin:
Let's put this on draft. Thanks again for the discussion, Paul's.

h-mayorquin:
Ok this can be accomplished by loading with interleaved samples and sampling slice. I will cose this.

PR #421, opened 2025-04-23
added slice_sample_index to ifd table

pauladkisson:


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -308,14 +304,16 @@ def _create_frame_to_ifd_table(
         # For ScanImage, the order is always CZT which means that the channel index comes first,
         # followed by depth and then acquisition cycle
```
h-mayorquin, line 305:
```suggestion
        # followed by the frames per slice, then depth and finally the acquisition cycle
```

h-mayorquin (approved):
Yes, thanks. Some advantages of your proposal:

* Keeps the meaning of `num_planes` "real" without having to hack it.
* Allows the mapping structure to represent the index which makes the masking easier and might be useful for debuggin.
* Avoids the need to remapping the depth_index.

Issue #394, opened 2025-02-28
`get_image_size()` is confusing for volumetric data


h-mayorquin:
This method returns a volume shape for volumetric data which is only the image shape in some extended sense. 

https://github.com/catalystneuro/roiextractors/blob/30b37f3d8c4f7f438b347b472997eae76d1b9d05/src/roiextractors/volumetricimagingextractor.py#L136-L146

I think it would be better to have a method called `get_image_shape` that always returns the size of the image. Then another method called `get_frame_shape` or `get_sample_shape` that does what the current `get_image_size` is currently doing.  

What do you think?

pauladkisson:
> [@pauladkisson](https://github.com/pauladkisson) [@alessandratrapani](https://github.com/alessandratrapani)
> 
> Any takes?

I don't know if we really need to change the name of the method (`get_image_size` seems fine), but it should definitely be just a single image (num_rows, num_columns), and then one can use a separate method for `get_num_frames` and `get_num_planes`.

h-mayorquin:
Thanks

shape vs size to be numpy friendly. That's the gain of changing. I would like it but I think is a marginal gain.

The big win is to change the return type of the method on the volumetric case though. As we seem to agree on that let's wait to see if someone else has any takes.

pauladkisson:
> shape vs size to be numpy friendly.

Ah, I see. Yeah that makes sense, we should follow numpy if we can.

alessandratrapani:
> > shape vs size to be numpy friendly.
> 
> Ah, I see. Yeah that makes sense, we should follow numpy if we can.

Agree

h-mayorquin:
In fact, using image in the sense of a frame clashes with how OME thinks about images. Just to add another data point:

![Image](https://github.com/user-attachments/assets/9dc1b38c-9e84-427f-81f3-317aa1073017)

https://genomebiology.biomedcentral.com/articles/10.1186/gb-2005-6-5-r47

pauladkisson:
If we really want to be pedantic and avoid any name conflicts, we could just split it up into `get_num_rows` and `get_num_columns`

h-mayorquin:
Related: https://github.com/catalystneuro/roiextractors/issues/406

PR #422, opened 2025-04-25
Fix imaging extractor string representation for volumetric imaging

h-mayorquin:
This brings some improvements:

1) The current representation is using image_size  which is deprecated and also leads to wrong estimates for volumetric data. This new PR uses a the shape of the sample so the memory estimation is correct.
2) Introduces the `get_sample_shape` which returns `get_image_shape` for planar extractor and `get_volume_shape` for volumetric extractors. Following on #406
3) It moves the methods to format times and memory to their own module to keep the base class clean.

### DIFF
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -112,8 +112,7 @@ class ImagingExtractor(ABC):
 
         Notes
         -----
-        This method is equivalent to get_image_shape() and is provided for consistency
-        with the naming convention used in volumetric extractors.
+        This method is equivalent to get_image_shape()
         """
         return self.get_image_shape()
 
@@ -121,8 +120,14 @@ class ImagingExtractor(ABC):
         """
         Get the shape of a single sample elements from the series.
 
-        If the series is volumetric, the shape is the shape of the volume, otherwise
-        thsi returns the shape of a single frame/image.
+        If the series is volumetric, the shape is the shape of the volume and otherwise
+        returns the shape of a single frame/image.
+
+        Returns
+        -------
+        sample_shape: tuple
+            Shape of a single sample from the time series (num_rows, num_columns, num_planes)
+            if volumetric (num_rows, num_columns) otherwise
         """
         if self.is_volumetric:
             return self.get_volume_shape()
--- a/src/roiextractors/imagingextractor.py
+++ b/src/roiextractors/imagingextractor.py
@@ -116,7 +116,7 @@ class ImagingExtractor(ABC):
         """
         return self.get_image_shape()
 
-    def get_sample_shape(self) -> Tuple[int, int] | Tuple[int, int, int]:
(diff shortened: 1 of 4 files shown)


h-mayorquin:
@weiglszonja or @alessandratrapani could you help me with a review of this?

src/roiextractors/imagingextractor.py
```diff
@@ -147,11 +112,28 @@ def get_frame_shape(self) -> Tuple[int, int]:
...
+        -------
+        sample_shape: tuple
+            Shape of a single sample from the time series (num_rows, num_columns, num_planes)
+            if volumetric (num_rows, num_columns) otherwise
+        """
+        if self.is_volumetric:
+            return self.get_volume_shape()
+        else:
+            return self.get_frame_shape()
+
```
weiglszonja, line 136: Thanks @h-mayorquin, I think this looks good. In the PR description you mention that `get_sample_shape` returns `get_image_shape` for planar extractor, but it actually returns `get_frame_shape`. 

I probably missed when we added  `get_image_shape` and  `get_frame_shape`. So the idea there is that each imaging extractor would implement their own `get_image_shape` but you could still call the base `get_frame_shape`?

src/roiextractors/imagingextractor.py
```diff
@@ -147,11 +112,28 @@ def get_frame_shape(self) -> Tuple[int, int]:
 
         Notes
         -----
-        This method is equivalent to get_image_shape() and is provided for consistency
-        with the naming convention used in volumetric extractors.
+        This method is equivalent to get_image_shape()
```
alessandratrapani, line 115: Just my curiosity: why do we keep `get_frame_shape()`?

alessandratrapani (approved):
Overall looks good! I don't have any major comments

src/roiextractors/imagingextractor.py
h-mayorquin: I think that this question and @alessandratrapani one below are related.

So, first here https://github.com/catalystneuro/roiextractors/issues/394 we changed `get_image_size` to `get_image_shape`. But then in #406 we changed to use the following terminology: frame for planar, volume for volumetric and sample in general. 

I am not sure how to organize this better. Maybe we should get rid of `get_image_shape` (that we just introduced, I know) to be super consistent? image_shape and frame_shape both seem super clear on their and intent but it would be less code to maintain. Anyway, I made a lot of changes recently and did not have time to push for this. We can move this forward in another PR or leave it as it is.

h-mayorquin:
Thanks for the review!

src/roiextractors/imagingextractor.py
weiglszonja: Yeah it definitely confused me when reviewing, but we can put this in the backlog and come back to it later.

h-mayorquin: Do you think that we should remove `get_image_shape` and leave only `get_frame_shape` to avoid confusions?

PR #424, opened 2025-04-29
Add plane slicing support  in `ScanImageImagingExtractor`

h-mayorquin:
Should come after #419 and has it as base.

Currently, the mechanism for getting a plane extractor in ScanImage is to use a method `get_plane_extractor`. This PR changes the mechanism to a keyword named `plane_inde` at initialization for a couple of reasons:

1) Doing a deep copy of the object is brittle specially with those tifffile_readers around. This could be solved but I don't think it is worth the effort unless we really want/need this.
2) It is more similar to what we have at the moment where the other scan image extractors take a `plane_name` argument but here I am making it an index (I can use name and make it type string if @pauladkisson @weiglszonja and @alessandratrapani prefer it that way).
3) It is overall easier this way.

This PR is mainly tests.


h-mayorquin:
> I think it's worth it to add a test for get_times for this case to check the timestamps behavior with the single-plane case.

I think this is a good idea.

h-mayorquin:
Hey, @pauladkisson I started working on adding full testing (not just for this case) to the `get_times` method here:

https://github.com/catalystneuro/roiextractors/pull/426

I just got an examples of volume data without flyback frames and without multiple samples per slice which I want to add to #426. Could we move this forward and keep working on generalized `get_times` testing on #426?

What do you think?

pauladkisson (approved):

PR #428, opened 2025-05-07
Add support for `ScanImageImagingExtractor` to read data as interleaved samples when multiple samples per depth are available

h-mayorquin:
While considering propagating `ScanImageImagingExtractor` for NeuroConv, I realized that in cases with multiple frames per depth, the current reads the multiple samples per volume as interleaved volumes: Modality A here: https://github.com/catalystneuro/neuroconv/issues/1260.

To maintain backward compatibility I’m adding functionality here to allow reading the data in this way to cause minimal disruption and provide backward compatibility when propagating the changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add `plane_index` to `ScanImageImagingExtractor` to obtain a planar extractor across a plane [PR #424](https://github.com/catalystneuro/roiextractors/pull/424)
 * Add paths as string support to `ScanImageImagingExtractor` [PR #427](https://github.com/catalystneuro/roiextractors/pull/427)
 * Add informative error for old ScanImage files with `ScanImageImagingExtractor` [PR #427](https://github.com/catalystneuro/roiextractors/pull/427)
+* Add the option to read interleaved data in `ScanImageImagingExtractor` [PR #TBD](https://github.com/catalystneuro/roiextractors/pull/TBD)
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


h-mayorquin:
This should enable the workflow in #420 by first this and then plane slicing I think.

pauladkisson (commented):
This looks good overall, and I think it's reasonable to include interleaved samples as an option, but I worry about it being the default path.

We should add a warning if `slice_sample is None and plane_index is None and frames_per_slice > 1`. Interleaving the samples is problematic bc it doesn't preserve acquisition order, and we should encourage users to be careful about that.

h-mayorquin:
> This looks good overall, and I think it's reasonable to include interleaved samples as an option, but I worry about it being the default path.
> 
> We should add a warning if `slice_sample is None and plane_index is None and frames_per_slice > 1`. Interleaving the samples is problematic bc it doesn't preserve acquisition order, and we should encourage users to be careful about that.

I agree with your concern. I actually would like something stronger than a warning. A way in which they can say they want this explicit. Let me think about it, I welcome suggestions.

pauladkisson:
> I actually would like something stronger than a warning. A way in which they can say they want this explicit. Let me think about it, I welcome suggestions.

Maybe an explicit opt-in parameter that is false by default?

h-mayorquin:
@pauladkisson, Ok, I added `interleave_slice_samples` to control this.

pauladkisson (approved):
Nice

h-mayorquin:
Thanks for helping me review all of these PRs, Paul. It is quite a bit of work.

PR #431, opened 2025-05-12
Deprecate `ScanImageTiffImagingExtractor`  

h-mayorquin:
Use `ScanImageLegacyImagingExtractor` 

See this comment:
https://github.com/catalystneuro/roiextractors/issues/413#issuecomment-2867961885

Dev tests will fail because we are still using this is in neuroconv. I will fix this after we release.


pauladkisson (approved):
Looks good.

Because I'm a pedant, I want to point out that ScanImageTiffImagingExtractor was _already_ deprecated (see definitions [here](https://grafana.com/docs/writers-toolkit/write/deprecate-remove/#:~:text=Deprecation%20occurs%20when%20a%20feature,removed%20and%20no%20longer%20supported.)), and now it is being officially removed/renamed.

h-mayorquin:
Yes, that make sense. I changed the changelog to be more precise in thatd direction.

PR #403, opened 2025-03-13
Deprecated write_imaging and write_segmentation

pauladkisson:
This PR deprecates `write_imaging` and `write_segmentation` methods that currently exist and removes versions that are already deprecated/notImplemented.

Cline helped for $2.10 (Claude 3.7 Sonnet w/ Memory Bank)

PR #440, opened 2025-06-06
Add FemtonicsImagingExtractor and related tests for Femtonics data

daphnedequatrebarbes:
Add Femtonics neuroconv : [#951](https://github.com/catalystneuro/roiextractors/compare/roiextractors_femtomics?expand=1)

### DIFF
--- a/src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
+++ b/src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
@@ -1,0 +1,330 @@
+"""A Femtonics imaging extractor with simplified initialization."""
+
+from pathlib import Path
+from typing import Optional, Tuple, Dict, Any, List
+from warnings import warn
+from datetime import datetime
+import xml.etree.ElementTree as ET
+
+import numpy as np
+import h5py
+from lazy_ops import DatasetView
+
+from ...imagingextractor import ImagingExtractor
+from ...extraction_tools import PathType, FloatType, ArrayType
+
+
+class FemtonicsImagingExtractor(ImagingExtractor):
+    """A Femtonics imaging extractor with simplified initialization."""
+
+    extractor_name = "FemtonicsImaging"
+    is_writable = False
+    mode = "file"
+
+    def __init__(
+        self,
+        file_path: PathType,
+        munit: int = 0,
+        channel_name: Optional[str] = None,
+        channel_index: Optional[int] = None,
+    ):
+        """Create a FemtonicsImagingExtractor from a .mesc file.
+
+        Parameters
+        ----------
+        file_path : str or Path
+            Path to the .mesc file.
+        munit : int, optional
+            Index of the measurement unit. Default is 0.
+        channel_name : str, optional
+            Name of channel to extract (e.g., 'UG', 'UR').
+        channel_index : int, optional
+            Index of channel to extract. Default is 0.
+        """
+        super().__init__(file_path=file_path)
+        
+        self.file_path = Path(file_path)
+        self._munit = munit
+        self._channel_name = channel_name
+        self._channel_index = channel_index or 0
+        
+        if self.file_path.suffix != ".mesc":
+            warn("File is not a .mesc file")
+
+        # Open file and setup basic access
+        self._file = h5py.File(file_path, "r")
+        self._setup_channel_selection()
+        self._setup_video_data()
+
+    def _setup_channel_selection(self):
+        """Determine which channel to extract."""
+        available_channels = self.get_available_channels_from_file()
+        
+        if self._channel_name is not None:
+            if self._channel_name not in available_channels:
+                raise ValueError(f"Channel '{self._channel_name}' not found. Available: {available_channels}")
+            self._selected_channel_index = available_channels.index(self._channel_name)
+            self._selected_channel_name = self._channel_name
+        else:
+            if self._channel_index >= len(available_channels):
+                raise ValueError(f"Channel index {self._channel_index} out of range. Available: {available_channels}")
+            self._selected_channel_index = self._channel_index
+            self._selected_channel_name = available_channels[self._channel_index]
+
+    def _setup_video_data(self):
+        """Setup access to the video dataset."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        channel_key = f'Channel_{self._selected_channel_index}'
+        
+        if (session_key in self._file and 
+            munit_key in self._file[session_key] and 
+            channel_key in self._file[session_key][munit_key]):
+            self._video = DatasetView(self._file[session_key][munit_key][channel_key])
+        else:
+            raise Exception(f"Cannot find data at {session_key}/{munit_key}/{channel_key}")
+
+    def get_available_channels_from_file(self) -> List[str]:
+        """Get available channels from the current file."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        
+        if session_key not in self._file or munit_key not in self._file[session_key]:
+            raise ValueError(f"MUnit {self._munit} not found in file")
+            
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        num_channels = attrs.get('VecChannelsSize', 2)
+        
+        channels = []
+        for i in range(num_channels):
+            name_arr = attrs.get(f'Channel_{i}_Name', [])
+            name = self._decode_string(name_arr) if len(name_arr) > 0 else f'Channel_{i}'
+            channels.append(name)
+        return channels
+
+    def _decode_string(self, arr) -> str:
+        """Convert int16 array to string."""
+        if len(arr) == 0:
+            return ""
+        return ''.join(chr(x) for x in arr if x != 0)
+
+    def get_image_shape(self) -> Tuple[int, int]:
+        """Get the shape of the video frame (num_rows, num_columns)."""
+        return self._video.shape[1], self._video.shape[2]  # (height, width)
+
+    def get_num_samples(self) -> int:
+        """Get the number of samples (frames) in the video."""
+        return self._video.shape[0]
+
+    def get_num_channels(self) -> int:
+        """Get the number of channels (always 1 for this extractor)."""
+        return 1
+
+    def get_channel_names(self) -> List[str]:
+        """Get the channel names."""
+        return [self._selected_channel_name]
+
+    def get_sampling_frequency(self) -> float:
+        """Get the sampling frequency in Hz."""
+        time_per_frame_ms = self._get_time_per_frame()
+        if time_per_frame_ms is not None:
+            return 1000.0 / time_per_frame_ms
+        return 30.0  # fallback
+
+    def _get_time_per_frame(self) -> Optional[float]:
+        """Get time per frame from metadata."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        return attrs.get('ZAxisConversionConversionLinearScale')
+
+    def get_dtype(self) -> np.dtype:
+        """Get the data type of the video."""
+        return self._video.dtype
+
+    def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0):
+        """Get specific video frames from indices."""
+        if channel != 0:
+            warn("Femtonics extractor extracts one channel at a time. Channel parameter ignored.")
+            
+        squeeze_data = False
+        if isinstance(frame_idxs, int):
+            squeeze_data = True
+            frame_idxs = [frame_idxs]
+        elif isinstance(frame_idxs, np.ndarray):
+            frame_idxs = frame_idxs.tolist()
+            
+        frames = self._video.lazy_slice[frame_idxs, :, :].dsetread()
+        if squeeze_data:
+            frames = frames.squeeze()
+        return frames
+
+    def get_series(self, start_sample: Optional[int] = None, end_sample: Optional[int] = None) -> np.ndarray:
+        """Get a series of frames."""
+        return self._video.lazy_slice[start_sample:end_sample, :, :].dsetread()
+
+    # Femtonics-specific getter methods
+    def get_pixel_size(self) -> Tuple[float, float]:
+        """Get pixel size in micrometers."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        
+        x_size = attrs.get('XAxisConversionConversionLinearScale', 1.0)
+        y_size = attrs.get('YAxisConversionConversionLinearScale', 1.0)
+        return x_size, y_size
+    
+    def get_image_shape_metadata(self) -> Tuple[int, int, int]:
+        """
+        Get the image shape metadata (X, Y, Z dimensions) from the measurement unit attributes.
+
+        Returns
+        -------
+        shape : tuple
+            (XDim, YDim, ZDim) as (num_columns, num_rows, num_frames)
+        """
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        munit_attrs = dict(self._file[session_key][munit_key].attrs)
+        x_dim = munit_attrs.get('XDim')
+        y_dim = munit_attrs.get('YDim')
+        z_dim = munit_attrs.get('ZDim')
+        return (x_dim, y_dim, z_dim)
+
+    def get_measurement_date(self) -> Optional[datetime]:
+        """Get measurement date."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        
+        posix_time = attrs.get('MeasurementDatePosix')
+        nano_secs = attrs.get('MeasurementDateNanoSecs', 0)
+        
+        if posix_time is not None:
+            return datetime.fromtimestamp(posix_time + nano_secs/1e9)
+        return None
+
+    def get_experimenter_info(self) -> Dict[str, str]:
+        """Get experimenter information."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        
+        return {
+            'username': self._decode_string(attrs.get('ExperimenterUsername', [])),
+            'setup_id': self._decode_string(attrs.get('ExperimenterSetupID', [])),
+            'hostname': self._decode_string(attrs.get('ExperimenterHostname', [])),
+        }
+    
+    def get_geometric_transformations(self) -> Dict[str, np.ndarray]:
+        """
+        Get geometric transformations: translation, rotation, and labeling origin.
+
+        Returns
+        -------
+        dict with keys 'translation', 'rotation', 'labeling_origin'
+        """
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        return {
+            'translation': attrs.get('GeomTransTransl'),
+            'rotation': attrs.get('GeomTransRot'),
+            'labeling_origin': attrs.get('LabelingOriginTransl'),
+        }
+
+
+    def get_mesc_version_info(self) -> Dict[str, Any]:
+        """Get MESc software version information."""
+        session_key = f'MSession_{self._munit}'
+        munit_key = f'MUnit_{self._munit}'
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        
+        return {
+            'version': self._decode_string(attrs.get('CreatingMEScVersion', [])),
+            'revision': attrs.get('CreatingMEScRevision'),
+        }
+
+    def get_pmt_settings(self) -> Dict[str, Dict[str, float]]:
+        """Get PMT settings if available from XML metadata."""
+        try:
+            session_key = f'MSession_{self._munit}'
+            munit_key = f'MUnit_{self._munit}'
+            attrs = dict(self._file[session_key][munit_key].attrs)
+            
+            xml_data = attrs.get('MeasurementParamsXML')
+            if xml_data is None:
+                return {}
+            
+            xml_str = xml_data.decode('latin-1')
+                
+            root = ET.fromstring(xml_str)
+            
+            pmt_settings = {}
+            for gear in root.findall('.//Gear[@type="PMT"]'):
+                try:
+                    channel_name = gear.find('.//param[@name="channel_name"]').get('value')
+                    voltage = float(gear.find('.//param[@name="reference_voltage"]').get('value'))
+                    warmup = float(gear.find('.//param[@name="warmup_time"]').get('value'))
+                    pmt_settings[channel_name] = {'voltage': voltage, 'warmup_time': warmup}
+                except (AttributeError, ValueError):
+                    continue
+            return pmt_settings
+        except Exception:
+            return {}
+
+    def get_scan_parameters(self) -> Dict[str, Any]:
+        """Get scan parameters from XML metadata."""
+        try:
+            session_key = f'MSession_{self._munit}'
+            munit_key = f'MUnit_{self._munit}'
+            attrs = dict(self._file[session_key][munit_key].attrs)
+            
+            xml_data = attrs.get('MeasurementParamsXML')
+            if xml_data is None:
+                return {}
+            
+            xml_str = xml_data.decode('latin-1')
+                
+            root = ET.fromstring(xml_str)
+            
+            scan_params = {}
+            for param in root.findall('.//param'):
+                name = param.get('name')
+                value = param.get('value')
+                if name in ['SizeX', 'SizeY', 'PixelSizeX', 'PixelSizeY', 'Pixelclock']:
+                    try:
+                        scan_params[name] = float(value)
+                    except ValueError:
+                        scan_params[name] = value
+            return scan_params
+        except Exception:
+            return {}
+
+    @staticmethod
+    def get_available_channels(file_path: PathType, munit: int = 0) -> List[str]:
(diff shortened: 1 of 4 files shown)


src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+
+from ...imagingextractor import ImagingExtractor
+from ...extraction_tools import PathType, FloatType, ArrayType
+
+
+class FemtonicsImagingExtractor(ImagingExtractor):
+    """A Femtonics imaging extractor with simplified initialization."""
+
+    extractor_name = "FemtonicsImaging"
+    is_writable = False
```
h-mayorquin, line 21: I think `is_writable` is not needed anymore:

https://github.com/catalystneuro/roiextractors/pull/403

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        channel_name: Optional[str] = None,
+        channel_index: Optional[int] = None,
+    ):
+        """Create a FemtonicsImagingExtractor from a .mesc file.
+
+        Parameters
+        ----------
+        file_path : str or Path
+            Path to the .mesc file.
+        munit : int, optional
```
h-mayorquin, line 37: What is an munit in femtonics?

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+            Index of the measurement unit to extract. The default is 0.
+        channel_name : str, optional
+            Name of the channel to extract (e.g., 'UG', 'UR').
+            If provided, takes precedence over channel_index.
+        channel_index : int, optional
```
h-mayorquin, line 42: I would rather have only one way of selecting channel and we already have `channel_name` in other extractors.  Is there a reason to have two here?

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def get_image_shape(self) -> Tuple[int, int]:
+        """Get the shape of the video frame (num_rows, num_columns)."""
+        return self._video.shape[1], self._video.shape[2]  # (height, width)
+
+    def get_num_samples(self) -> int:
+        """Get the number of samples (frames) in the video."""
+        return self._video.shape[0]
+
+    def get_channel_names(self) -> List[str]:
```
h-mayorquin, line 122: There should be also a static method that allows the user to see which channels are available so they can initialize the extractor. See here for an example:

https://github.com/catalystneuro/roiextractors/blob/afd085ab9d784d3e9c6fb6f2e27f031e803cc6d2/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L699-L720

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+                    try:
+                        scan_params[name] = float(value)
+                    except ValueError:
+                        scan_params[name] = value
+            return scan_params
+        except Exception:
+            return {}
+
+    @staticmethod
+    def get_available_channels(file_path: PathType, munit: int = 0) -> List[str]:
```
h-mayorquin, line 318: Ah, here it is, great.

tests/test_femtonics.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def test_femtonicsimagingextractor_p29_mesc():
+    """
+    Test with Femtonics p29.mesc file.
+
+    File: p29.mesc
+    Metadata (note: metadata vs actual data mismatch):
+    - Source: Femtonics .mesc file (MESc 3.3)
+    - Metadata claims: 56250 samples (frames)
```
h-mayorquin, line 17: Thanks for doing this docstrings. This mismatch that you detected here is my fault. I stubbed the file (i.e. made a smaller version of it) and I changed the frames but did not change the metadata. Thanks for noticing.

tests/test_femtonics.py
```diff
@@ -0,0 +1,149 @@
...
+    This test verifies that the extractor correctly:
+    1. Loads the file and extracts metadata (frame count, resolution, channels, etc.).
+    2. Retrieves the correct channels and properties.
+    3. Tests Femtonics-specific functionality like channel selection and metadata extraction.
+    4. Handles metadata vs actual data mismatches gracefully.
+
+    Note : the actual data only contains 5 frames, not the 56250 claimed in metadata likely due to being truncated for testing.
+    """
+    file_path = OPHYS_DATA_PATH / "imaging_datasets" / "Femtonics" / "moser_lab_mec" / "p29.mesc"
+    extractor = FemtonicsImagingExtractor(file_path=file_path)
```
h-mayorquin, line 36: When there is more than one channel, the extractor should throw an error showing the user the options so they can be deliberate about their choice.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: Oh, you did below. Thanks.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        return attrs.get(
+            "ZAxisConversionConversionLinearScale"
+        )  # ZAxisConversionConversionLinearScale gives the duration of one frame in milliseconds.
+
+    def get_dtype(self) -> np.dtype:
+        """Get the data type of the video."""
+        return self._video.dtype
+
+    def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0):
```
h-mayorquin, line 146: We don't need this method. It is not really necessary and we aim to remove it at some point.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        frames = self._video.lazy_slice[frame_idxs, :, :].dsetread()
+        if squeeze_data:
+            frames = frames.squeeze()
+        return frames
+
+    def get_series(self, start_sample: Optional[int] = None, end_sample: Optional[int] = None) -> np.ndarray:
+        """Get a series of frames."""
+        return self._video.lazy_slice[start_sample:end_sample, :, :].dsetread()
+
+    # Femtonics-specific getter methods
```
h-mayorquin, line 167: Thanks for adding all of these. I think they should be private methods.

h-mayorquin (changes requested):
This looks great. I left some comments. In fact, the biggest problem comes from me here. I did not stub the file properly, thanks for notice. I will fix this and then we can move forward with this.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
daphnedequatrebarbes: In Femtonics MESc files, an MUnit ("Measurement Unit") represents a single imaging acquisition or experiment,
including all associated imaging data and metadata. A single MESc file can contain multiple MUnits,
each corresponding to a separate imaging run/experiment performed during the session.
MUnits are indexed (0, 1, 2, ...) and are stored in the HDF5 hierarchy as /MSession_X/MUnit_0/ ; /MSession_X/MUnit_1/ ...
            
Also added to the init documentation

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
daphnedequatrebarbes: No good reason, I hesitated between the 2. I removed the channel index.

tests/test_femtonics.py
daphnedequatrebarbes: Adapted the  _setup_channel_selection(self):
        - Gets available channels for the selected MUnit.
        - If no channels, raises an error.
        - If channel_name is not given and multiple channels exist, raises an error listing options.
        - If channel_name is given, checks it exists; otherwise, raises an error.
        - If only one channel, selects it automatically.

logic explanation also added to the function documentation

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: Thanks. The documentation addition is great.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        if (
+            session_key in self._file
+            and munit_key in self._file[session_key]
+            and channel_key in self._file[session_key][munit_key]
+        ):
+            self._video = DatasetView(self._file[session_key][munit_key][channel_key])
+        else:
+            raise Exception(f"Cannot find data at {session_key}/{munit_key}/{channel_key}")
+
+    def get_available_channels_from_file(self) -> List[str]:
```
h-mayorquin, line 107:
```suggestion
    def get_available_channels_from_file(self) -> list[str]:
```

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        """Get pixel size in micrometers."""
+        session_key = f"MSession_{self._munit}"
+        munit_key = f"MUnit_{self._munit}"
+        attrs = dict(self._file[session_key][munit_key].attrs)
+
+        x_size = attrs.get("XAxisConversionConversionLinearScale", 1.0)
+        y_size = attrs.get("YAxisConversionConversionLinearScale", 1.0)
+        return x_size, y_size
+
+    def _get_image_shape_metadata(self) -> Tuple[int, int, int]:
```
h-mayorquin, line 190:
```suggestion
    def _get_image_shape_metadata(self) -> tuple[int, int, int]:
```

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        return [self._selected_channel_name]
+
+    def get_sampling_frequency(self) -> float:
+        """Get the sampling frequency in Hz."""
+        time_per_frame_ms = self._get_time_per_frame()
+        if time_per_frame_ms is None:
+            raise ValueError("Sampling frequency could not be determined from metadata.")
+        return 1000.0 / time_per_frame_ms
+
+    def _get_time_per_frame(self) -> Optional[float]:
```
h-mayorquin, line 150:
```suggestion
    def _get_frame_duration_in_milliseconds(self) -> Optional[float]:
```

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+                        scan_params[name] = float(value)
+                    except ValueError:
+                        scan_params[name] = value
+            return scan_params
+        except Exception:
+            return {}
+
+    @staticmethod
+    def get_available_channels(file_path: PathType, munit: int = 0) -> List[str]:
+        """Get available channels in the file."""
```
h-mayorquin, line 319: Add docstring to parameters here. You can copy the one from the `__init__`

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: Also, I think this should be private.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        """
+        session_key = f"MSession_{self._munit}"
+        munit_key = f"MUnit_{self._munit}"
+        munit_attrs = dict(self._file[session_key][munit_key].attrs)
+        x_dim = munit_attrs.get("XDim")
+        y_dim = munit_attrs.get("YDim")
+        z_dim = munit_attrs.get("ZDim")
+        return (x_dim, y_dim, z_dim)
+
+    def _get_measurement_date(self) -> Optional[datetime]:
```
h-mayorquin, line 207:
```suggestion
    def _get_session_start_time(self) -> Optional[datetime]:
```

What do you think of just using nwb centered terminology? the main purpose of this library is to convert data to NWB so I think it is a good idea if we use this terminology.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        return attrs.get(
+            "ZAxisConversionConversionLinearScale"
+        )  # ZAxisConversionConversionLinearScale gives the duration of one frame in milliseconds.
+
+    def get_dtype(self) -> np.dtype:
+        """Get the data type of the video."""
+        return self._video.dtype
+
+    def get_series(self, start_sample: Optional[int] = None, end_sample: Optional[int] = None) -> np.ndarray:
+        """Get a series of frames."""
```
h-mayorquin, line 164:
```suggestion
        """Get a series of samples."""
```

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+        attrs = dict(self._file.attrs)
+        return attrs.get("Uuid")
+
+    def _get_pixel_size(self) -> Tuple[float, float]:
+        """Get pixel size in micrometers."""
+        session_key = f"MSession_{self._munit}"
+        munit_key = f"MUnit_{self._munit}"
+        attrs = dict(self._file[session_key][munit_key].attrs)
+
+        x_size = attrs.get("XAxisConversionConversionLinearScale", 1.0)
```
h-mayorquin, line 186: Actually, don't you need to decode "XAxisConversionUnitName" from the corresponding metadata to see which units this is? How does this work?

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+    def get_available_channels_from_file(self) -> List[str]:
+        """Get available channels from the current file."""
+        session_key = f"MSession_{self._munit}"
+        munit_key = f"MUnit_{self._munit}"
+
+        if session_key not in self._file or munit_key not in self._file[session_key]:
+            raise ValueError(f"MUnit {self._munit} not found in file")
+
+        attrs = dict(self._file[session_key][munit_key].attrs)
+        num_channels = attrs.get("VecChannelsSize", 2)
```
h-mayorquin, line 116: I think this is also in the top level:

alt text

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+                )
+            self._selected_channel_name = available_channels[0]
+        else:
+            if self._channel_name not in available_channels:
+                raise ValueError(f"Channel '{self._channel_name}' not found. Available: {available_channels}")
+            self._selected_channel_name = self._channel_name
+
+    def _setup_video_data(self):
+        """Set up access to the actual imagon data for the selected measurement unit and channel."""
+        session_key = f"MSession_{self._munit}"
```
h-mayorquin, line 90: One big assumption that we are making is that the unit (self._munit) is also the index of the session. That is, you have:

```python
session_key = f"MSession_{self._munit}"
munit_key = f"MUnit_{self._munit}"
```

Maybe this is not the case? This is kind of confusing if m unit is an experimental session, why do we need another top level on the hierarchy?

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: Can you defined MESC here?

h-mayorquin (commented):
This is looking better:

image

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,343 @@
...
+
+    def get_num_samples(self) -> int:
+        """Get the number of samples (frames) in the video."""
+        return self._video.shape[0]
+
+    def get_channel_names(self) -> List[str]:
+        """Get the channel names."""
+        return [self._selected_channel_name]
+
+    def get_sampling_frequency(self) -> float:
```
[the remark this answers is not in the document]
h-mayorquin: Thanks, don't forget to push once you make commits : )

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
daphnedequatrebarbes: Yes ! VecMUnitsSize is at the top level. This represents the total number of measurement units across all sessions in the file.
changed it to : I'm using VecMUnitsSize for initial global bounds checking, while maintaining session-specific validation since units are distributed across sessions. The validation flow I implemented is: Check global bounds → Validate session exists → Check session-specific unit availability.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
daphnedequatrebarbes: You are right! I omitted this scenario.
Changed it to : 
The logic in the code is that the session index  and the measurement unit index [munit_index] are treated as separate, independent indices. When initializing the extractor, the code first uses [session_index] to select the appropriate session group (e.g., MSession_0). Then, within that session, it lists all available measurement units (e.g., MUnit_0, MUnit_1, etc.) and uses [munit_index] to select the specific unit

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,418 @@
...
+                    voltage = float(gear.find('.//param[@name="reference_voltage"]').get("value"))
+                    warmup = float(gear.find('.//param[@name="warmup_time"]').get("value"))
+                    pmt_settings[channel_name] = {"voltage": voltage, "warmup_time": warmup}
+                except (AttributeError, ValueError):
+                    continue
+            return pmt_settings
+        except Exception:
+            return {}
+
+    def _get_scan_parameters(self) -> Dict[str, Any]:
```
h-mayorquin, line 314:
```suggestion
    def _get_scan_parameters(self) -> dict[str, Any]:
```
You can also use it here like this.

tests/test_femtonics.py
```diff
@@ -0,0 +1,292 @@
...
+
+    """
+    file_path = OPHYS_DATA_PATH / "imaging_datasets" / "Femtonics" / "moser_lab_mec" / "p29.mesc"
+
+    # Test available channels static method
+    available_channels = FemtonicsImagingExtractor.get_available_channels(file_path, session_index=0, munit_index=0)
+    assert available_channels == ["UG", "UR"]
+    assert len(available_channels) == 2
+
+    # Test that extractor throws error when multiple channels exist and none specified
```
h-mayorquin, line 40: Good.

tests/test_femtonics.py
```diff
@@ -0,0 +1,292 @@
...
+    assert scan_params["PixelSizeX"] == 1.7821140546875001
+    assert scan_params["PixelSizeY"] == 1.7821140546875001
+    assert scan_params["Pixelclock"] == 8.116736e6
+
+    # Test data type
+    data_type = extractor.get_dtype()
+    assert isinstance(data_type, np.dtype)
+
+    # Test data retrieval with the actual 5 frames
+    if extractor.get_num_samples() > 0:
```
h-mayorquin, line 275: I don't think this if is necessary.

tests/test_femtonics.py
```diff
@@ -0,0 +1,292 @@
...
+        FemtonicsImagingExtractor(file_path=file_path, session_index=0, munit_index=0)
+
+    # Create extractor with explicit channel selection for the rest of the test
+    extractor = FemtonicsImagingExtractor(file_path=file_path, session_index=0, munit_index=0, channel_name="UG")
+
+    assert extractor.get_image_shape() == (512, 512)
+    assert extractor.get_num_samples() == 5
+
+    # Test metadata claims
+    assert extractor._get_image_shape_metadata() == (512, 512, 5)
```
h-mayorquin, line 193: This is failing because the outputs of _get_image_shape_metadata rae numpy scalars. I will add a comment there so you know how to correct them.

tests/test_femtonics.py
```diff
@@ -0,0 +1,292 @@
...
+    assert scan_params["PixelSizeX"] == 1.7821140546875001
+    assert scan_params["PixelSizeY"] == 1.7821140546875001
+    assert scan_params["Pixelclock"] == 8.116736e6
+
+    # Test data type
+    data_type = extractor.get_dtype()
+    assert isinstance(data_type, np.dtype)
+
+    # Test data retrieval with the actual 5 frames
+    if extractor.get_num_samples() > 0:
```
h-mayorquin, line 134: Same, we don't need this if.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: All of this makes sense and are good decisions but I am thinking. I think it would be better to return the  x_size, y_size and the units as  strings. Just provide the metadata as it is. Then we can handle the rest in neuroconv which is where metadata like this actually matters.

 This should be handled in neuroconv where we actually are concerned about the metadata.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,418 @@
...
+    def _get_session_uuid(self):
+        """
+        Get the session UUID from the root attributes.
+
+        Returns
+        -------
+        uuid : array or value
+            The session UUID.
+        """
+        attrs = dict(self._file_handle.attrs)
```
h-mayorquin, line 199: isn't this an array as well that should be decoded?

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,418 @@
...
+
+    def get_channel_names(self) -> list[str]:
+        """Get the channel names."""
+        return [self._selected_channel_name]
+
+    def get_sampling_frequency(self) -> float:
+        """Get the sampling frequency in Hz."""
+        return self._sampling_frequency
+
+    def _get_frame_duration_in_milliseconds(self) -> Optional[float]:
```
h-mayorquin, line 184: Here on the other hand, we do need th check of the `ZAxisConversionUnitName` and throw an error if it is not in milliseconds. It is very important that get the sampling frequency in Hz correclty.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
```diff
@@ -0,0 +1,418 @@
...
+        """
+        Get the image shape metadata (X, Y, Z dimensions) from the measurement unit attributes.
+
+        Returns
+        -------
+        shape : tuple
+            (XDim, YDim, ZDim) as (num_columns, num_rows, num_frames)
+        """
+        munit_attrs = dict(self._file_handle[self._session_key][self._munit_key].attrs)
+        x_dim = munit_attrs.get("XDim")
```
h-mayorquin, line 237:
```suggestion
        x_dim = int(munit_attrs.get("XDim"))
```
Because otherwise you get numpy scalars:

https://numpy.org/doc/stable/reference/arrays.scalars.html

And that is why you test is failing.

h-mayorquin (commented):
This is almost ready to go. One last request, can you add another private method that is called _get_metadata() -> dict

This will return a dictionary with all the metadata that you are able to extract:

metadata_dict = {
"session_info" : _get_session_info(),
"session_start_time": _get_session_start_time(),
"geometric_transofmrations": _get_geometric_transomatiosn()

}

Feel free to organize it as you please, the only reason for this is so, when extracting metadata in neuroconv, we only need to call one method instead of many.

Does this makes sense?

After the tests are passing and we have this I think this should be ready to go.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: And by the rest I mean the warning and, if necessary at some point, conversion.

src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py
h-mayorquin: Thanks for implementing this.

tests/test_femtonics.py
h-mayorquin: And by this I mean the if.

tests/test_femtonics.py
daphnedequatrebarbes: I updated the [_get_image_shape_metadata]method in [FemtonicsImagingExtractor] to always return a tuple of native Python [int] types instead of numpy scalars.

h-mayorquin (approved):
LGTM.

Yes, I also made a mistake uploading the data to S3 and probably made @daphnedequatrebarbes confused for a bit (cc @alessandratrapani )

PR #442, opened 2025-06-06
Remove is writable

h-mayorquin:
@pauladkisson , after we merged https://github.com/catalystneuro/roiextractors/pull/403 I think this attribute is no longer necessary. Also, I am removing some deprecated arguments and eliminating a warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 ### Features
 
+
 ### Fixes
 * Added missing function to retrieve inscopix metadata [PR #436](https://github.com/catalystneuro/roiextractors/pull/436)
+* Fixed deprecated scipy import warning by updating `mat_struct` import path [PR #442](https://github.com/catalystneuro/roiextractors/pull/442)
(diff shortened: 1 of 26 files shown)


pauladkisson (commented):
Some test is failing for Multi Imaging Extractor, otherwise this looks good. 

Do you think this change warrants a new "major" version?

pauladkisson (approved):
Looks good!

> Do you think this change warrants a new "major" version?

Or maybe we wait until removing the rest of the write methods in September?

pauladkisson:
> I don't think we have removed or change behavior other than deprecating, have we? I think that's the logic for a version change. But maybe you have a better pulse, there certainly have been a lot of changes...

There's been at least a couple of changes like this one that might be technically breaking but are not major departures from the way the extractors work. For example if someone were to run the following code:

```python
from roiextractors import CaimanSegmentationExtractor
if CaimanSegmentationExtractor.is_writeable: 
    # do something
```

It would break after this PR is merged. But, it's such a minor detail that shouldn't be used anywhere, so I think it's fine to wait for something bigger to make a "major" release.

PR #445, opened 2025-06-09
add `get_original_frame_indices` to `ScanImageImagingExtractor` for synchronization

h-mayorquin:
The KIM lab has precise synchronized timestamps for every frame derived from TTL signals and needs to map the extractor samples back to their original frame indices in the raw microscopy data for synchronization. See here:

https://github.com/catalystneuro/cohen-u01-to-nwb/blob/b2dbec7de580cc04e748d5b7047a59d5906f55b0/src/kim_lab_to_nwb/kim_conversion.py#L148-L153

This PR adds the `get_original_microscopy_frames_for_samples()` method to `ScanImageImagingExtractor` which returns an array of original frame indices corresponding to each processed samples:

The method is quite simple but most of the work of the PR is in adding test cases for the various possible configurations and experimental setups that the extractor covers.

### DIFF
--- a/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
+++ b/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
@@ -891,6 +891,80 @@ class ScanImageImagingExtractor(ImagingExtractor):
         frames_per_slice = non_varying_frame_metadata.get("SI.hStackManager.framesPerSlice", 1)
         return frames_per_slice
 
+    def get_original_frame_indices(self, plane_index: Optional[int] = None) -> np.ndarray:
+        """
+        Get the original frame indices for each sample.
+
+        Returns the index of the original frame for each sample, mapping processed samples
+        back to their corresponding frames in the raw microscopy data. This accounts for
+        any filtering, subsampling, or exclusions (such as flyback frames) performed by
+        the extractor.
+
+        Parameters
+        ----------
+        plane_index : int, optional
+            Which plane to use for frame index calculation in volumetric data.
+            If None, uses the convention of using the last plane as its the default for
+            volumetric data. It must be less than the total number of planes.
+
+        Returns
+        -------
+        np.ndarray
+            Array of original frame indices (dtype: int64) with length equal to the
+            number of samples. Each element represents the index of the original
+            microscopy frame that corresponds to that sample.
+
+        Notes
+        -----
+        **Frame Index Calculation:**
+
+        - **Planar data**: Frame indices are sequential (0, 1, 2, ...)
+        - **Multi-channel data**: Accounts for channel interleaving
+        - **Volumetric data**: Uses the specified plane (default: last plane)
+        - **Multi-file data**: Includes file offsets for global indexing
+        - **Flyback frames**: Automatically excluded from indexing
+
+        **Common Use Cases:**
+
+        - Synchronizing with external timing systems
+        - Mapping back to original acquisition timestamps
+        - Data provenance and traceability
+        - Cross-referencing with raw data files
+
+        **Examples:**
+
+        For a 3-sample volumetric dataset with 5 planes per volume:
+        - Default behavior returns indices [4, 9, 14] (last plane of each volume)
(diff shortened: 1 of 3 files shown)


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -891,6 +891,80 @@ def get_frames_per_slice(file_path: PathType) -> int:
...
+        - Synchronizing with external timing systems
+        - Mapping back to original acquisition timestamps
+        - Data provenance and traceability
+        - Cross-referencing with raw data files
+
+        **Examples:**
+
+        For a 3-sample volumetric dataset with 5 planes per volume:
+        - Default behavior returns indices [4, 9, 14] (last plane of each volume)
+        - With plane_index=0 returns indices [0, 5, 10] (first plane of each volume)
```
pauladkisson, line 938: This example is a little confusing to me. With plane_index=0, doesn't it return each frame (index) in the first plane?

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -891,6 +891,80 @@ def get_frames_per_slice(file_path: PathType) -> int:
...
+        Returns the index of the original frame for each sample, mapping processed samples
+        back to their corresponding frames in the raw microscopy data. This accounts for
+        any filtering, subsampling, or exclusions (such as flyback frames) performed by
+        the extractor.
+
+        Parameters
+        ----------
+        plane_index : int, optional
+            Which plane to use for frame index calculation in volumetric data.
+            If None, uses the convention of using the last plane as its the default for
```
pauladkisson, line 907: This seems like a bit of a strange default. Is there a reason for this choice?

pauladkisson (commented):
Overall looks good, lots of great tests.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: This is the one that we used for default in volumes.

h-mayorquin: You're right. This isn't clear enough. Let me improve this.

As we've discussed before, the timestamp assigned to each volumetric sample corresponds to the last frame of the volume. That's the behavior currently implemented:

https://github.com/catalystneuro/roiextractors/blob/2a94b760ab4944851136d74d77e0fa9f8d999ba9/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L753-L779

So, when I am extracting the frame that corresponds to a volume sample I am just using the same policy in reverse but also providing flexibility. 

Does that makes sense?

h-mayorquin: I think this reveals that the docstrings should be improved though.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -891,6 +891,80 @@ def get_frames_per_slice(file_path: PathType) -> int:
...
+            volumetric data. It must be less than the total number of planes.
```
pauladkisson, line 908:
```suggestion
            If None, plane_index is set to the last plane in the volume. This is because the timestamp of the acquisition of the last plane in a volume is typically set as the timestamp of the volume as a whole. It must be less than the total number of planes.
```

My stab at making this a bit more clear in the docstring itself.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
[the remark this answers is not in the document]
h-mayorquin: volumetric sample is a better term indeed.

pauladkisson (approved):

PR #447, opened 2025-06-11
Deprecate `frames_to_time` in segmentation extractor

h-mayorquin:
For consistency and in line with the movement to a language center around samples as well.


alessandratrapani (approved):
OK!

PR #465, opened 2025-06-24
Handle native timestamps 

### LINKED ISSUE #448: How to handle native timestamps
Some formats like ScanImage, Miniscope, Minian, and Bruker have native timestamps that we expose in one way or another. While sometimes we overwrite those timestamps by using `set_times` they should be the default timestamps used when writing data to NWB.

In neuroconv, we use the following code to decide whether to write timestamps:

https://github.com/catalystneuro/neuroconv/blob/008e0c70c0b55867889921f4daf140310a46e850/src/neuroconv/tools/roiextractors/roiextractors.py#L439-L457

Crucially, the current code requires that an attribute `_time` is defined; otherwise, it simply returns the indices divided by the sampling frequency:

https://github.com/catalystneuro/roiextractors/blob/2a94b760ab4944851136d74d77e0fa9f8d999ba9/src/roiextractors/imagingextractor.py#L395-L412

I am currently discussing this issue with @alessandratrapani [here](https://github.com/catalystneuro/roiextractors/pull/368#discussion_r2140771955), and I just realized that it's not very clear how to propagate easily and the suggestion in that PR was to override `sample_indices_to_time` which works but it is hard to remember. I also noticed that I haven’t implemented anything similar for the `ScanImageInterface`, which only has a method called `get_times` that must be called explicitly:

https://github.com/catalystneuro/roiextractors/blob/2a94b760ab4944851136d74d77e0fa9f8d999ba9/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L752-L768

That is, I forgot to propagate those times to `sample_indices_to_time`.

I think we should ensure that native timestamps are called by default. One option would be to call them during `__init__`, but I think that’s a bad idea because it would slow down initialization and make the object heavier than necessary.

Instead, I propose we create another abstract method called `get_native_timestamps`, which would be invoked by the existing `sample_indices_to_time` method. This way, native timestamps would be used automatically if they exist. Whoever is implementing a new interface can add an empty method that returns `None` from this function if native timestamps aren’t available, but this approach forces them to be intentional about it.

Then the logic will look like this: the behavior becomes automatic, and we no longer have to remember to add it manually, since it will be enforced by the abstract method.

```python
    def sample_indices_to_time(self, sample_indices: Union[FloatType, np.ndarray]) -> Union[FloatType, np.ndarray]:
        """Convert user-inputted sample indices to times with units of seconds.

        Parameters
        ----------
        sample_indices: int or array-like
            The sample indices to be converted to times.

        Returns
        -------
        times: float or array-like
            The corresponding times in seconds.
        """
        # Default implementation
        
        # Always use cached timestamps if available
        if self._times is not None:
            return self._times[sample_indices]
        
        native_timestamps = self.get_native_timestamp()
        if native_timestamps is not None:
            self._times = native_timestamps
            return native_timestamps[sample_indices]
        else:
            return sample_indices / self.get_sampling_frequency()
            
```

What do you guys think @alessandratrapani @pauladkisson @weiglszonja

h-mayorquin:
Should close #448


src/roiextractors/imagingextractor.py
```diff
@@ -394,6 +394,66 @@ def get_frames(self, frame_idxs: ArrayType, channel: Optional[int] = 0) -> np.nd
...
+
+        # Try to get original timestamps
+        original_timestamps = self.get_original_timestamps()
+        if original_timestamps is not None:
+            self._times = original_timestamps
+            return original_timestamps[start_sample:end_sample]
+
+        # Fallback to calculated timestamps from sampling frequency
+        sample_indices = np.arange(start_sample, end_sample)
+        return sample_indices / self.get_sampling_frequency()
```
pauladkisson, line 455: Why not include this in get_original_timestamps? Then we wouldn't have to return None

h-mayorquin: That would be more convenient, but it's not the semantic intent I had in mind. `get_original_timestamps` is a provenance function that answers the question: "Does the format have native timestamps?" If it does, the function returns them; if not, it returns `None`.

In contrast, `get_timestamps` is the convenience function that provides that behavior, it always return timestamps. You can always call it, while `get_original_timestamps` serves as the underlying provenance function.

How are you thinking about it?

pauladkisson (approved):
Looks good.

src/roiextractors/imagingextractor.py
h-mayorquin: After thinking about it, there's an additional complication. `get_original_timestamps` doesn't serve the same purpose here as it does in Neuroconv. In neuroconv, `get_original_timestamps` returns the timestamps at initialization (which can be frame-based) and doesn't indicate anything about provenance. Basically, it returns the timestamps before the user modifies or shifts them.

So maybe we should go with `native_timestamps` after all, to avoid a conflict with neuroconv. What do you think?

pauladkisson: > In neuroconv, get_original_timestamps returns the timestamps at initialization (which can be frame-based) and doesn't indicate anything about provenance. Basically, it returns the timestamps before the user modifies or shifts them.

Yeah, I think this is what I was remembering, but I was having a hard time articulating it.

> So maybe we should go with native_timestamps after all, to avoid a conflict with neuroconv. What do you think?

Do we need this sort of provenance? What difference does it make whether the timestamps were recorded by the acquisition system or implicitly generated based on the sampling rate?

h-mayorquin: In neuroconv, we currently have logic for deciding whether to write the sampling rate or timestamps:

https://github.com/catalystneuro/neuroconv/blob/ff334641815286fae399e2d7c0c1bc9586d1c3b3/src/neuroconv/tools/roiextractors/roiextractors.py#L439-L456

At the moment, we determine this based the `has_time_vector` method but I don't like this too much. `has_time_vector` only reflects whether timestamps have been cached either because the user set them manually or they were accessed previously and does not check whether native timestamps exist for the extractor.

I think we should instead have the following logic:

1) If timestamps are cached, use them to check if the timestamps are regular.
2) If not cached, check if the format provides native timestamps and do the check with them.
3) If there are neither cached nor native timestamps, then just write sampling rate.

Currently, we are not really doing step 2.

Without a method to check for native timestamp we are left with two choices:

* Option 1 (current behavior): Skip native timestamps entirely if not cached.
* Option 2 (wasteful in some cases): Always generate a time vector (by calling native timestamps or the implicitly generated ones) which generates unnecessary memory usage.

So, having an option to get the native timestamps allows to implement a better and more performant logic and in general I think it is better compartmentalization of method purpose.

h-mayorquin: Yes, it is bad there as well. It is even worse because ecephys timestamps are way longer and therefore larger size than in ophys.

h-mayorquin: It’s not clear to me whether you’re opposed to adding a method for provenance. Let me summarize my view so we can see if there’s any disagreement and try to move forward:

- `get_timestamps`: This should be the main method to use. It always returns timestamps, whether they are native, cached, or implicit (generated from the sampling frequency).
- `get_native_timestamps` (currently called `get_original_timestamps` here): This method can be used by extractor instances when the format includes native timestamps so they are propagated in an automated way with clear semantics.
- `get_original_timestamps` (used only in NeuroConv): This is used to re-instantiate an extractor and to remove shifts or previously set timestamps. I don’t think this method is needed at the extractor level here at roiextractors.

Regardless, one thing we should definitely do is resolve the name conflict, since the current `get_original_timestamps` does something different in NeuroConv.

What do you think?

pauladkisson: Yep, that all sounds good. I just wanted to make sure that we had good reasons for these changes.

PR #467, opened 2025-06-25
Add properties to segmentation extractor

### LINKED ISSUE #462: Add properties to SegmentationExtractor
SpikeInterface uses the concept of extractor properties: a dictionary where the keys are property names and the values are vectors, each with a length equal to the number of neurons. We use this in NeuroConv to automatically propagate properties from the extractors to neuroconv :

https://github.com/catalystneuro/neuroconv/blob/ae2fb24432bc4e15e660e287128f0f26dbba01f9/src/neuroconv/tools/spikeinterface/spikeinterface.py#L810-L832

However, for segmentation extraction, we’ve taken a different approach. Instead of using general properties, we define specific keyword for each type (e.g., `roi_centroids` and `roi_acceptance`):

https://github.com/catalystneuro/neuroconv/blob/ae2fb24432bc4e15e660e287128f0f26dbba01f9/src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py#L205-L206

This introduces a limitation: when additional properties per ROI need to be extracted, there is no straightforward way to do so. Here is an example with quality metrics.

https://github.com/catalystneuro/roiextractors/issues/461

I think it would be better to propagate these quantities as properties, similar to how it’s done with SpikeInterface. This approach has several advantages:

* We don't need to introduce a new keyword into the ophys pipeline every time we want to propagate a new property.
* It allows for customization: users can add custom properties to the `PlaneSegmentation` table using the same workflow they use to add custom properties to the Units table (see: https://neuroconv.readthedocs.io/en/stable/how_to/annotate_ecephys_data.html#how-to-add-custom-properties-to-the-electrodes-table).
* It avoids the need to override add_to_nwbfile just to selectively disable specific quantity keywords. If the segmentation extractor contains the property, it will be added; if not, it won’t be. This logic can be automated, reducing manual work and errors in our side. See here for an instance of this situation: https://github.com/catalystneuro/neuroconv/pull/1107/files#r2138783965

Thoughts? Drawbacks? @alessandratrapani @pauladkisson @weiglszonja

h-mayorquin:
Should fix #462.

### DIFF
--- a/tests/test_internals/test_properties.py
+++ b/tests/test_internals/test_properties.py
@@ -51,7 +51,7 @@ def test_multiple_properties_and_types():
     assert np.array_equal(extractor.get_property(key="active", ids=roi_ids), active)
 
 
-def test_property_validation():
+def test_setter_property_validation():
     """Test property validation."""
     extractor = generate_dummy_segmentation_extractor(num_rois=5, num_frames=10)
     roi_ids = extractor.get_roi_ids()
@@ -79,7 +79,7 @@ def test_property_validation():
         extractor.set_property(key="quality", values=[0.8, 0.9, 0.7, 0.5, 0.4], ids=mixed_ids)
 
 
-def test_empty_properties():
+def test_getter_property_validation():
     """Test mismatch in property name."""
     extractor = generate_dummy_segmentation_extractor(num_rois=3, num_frames=10)
 


alessandratrapani (approved):
That is great, thanks Heberto!

PR #484, opened 2025-08-05
Remove num channels

h-mayorquin:
Remove `get_num_channels` from imaging extractors. We deprecated this. Now is the time to remove it.

### DIFF
--- a/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
+++ b/src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
@@ -232,9 +232,6 @@ class Hdf5ImagingExtractor(ImagingExtractor):
     def get_channel_names(self):
         return self._channel_names
 
-    def get_num_channels(self):
-        return self._num_channels
-
     def get_native_timestamps(
         self, start_sample: Optional[int] = None, end_sample: Optional[int] = None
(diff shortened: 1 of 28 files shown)


src/roiextractors/extractors/hdf5imagingextractor/hdf5imagingextractor.py
```diff
@@ -232,9 +232,6 @@ def get_sampling_frequency(self):
     def get_channel_names(self):
```
alessandratrapani, line 232: Are we going to remove get_channel_names as well?

src/roiextractors/extractors/memmapextractors/memmapextractors.py
```diff
@@ -226,7 +218,7 @@ def write_imaging(
             buffer_size_in_bytes = int(buffer_size_in_bytes)
             type_size = np.dtype(dtype).itemsize
 
-            n_channels = imaging.get_num_channels()
+            n_channels = 1
```
alessandratrapani, line 221: Why do we need to hard-code n_channels = 1?

h-mayorquin: That extractor creates a structure that includes a channel at the moment. We should change this, but this will make the PR larger and harder to review. I will do it once I find some time.

PR #479, opened 2025-07-08
Refactor Femtonics extractor session and munit logic

daphnedequatrebarbes:
Fix #475

PR #485, opened 2025-08-20
Post-Release 0.6.2

pauladkisson:


h-mayorquin (approved):
I am confused, the feature that I wanted was this:

https://github.com/catalystneuro/roiextractors/pull/479/files

And as far as I understand, it was not released yet.

Maybe the changelog entry was added to the wrong place when we merged? 

In that case, my bad.

PR #494, opened 2025-09-23
Add has_time_vector method to SampleSlicedSegmentationExtractor

alessandratrapani:
Override `has_time_vector` method to check for the presence of a time vector in the parent segmentation extractor.
When defining a `has_time_vector` method in an ad hoc segmentation extractor, tests in `Test[...]SegmentationInterfaceWithStubTest` fail. 
This occurs because `Test[...]SegmentationInterfaceWithStubTest` uses `SampleSlicedSegmentationExtractor`, which defaults to the base `SegmentationExtractor.has_time_vector` implementation rather than the actual implementation of the parent segmentation extractor. For custom implementations like `MinianSegmentationExtractor`, this causes test failures https://github.com/catalystneuro/neuroconv/pull/1107.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Deprecations And Removals
 * Remove deprecated `get_num_channels()` method from `ImagingExtractor` [PR #484](https://github.com/catalystneuro/roiextractors/pull/484)
 * Remove deprecated `write_segmentation()` methods from all segmentation extractors [PR #491](https://github.com/catalystneuro/roiextractors/pull/491)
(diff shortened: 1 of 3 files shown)


h-mayorquin (changes requested):
This LGTM.

Can you add tests and this should be good to go.

h-mayorquin:
> > Why can't sample sliced segmentation extractor just use the parent's _times?
> 
> Because those might be modified by `set_times` and then sliced and the parent extractor should not be coupled in that way.

@pauladkisson I am wrong about this as well. Check:
https://github.com/catalystneuro/roiextractors/pull/498

Basically, because `set_times` sets a completely new self._times  attribute this problems never appears. 

Thanks.

PR #497, opened 2025-09-24
Add optional sampling frequency and timestamps support to MinianSegmentationExtractor

### LINKED ISSUE #496: The `MinianSegmentationExtractor` should accept an optional sampling frequency and optional path to a `timeStamps.csv`
The error when running on main is:

```
  if segmentation_extractor.has_time_vector():
      timestamps = segmentation_extractor.get_timestamps()
      estimated_rate = calculate_regular_series_rate(series=timestamps)
      if estimated_rate:
          roi_response_series_kwargs.update(starting_time=timestamps[0], rate=estimated_rate)
      else:
          roi_response_series_kwargs.update(timestamps=timestamps, rate=None)
  else:
>           rate = float(segmentation_extractor.get_sampling_frequency())
E           TypeError: float() argument must be a string or a real number, not 'NoneType'
```

This breaks the `SegmentationExtractor` contract, which assumes that a sampling frequency is always available.

Looking at the [MinianSegmentationExtractor](https://github.com/catalystneuro/roiextractors/blob/5b3971e5d0de7b1a2fe715a9c728da2ee1770597/src/roiextractors/extractors/minian/miniansegmentationextractor.py#L21), no sampling frequency is set. A search of the Minian repository also does not reveal a sampling rate or frequency field. Can you confirm if this is the case? that the sampling frequency is not really available in the source format.

We could work with timestamps only but the [Minian `save_minian`](https://github.com/denisecailab/minian/blob/f64c456ca027200e19cf40a80f0596106918fd09/minian/utilities.py#L440) function does not mention `timeStamps.csv`, and the documentation provides no reference either. The `timeStamps.csv` files we currently have in our dataset appear to have been manually added by the original authors, likely copied from Miniscope outputs (which makes total sense). We should not assume future users will be as careful as Zaki.

For writing to NWB we need to ensure that we have either the timestamps and/or the sampling frequency. Besides solving [the slicing of the timestamps issue](https://github.com/catalystneuro/roiextractors/pull/494) I propose the following improvements:

1. The `MinianSegmentationExtractor` should accept an **optional sampling frequency** at construction.  
2. The `MinianSegmentationExtractor` should accept an **optional path to a `timeStamps.csv` file** at construction. If not provided, assume the default location as in the Zaki example dataset.  
3. If neither a sampling frequency nor a valid `timeStamps.csv` is provided, the extractor should raise an error at construction with a clear message instructing the user to supply a sampling frequency.  
4. The `get_sampling_frequency()` method should raise an **informative error** when no sampling frequency is available, instead of returning `None`.  

This will ensure that we always have the information necessary to write to NWB and provide a better user experience when we don't. At the same time, this is flexible enough so we won't require future modifications for different users.

What do you think?

_Originally posted by @h-mayorquin in https://github.com/catalystneuro/neuroconv/issues/1107#issuecomment-3324822660_

alessandratrapani:
Fixes #496
Introduce optional parameters for sampling frequency and timestamps path in the MinianSegmentationExtractor. Implement validation for these parameters and enhance error handling. Add tests to ensure correct functionality and error scenarios.

### DIFF
--- a/src/roiextractors/extractors/minian/miniansegmentationextractor.py
+++ b/src/roiextractors/extractors/minian/miniansegmentationextractor.py
@@ -13,6 +13,7 @@ from typing import Optional
 import numpy as np
 import pandas as pd
 import zarr
+from neuroconv.utils import calculate_regular_series_rate
 
 from ...extraction_tools import FloatType, PathType
 from ...segmentationextractor import SegmentationExtractor
@@ -48,7 +49,12 @@ class MinianSegmentationExtractor(SegmentationExtractor):
 
     extractor_name = "MinianSegmentation"
 
-    def __init__(self, folder_path: PathType, sampling_frequency: Optional[float] = None):
+    def __init__(
+        self,
+        folder_path: PathType,
+        sampling_frequency: Optional[float] = None,
+        timestamps_path: Optional[PathType] = None,
+    ):
         """Initialize a MinianSegmentationExtractor instance.
 
         Parameters
@@ -57,10 +63,15 @@ class MinianSegmentationExtractor(SegmentationExtractor):
             The location of the folder containing minian .zarr output.
         sampling_frequency: float, optional
             The sampling frequency in Hz. If not provided, will attempt to derive from timeStamps.csv.
+        timestamps_path: str or Path, optional
+            Path to the timeStamps.csv file. If not provided, assumes default location at folder_path/timeStamps.csv.
         """
         SegmentationExtractor.__init__(self)
         self.folder_path = Path(folder_path)
         self._sampling_frequency = sampling_frequency
+        self._timestamps_path = (
+            Path(timestamps_path) if timestamps_path is not None else self.folder_path / "timeStamps.csv"
+        )
         self._roi_response_denoised = self._read_trace_from_zarr_field(field="C")
         self._roi_response_baseline = self._read_trace_from_zarr_field(field="b0")
         self._roi_response_neuropil = self._read_trace_from_zarr_field(field="f")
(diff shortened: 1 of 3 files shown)


src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -326,6 +348,42 @@ def get_images_dict(self) -> dict:
...
+
+        # If no sampling frequency provided, try to derive from timestamps
+        if self._timestamps_path.exists():
+            try:
+                # Try to calculate sampling frequency from timestamps
+                timestamps = self.get_native_timestamps()
+                if len(timestamps) > 1:
+                    # Calculate average sampling frequency from time differences
+                    derived_freq = calculate_regular_series_rate(timestamps)
+                    return float(
```
alessandratrapani, line 375: add warning for None sampling freq

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -326,6 +348,42 @@ def get_images_dict(self) -> dict:
...
+                        derived_freq
+                    )  # it will be None if timestamps are irregularly spaced, so _times will be set in get_native_timestamps
+            except Exception:
+                pass
+
+        raise ValueError(
+            "No sampling frequency available. Please provide a sampling_frequency parameter "
```
alessandratrapani, line 382: move the error on the valide timestamps.csv file on construction

alessandratrapani:
I am not sure why test are failing

h-mayorquin:
Neuroconv is not a dependency of this library, that's why the tests are failing. In these cases, I am in for just duplicating the code and having it as well but I don't know how @pauladkisson wants to handle this.

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -13,6 +13,7 @@
 import numpy as np
 import pandas as pd
 import zarr
+from neuroconv.utils import calculate_regular_series_rate
```
h-mayorquin, line 16: It should be imported within the extractor and not at the top level anyway. 

Consider doing the same for zarr just for the sake of import time efficiency (so all of those things are not front loaded when someone does import roiextractors)

pauladkisson:
> Neuroconv is not a dependency of this library, that's why the tests are failing. In these cases, I am in for just duplicating the code and having it as well but I don't know how @pauladkisson wants to handle this.

Yes, I agree. Calculating the regular series rate is a tiny function; it shouldn't be a big deal to duplicate here. I'd like to avoid making Neuroconv a dependency of roixtractors since then it becomes circular and we'd have to be quite careful to avoid circular dependency issues.

h-mayorquin (approved):
LGTM

PR #500, opened 2025-09-26
Add Tuple -> tuple and Union[x, y] to x | y ruff rule

h-mayorquin:


pauladkisson:
> This looks good. I'd love to add the ruff-rule for Optional as well. Do you know if one exists?

Looks like there is. See here: https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/

PR #501, opened 2025-09-26
Updated version to 0.7.0 pre-release

pauladkisson:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "roiextractors"
-version = "0.6.3"
+version = "0.7.0"
```
h-mayorquin, line 7: actually, let's drop the `write_imaging` as well whose times has come so you avoid the change of version after we remove those.

pauladkisson: Sounds good, can you take care of that?

h-mayorquin: https://github.com/catalystneuro/roiextractors/pull/502

h-mayorquin (approved):

PR #502, opened 2025-09-26
Remove write method on imaging extractors

h-mayorquin:


h-mayorquin:
> Looks great, feels good to be cleaning out some of this old deprecated junk.

And the deprecaton of October is just around the corner!

PR #505, opened 2025-09-27
Improve Segmentation Extractor Roi Response Representation

### LINKED ISSUE #492: Improve Segmentation Extractor Roi Response Representation
I’ve been thinking about [#375](https://github.com/catalystneuro/roiextractors/issues/375), and I believe there’s a more general solution that would both address that issue and improve the code base.

The purpose of this library is to extract data that can be standardized to the NWB format. From that perspective, what matters is mapping ROI responses to specific ROIs. Currently, we hard-code traces like this:

https://github.com/catalystneuro/roiextractors/blob/e7eb4fa2fb3021bcc3ba210f32adf53dddb0c1c9/src/roiextractors/segmentationextractor.py#L42-L45

In each segmentation extractor, we set these variables and implicitly assume they map to all ROIs except “background.” This has worked until now, but #375 highlights some problems:

1. Terminology mismatch:  Each algorithm uses its own naming conventions, which we currently have to shoehorn into the fixed, hard-coded trace names in the extractors.
2. Conceptual differences: Even when the same term is used, it may refer to different model components. For example, in one algorithm “neuropil” may be defined as a residual trace tied to a specific ROI, while in another it may represent an independent decomposition of a global residual term with its own ROI set. The same applies to “background.”

From the NWB perspective, these naming conventions should not matter. What matters is that each format provides specific traces and the ROI IDs they correspond to. As long as we preserve that mapping, we are free to label the response types in a way that is consistent with the source format while still ensuring they are correctly tied to their ROIs. If a format defines background or neuropil ROIs that differ from the “neural” ones (e.g., global background traces), we can simply assign extra ROI IDs to capture those cases.

In this direction, a more robust solution would be to represent responses explicitly, for example:

```python
@dataclass
class RoiResponse:
    """Represents a fluorescence response (trace) with its metadata."""

    response_type: str # e.g. "raw", "dff", "neuropil", "baseline", "background", "denoised", "deconvolved"
    data: ArrayLike  # Shape: (num_samples, num_rois)
    roi_ids: list[str | int]  # Which ROIs this trace corresponds to
```

and then on each extractor we collect these objects like this and map the ROIs explicitly.

```python
raw_data = extract_raw_data(**kwargs)
roi_ids = # the list of ROI IDs corresponding to the data
self._roi_responses.append(RoiResponse("raw", raw_data, roi_ids))

denoised_data = extract_denoised_data(**kwargs)
roi_ids = # usually same roi_ids as raw_data
self._roi_responses.append(RoiResponse("denoised", denoised_data,

background_data = extract_background_data(**kwargs)
background_ids = # can be the same ids as the raw ids or different
self._roi_responses.append(RoiResponse("background", background_data, background_ids

neuropil_data = extract_neurpil_data("f")
neuropil_ids = # can be the same ids as the raw ids or different
self._roi_responses.append(RoiResponse("neuropil", neuropil_data, neuropil_ids))
```

This has several advantages:
* **Explicit**: Each trace is explicitly tied to the ROI IDs provided by the format. We avoid relying on implicit assumptions about which traces map to which ROIs. We set them once correctly on every format and then handle them downstream as we wish.

* **Flexible**: We no longer need special handling for background or neuropil, nor do we need to hard-code names for every segmentation extractor. The only requirement is to store traces alongside the ROI IDs as provided by the source and the name is free form.

* **Future-proof**: This approach is less likely to break with future non-CNMF formats. As long as there are traces and corresponding ROI mappings (a requirement for NWB), they can be stored with the same data model.

Thoughts, @alessandratrapani @pauladkisson @weiglszonja ?

h-mayorquin:
Should close #492

### DIFF
--- a/src/roiextractors/extractors/caiman/caimansegmentationextractor.py
+++ b/src/roiextractors/extractors/caiman/caimansegmentationextractor.py
@@ -14,7 +14,7 @@ import numpy as np
 from scipy.sparse import csc_matrix
 
 from ...extraction_tools import PathType, get_package
-from ...segmentationextractor import SegmentationExtractor
+from ...segmentationextractor import RoiResponse, SegmentationExtractor
 
 
 class CaimanSegmentationExtractor(SegmentationExtractor):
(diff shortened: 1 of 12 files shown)


src/roiextractors/extractors/caiman/caimansegmentationextractor.py
```diff
@@ -140,11 +140,39 @@ def __init__(self, file_path: PathType):
...
+                cell_ids = list(range(denoised_traces.shape[1]))
+            self._roi_responses.append(RoiResponse("denoised", denoised_traces, cell_ids))
+
+        deconvolved_traces = self._trace_extractor_read("S")
+        if deconvolved_traces is not None:
+            if cell_ids is None:
+                cell_ids = list(range(deconvolved_traces.shape[1]))
+            self._roi_responses.append(RoiResponse("deconvolved", deconvolved_traces, cell_ids))
+
+        neuropil_traces = self._trace_extractor_read("f")
```
alessandratrapani, line 168: I would call it background, not neuropi,l to be consistent with Caiman naming

src/roiextractors/extractors/minian/miniansegmentationextractor.py
```diff
@@ -71,28 +71,46 @@ def __init__(
...
+
+        denoised_traces = self._read_trace_from_zarr_field(field="C")
+        if denoised_traces is not None:
+            self._roi_responses.append(RoiResponse("denoised", denoised_traces, list(self._roi_ids)))
+
+        baseline_traces = self._read_trace_from_zarr_field(field="b0")
+        if baseline_traces is not None:
+            self._roi_responses.append(RoiResponse("baseline", baseline_traces, list(self._roi_ids)))
+
+        neuropil_traces = self._read_trace_from_zarr_field(field="f")
```
alessandratrapani, line 87: Again background not neuropil

alessandratrapani (approved):
Just a couple of questions, the rest looks good, and leaves room to generalize for different segmentation algorithms

PR #507, opened 2025-09-29
Optional -> None as a ruff rule

h-mayorquin:
Suggested by @pauladkisson in a previous PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,7 +110,8 @@ select = [
     "F401",  # Unused import
     "I",  # All isort rules
     "UP006", # non-pep585 annotation (tuple, list -> tuple, list)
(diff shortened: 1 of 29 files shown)


h-mayorquin:
> Looks pretty good overall. I noticed that there are still some lingering optionals in the docstrings of a few functions: Ex. inscopiximagingextractor.py Line 83:
> 
> https://github.com/catalystneuro/roiextractors/blob/0cb2cb7f2f0ffba7f24fa12f0b1bba0996e5906c/src/roiextractors/extractors/inscopixextractors/inscopiximagingextractor.py#L83
> 
> Also, I think this PR warrants a changelog entry.

Ok, I will do this after: https://github.com/catalystneuro/roiextractors/pull/505 

as I think the conflicts will be easier that way.

PR #518, opened 2025-10-22
Update macos and python version for testing

h-mayorquin:
I am also switching the testing to only test the oldest and newest python version as we do in neuroconv. @pauladkisson , what do you think?


pauladkisson (approved):
Makes sense to me, I just updated the branch protection rules accordingly.

PR #520, opened 2025-10-23
Fix lingering references to _image_masks

h-mayorquin:
See changelog. We left lingering references to the removed _image_masks but we were not testing for them so the CI did not catch them.


h-mayorquin:
I think that's nice as extra safety but the main culprit here is the lack of testing in this library.

PR #524, opened 2025-11-07
Depreca-thon November 2025: Spring Cleaning in Autumn

h-mayorquin:
A lot of functionality whose times has come. See the changelog.

@pauladkisson this is a large PR. If you would prefer let me know and I will deprecate things bit by bit and we can use this as a reference. 

As you prefer.

### DIFF
--- a/tests/test_volumetricimagingextractor.py
+++ b/tests/test_volumetricimagingextractor.py
@@ -8,12 +8,8 @@ num_samples = 10
 
 
 @pytest.fixture(scope="module")
-def imaging_extractors():
-    return [generate_dummy_imaging_extractor(num_samples=num_samples) for _ in range(3)]
-
-
-@pytest.fixture(scope="module")
-def volumetric_imaging_extractor(imaging_extractors):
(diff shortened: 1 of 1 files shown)


h-mayorquin:
The dev tests are failing because we have not done https://github.com/catalystneuro/neuroconv/pull/1576

tests/test_volumetricimagingextractor.py
```diff
@@ -9,8 +9,8 @@
 
 @pytest.fixture(scope="module", params=[1, 2])
 def imaging_extractors(request):
-    num_channels = request.param
-    return [generate_dummy_imaging_extractor(num_channels=num_channels, num_samples=num_samples) for _ in range(3)]
+    # num_channels parameter is deprecated and removed - all extractors are single channel now
```
pauladkisson: If we are removing `num_channels`, this test fixture probably should not have an unused parameter that references `num_channels`.

pauladkisson (approved):

PR #529, opened 2025-11-18
Add `get_native_timestamps` functionality to nwb extractor

h-mayorquin:
As in the title, plus, other two changes:
1) The method `make_nwb_metadata` is marked to be removed. This was form the time that we were using this library to write to NWB but is no longer necessary.
2) The tests were switched to use pytest and the mocks from the pynwb library.


pauladkisson (approved):
Looks good!

PR #533, opened 2025-12-01
Handle `get_accepted_list` with segmentation extractor properties

h-mayorquin:
The current `get_accepted_list()` and `get_rejected_list()` abstract methods were heavily modeled after Suite2p's and I think CaImAn butd don't generalize well across different segmentation formats. Formats like Inscopix uses string status values, and formats like SIMA or CNMF-E don't have this concept at all.

Previously, [we introduced the property system](https://github.com/catalystneuro/roiextractors/pull/467) that allows flexible per-ROI metadata storage. This PR proposes using properties for ROI classification instead of abstract methods, offering several advantages: (1) we preserve provenance by keeping format-native terminology (e.g., `iscell` for Suite2p, `cell_status` for Inscopix), (2) formats without acceptance concepts (like Minian or SIMA) no longer need placeholder implementations that return meaningless defaults, and (3) the approach generalizes to non-binary cases: for example if the format stores a confidence index instead of a binary classification. 

In short, the property-based approach uses existing infrastructure, better respects source format provenance, and scales to cases where binary accepted/rejected doesn't apply.

Here a table of the property that would replace the is_accepted
| Format | Property Name | Values | Notes |
|--------|---------------|--------|-------|
| Suite2p | `iscell` | 0/1 | Native binary classification |
| CaImAn | `is_accepted` | bool | Derived from `idx_components` |
| Inscopix | `cell_status` | "accepted"/"rejected" | Native string enum |
| NewExtract | (dynamic) | - | Computed from mask content |
| Numpy | `is_accepted`, `is_rejected` | bool | If provided by user |
| SIMA, CNMF-E, Minian, LegacyExtract | (none) | - | Use base class default |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Bruker series can now read sequences of type `TSeries Brightness Over Time Element` [PR #527](https://github.com/catalystneuro/roiextractors/pull/527)
 
 ### Deprecations And Removals
+* Deprecated `get_accepted_list()` and `get_rejected_list()` methods (will be removed in May 2026). ROI classification is now stored as properties using format-native nomenclature. Use `get_property()` instead to access format-specific acceptance data. [PR #536](https://github.com/catalystneuro/roiextractors/pull/536)
 * Deprecated `get_channel_names()` method across all extractors (will be removed on or after May 2026). The method is no longer abstract and provides a default implementation returning `["channel_0", "channel_1", ...]` before removal. [PR #530](https://github.com/catalystneuro/roiextractors/pull/530)
 * Removed deprecated `get_image_size()` method from `ImagingExtractor`, `MultiImagingExtractor`, and `VolumetricImagingExtractor` (deprecated September 2025). Use `get_image_shape()` instead. [PR #372](https://github.com/catalystneuro/roiextractors/pull/372)
 * Removed deprecated `get_num_frames()` method from `ImagingExtractor`, `MultiImagingExtractor`, and `VolumetricImagingExtractor` (deprecated September 2025). Use `get_num_samples()` instead. [PR #372](https://github.com/catalystneuro/roiextractors/pull/372)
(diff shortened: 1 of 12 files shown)


h-mayorquin:
I realize now that I was probably using the term *provenance* in my earlier comment very [loosely and maybe even wrongly](https://en.wikipedia.org/wiki/Provenance#Data_provenance)

What I *meant* is the following principle that I do want to defend:

**When automatically writing data to NWB, we should avoid modifying or reinterpreting the source data unless there is a strong, explicit justification.**

Some context for my comment as I want to share this with other people. We are discussing how to store ROI-quality information. I think this is similar to the discussion of quality metrics in the spike sorting side. In the ophys side, [neuroconv](https://github.com/catalystneuro/neuroconv/blob/010703186ebedbaeb27048003d9d6624aa6057c6/src/neuroconv/tools/roiextractors/roiextractors.py#L1297-L1307) currently introduces an `is_accepted` boolean, but in principle segmentation pipelines could output very different kinds of quality indicators: continuous confidence scores (I think @alessandratrapani has a case like that?), categorical labels (`"good"`, `"bad"`, `"maybe"`), or binary acceptance flags like we do. In summary, I don't see a standard in the segmentation algorithms and I am unsure the current `is_accepted` bool can cover all the bases there. 

Now, @pauladkisson bring the valid concern that if every dataset uses different property names, downstream users lose a uniform way to query accepted vs. rejected ROIs. The situation is that at the moment this “canonical” pathway exists *only* as a NeuroConv convention. The NWB schema itself has no reserved or standardized place for ROI quality. I think that this suggest that maybe we should try to add a canonical column in the schema if that would be possible. @alejoe91 what do you think? you have work extensively with quality metrics and you have been thinking recently about segmentation pipelines, do you think this is feasible? I am looping in @chrishalcrow as well.

If we had that, we could propagate all the properties of the source format and in addition write the canonical column to standardized the data. Meanwhile, my preference is to only write the source properties as they come, maybe this errs too much in the cautious side?

PR #548, opened 2026-01-28
Post-Release 0.7.3

pauladkisson:


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "roiextractors"
-version = "0.7.3"
+version = "0.7.4"
```
h-mayorquin: We removed a lot of things, shouldn't we do a minor bump here?

pauladkisson:
> Mmm now that I think about it, it is the recent release the one that should be 8.0. Maybe we should jank that one and make it with 8.0. I am unsure about this, to be honest, I don't know if it is worth our time.

Yes, it's the release that I already made (0.7.3) that had all of the removals that would probably upgrade it to 0.8.0. Personally, I don't think it's worth the time. It is after all just a number.

PR #564, opened 2026-03-24
 Fix `time_to_sample_indices` and add it to `SegmentationExtractor`

h-mayorquin:
There is a bug in the implementation of `time_to_sample_indices` on `ImagingExtractor`. The method checks `self._times is None` to decide whether to use `searchsorted` or the sampling-frequency formula. Before [PR #465](https://github.com/catalystneuro/roiextractors/pull/465) this worked fine because `_times` was only ever set through an explicit `set_times()` call. When we introduced the lazy `get_native_timestamps` resolution path in PR #465 (and extended it in [PR #529](https://github.com/catalystneuro/roiextractors/pull/529)), `_times` started being populated lazily on the first `get_timestamps()` call. `time_to_sample_indices` was never updated to go through that same resolution path, so for formats that provide native irregular timestamps it silently falls back to the frequency-based formula and returns wrong indices.

This PR fixes it by calling `self.get_timestamps()` first so that native timestamps are resolved and cached before the branch decision. It also aligns the `searchsorted` call with SpikeInterface's `BaseRecordingSegment.time_to_sample_index`, which uses `side="right"` and subtracts 1 to return the sample at or just before the queried time. The method was missing from `SegmentationExtractor` despite that class having the same timestamp infrastructure, so I have added it there for API consistency.

Tests cover three scenarios for each extractor: round-trip with sampling frequency, explicit irregular time vectors via `set_times`, and the core bug case where native timestamps exist but haven't been cached yet.


src/roiextractors/imagingextractor.py
```diff
@@ -323,11 +323,12 @@ def time_to_sample_indices(self, times: float | ArrayLike) -> float | np.ndarray
         sample_indices: float or array-like
             The corresponding sample indices.
         """
-        # Default implementation
-        if self._times is None:
-            return np.round(times * self.get_sampling_frequency()).astype("int64")
+        # Ensure native timestamps are cached if available
+        self.get_timestamps()
```
pauladkisson: Relying on a side effect from `self.get_timestamps` smells like bad design to me and is a little bit opaque. How about this?

```suggestion
        native_timestamps = self.get_native_timestamps()
        if native_timestamps is not None:
            self._times = native_timestamps  # Cache the native timestamps
```

src/roiextractors/segmentationextractor.py
```diff
@@ -838,6 +838,26 @@ def get_timestamps(self, start_sample: int | None = None, end_sample: int | None
...
+        times: float or array-like
+            The times (in seconds) to be converted to sample indices.
+
+        Returns
+        -------
+        sample_indices: np.ndarray
+            The corresponding sample indices.
+        """
+        # Ensure native timestamps are cached if available
+        self.get_timestamps()
```
pauladkisson: Similarly, here. 
```suggestion
        native_timestamps = self.get_native_timestamps()
        if native_timestamps is not None:
            self._times = native_timestamps  # Cache the native timestamps
```

pauladkisson (changes requested):
Small comment about code design/readability; otherwise looks good to me.

h-mayorquin:
> Small comment about code design/readability; otherwise looks good to me.

Done

pauladkisson (approved):
Awesome. Looks great.

PR #566, opened 2026-04-07
Derive sampling frequency from OME-XML `TimeIncrement`

h-mayorquin:
This should come after https://github.com/catalystneuro/roiextractors/pull/571

The OME data model defines a [`TimeIncrement`](https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome.html) attribute on the `Pixels` element that stores the global time interval between timepoints. This is the inverse of sampling frequency. With this change, the `sampling_frequency` parameter in `OMETiffImagingExtractor` becomes optional: when omitted, the extractor computes it as `1.0 / TimeIncrement`, converting units via `TimeIncrementUnit` (which defaults to seconds per the OME spec). If neither `sampling_frequency` nor `TimeIncrement` is available, a `ValueError` is raised.

In practice `TimeIncrement` is rarely populated by vendors (not present in bioformats-artificial, MitoCheck, or tubhiswt sample data), so the explicit `sampling_frequency` parameter remains as the primary way to provide this value. ~~The automatic derivation is most useful for vendor-specific subclasses (like a future Bruker extractor) whose output files do populate this field~~

[EDIT]
Remove Bruker mention as it is not a format that populates `TimeIncrement`:
https://github.com/catalystneuro/roiextractors/pull/567/changes#r3087655176


h-mayorquin:
The dev tests are fixed here:
https://github.com/catalystneuro/neuroconv/pull/1707

tests/test_brukertiffimagingextactor.py
h-mayorquin, line 1: The modifications are here because I stubbed even more some of the files.

pauladkisson (commented):
> In practice TimeIncrement is rarely populated by vendors (not present in bioformats-artificial, MitoCheck, or tubhiswt sample data), so the explicit sampling_frequency parameter remains as the primary way to provide this value. The automatic derivation is most useful for vendor-specific subclasses (like a future Bruker extractor) whose output files do populate this field.

It seems a bit weird to me to have a parameter that is technically optional but in practice required for the majority of cases. I'd like to consider some design alternatives. 

What if sampling frequency remains required, but we add a method to derive the sampling frequency from the metadata or a helper function to do so? Then the vendor-specific subclasses like Bruker could use this method/function and make sampling frequency truly optional in those cases. 

Open to other ideas as well.

h-mayorquin:
> It seems a bit weird to me to have a parameter that is technically optional but in practice required for the majority of cases. I'd like to consider some design alternatives.

Mmm I kinda feel what you mean but is that bad? Can you expand on how are you thinking?

The extractor throws an error when the sampling frequency is not provided when there is not a  `TimeElement`. Does the weirdness come from people reading the API and/or the docs and wrongly inferring that in most cases they don't require sampling frequency?

h-mayorquin:
> What if sampling frequency remains required, but we add a method to derive the sampling frequency from the metadata or a helper function to do so? Then the vendor-specific subclasses like Bruker could use this method/function and make sampling frequency truly optional in those cases.

What I wrote above is wrong for Bruker, I will edit.

I think that most of the vendors that base their format on OME will store either timings or sampling frequency. Bruker, for example stores, timestamps which should be considered the source of truth:
https://github.com/catalystneuro/roiextractors/pull/567/changes#r3087655176

I suspect `TimeElement` will mostly used by people who wrote their own OME but I will take a look at Thor and the other OME based formats to see if this is true. 

This is just info for you to make a decision.

pauladkisson:
> Mmm I kinda feel what you mean but is that bad? Can you expand on how are you thinking?

Basically, I was worried about the situation where a user has to run into a ValueError in practice to discover whether or not they need to input the sampling frequency. Especially if the automatic derivation of the sampling frequency is being added primarily for vendor-specific subclasses of OME TIF. In such cases, it seems to me better to make the sampling frequency explicitly required up front and provide some convenience function to go about getting it.

The math changes a bit if the primary use case is custom OME-TIFFS and/or if `TimeElement` is frequently specified. Then I feel like it becomes a bit more legitimate by default, extracting the sampling frequency from the metadata, and only occasionally falling back to manually specify it.

h-mayorquin:
Ok, thanks for making me look deper into this. Here is the state of affairs. 

Most of the OME-tif based readers taht we could impelment with this as a base have their own logic so they would not benefit from this:

* Bruker Prairie View: derives sampling frequency from `Frame/@relativeTime` in the Bruker configuration XML
* Micro-Manager: derives from `PlaybackFPS` in `DisplaySettings.json`
* ThorImageLS: derives from `LSM/@frameRate` in `Experiment.xml`

So we don't have anything in the inheritance-path that would make use of this. Let's assume is very uncommon or null.

The other relevant case is Ome-Tiff directly and we have two cases here:
*  if an experimenter dediced to save their own data in the format reader and use `TimeIntervals` as indicated by the spec. 
* IF someone uses BIO-formats to convert their data to ome-tiff standard and then they want to conver those files to NWB. 

In the latter case there is many formats that acutally propagate this info. I cloned the repo of the Java [Bio-Formats](https://github.com/ome/bioformats) and had claude fetch which formats write to `TimeIntervals` and here is the table


| Reader | Format | Source of TimeIncrement |
|---|---|---|
| ZeissCZIReader | Zeiss CZI | `Interval/Increment` |
| ZeissLSMReader | Zeiss LSM | `track.timeIncrement` |
| LeicaReader | Leica LEI | `physicalSizes` array |
| LeicaHandler | Leica LIF (XML) | `nDelayTime_s` |
| LIFReader | Leica LIF | `tSteps` |
| PrairieReader | Bruker Prairie View | `waitTime` |
| FluoviewReader | Olympus Fluoview | `voxelT` |
| FV1000Reader | Olympus FV1000 | `pixelSizeT` |
| ICSReader | ICS/IDS | time axis scale |
| NiftiReader | NIfTI | `deltaT` |
| CellVoyagerReader | Yokogawa CellVoyager | `readFrameInterval()` |
| FEITiffReader | FEI TIFF | `timeIncrement` |
| ImprovisionTiffReader | Improvision TIFF | `pixelSizeT` |
| IPLabReader | IPLab | `timeIncrement` |
| IvisionReader | iVision | `increment` |
| ImarisReader | Imaris | hardcoded `1.0` |
| TiffReader | Generic TIFF | `timeIncrement` (if present) |


Given how many formats propagate `TimeIncrement` through Bio-Formats, I think the automatic derivation is justified. The Bio-Formats conversion path is common enough that users will encounter OME-TIFF files with `TimeIncrement` populated without knowing the sampling frequency offhand. Keeping `sampling_frequency` optional with a clear `ValueError` when neither source is available seems like the right trade-off to me. 

What do you think?

pauladkisson:
> Given how many formats propagate `TimeIncrement` through Bio-Formats, I think the automatic derivation is justified. The Bio-Formats conversion path is common enough that users will encounter OME-TIFF files with `TimeIncrement` populated without knowing the sampling frequency offhand. Keeping `sampling_frequency` optional with a clear `ValueError` when neither source is available seems like the right trade-off to me.
> 
> What do you think?

Yep, sounds good to me. Thank you for taking the time to properly justify this choice. I think it will prove helpful if we ever need to revisit.

pauladkisson (approved):

PR #567, opened 2026-04-07
Add unified BrukerTiffImagingExtractor

h-mayorquin:
Close https://github.com/catalystneuro/roiextractors/issues/565

This PR depends on #565 (OMETiffImagingExtractor) and should be merged after it.

`BrukerTiffImagingExtractor` replaces the existing `BrukerTiffSinglePlaneImagingExtractor` and `BrukerTiffMultiPlaneImagingExtractor` with a single class that inherits from `OMETiffImagingExtractor`. Structural metadata (dimensions, channels, planes, file layout) comes from the OME-XML embedded in the TIFF files. Timing comes from the Bruker configuration XML's `Frame/@relativeTime` attributes, which are the ground truth for timestamps. The sampling frequency is derived from these timestamps via `calculate_regular_series_rate`. For volumetric data, timestamps are subsampled by `num_planes` to give one per volume.

The old classes are deprecated with `FutureWarning` (October 2026) but remain functional. The new class is self-contained so the old code can be dropped cleanly later.


src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -122,6 +125,140 @@ def _parse_xml(folder_path: PathType) -> etree.Element:
...
+                "https://github.com/catalystneuro/roiextractors/issues \n"
+                "We welcome sample data for improving our test coverage and ensuring correctness.",
+                stacklevel=2,
+            )
+
+        # Pre-set _num_planes so get_native_timestamps() can be called before super().__init__().
+        # MultiTIFFMultiPageExtractor.__init__() will set it again to the same value.
+        self._num_planes = num_planes
+        timestamps = self.get_native_timestamps()
+        sampling_frequency = calculate_regular_series_rate(timestamps)
```
h-mayorquin, line 189: @pauladkisson Bruker has its own way of calculating the sampling frequency as the acquisition system has native timestamps.

pauladkisson (approved):
Looks good to me!

PR #568, opened 2026-04-07
 Remove typing indirection in extraction_tools

h-mayorquin:
A requested by @pauladkisson comment on #563 noted that the `ArrayType`, `DtypeType`, `FloatType`, and `NumpyArray` aliases in `extraction_tools` were originally hand-rolled unions predating `numpy.typing`, then swapped to `ArrayLike`/`DTypeLike` in #131 but kept under the old names, adding indirection without benefit. This PR removes all four aliases and replaces their usages with direct imports of `ArrayLike` and `DTypeLike` from `numpy.typing`, built-in `float`, and `np.ndarray`.


pauladkisson (approved):
Looks good to me!

PR #571, opened 2026-04-14
Add `OMETiffImagingExtractor`

h-mayorquin:
With the introduction of `MultiTIFFMultiPageExtractor` ([PR #402](https://github.com/catalystneuro/roiextractors/pull/402)), we now have a generalized mechanism for reading multi-file, multi-page TIFF data with flexible dimension ordering. [OME-TIFF](https://docs.openmicroscopy.org/ome-model/5.6.3/ome-tiff/specification.html) is a widely used format that follows the same multi-page, multi-file structure but adds an XML metadata layer conforming to the [OME data model schema](https://www.openmicroscopy.org/Schemas/). This metadata encodes dimension order, channel count, depth planes, and file references, meaning all the parameters that `MultiTIFFMultiPageExtractor` requires can be parsed automatically. `OMETiffImagingExtractor` does exactly this: it reads the OME-XML, extracts the necessary parameters, and delegates all data access to `MultiTIFFMultiPageExtractor`. The user only needs to provide a file path and sampling frequency.

I have introduced this extractor primarily to lay the groundwork for improving the Bruker readers ([#565](https://github.com/catalystneuro/roiextractors/issues/565)). Since Bruker's output files are OME-TIFF, a future Bruker extractor can inherit from `OMETiffImagingExtractor` and only add Bruker-specific logic (computing sampling frequency from the Bruker configuration XML). Beyond Bruker, OME-TIFF is common enough across microscopy tools that supporting it directly is valuable on its own.

I have left out sampling frequency derivation from the OME-XML `TimeIncrement` attribute to keep this PR focused on the structural reading. That will come in [PR #566](https://github.com/catalystneuro/roiextractors/pull/566).


pauladkisson (approved):
Looks good to me!

PR #572, opened 2026-04-15
Deprecate MiniscopeMultiRecordingImagingExtractor 

h-mayorquin:
`MiniscopeMultiRecordingImagingExtractor` was built to handle the Tye Lab folder layout, where multiple recordings live in timestamp subfolders under a single session directory. We no longer need it because `MiniscopeImagingExtractor` accepts explicit `file_paths` directly, which is better in two ways: it doesn't silently concatenate recordings under the assumption they are back-to-back (which catalystneuro/neuroconv#1536 notes "is not generally true"), and it leaves folder discovery and timestamp alignment to the caller rather than baking in a single lab's conventions. In neuroconv, the modern `MiniscopeConverter` config-file mode already creates separate `MiniscopeImagingInterface` instances per recording with external timestamp alignment, and the old legacy path has been deprecated in https://github.com/catalystneuro/neuroconv/pull/1706 with a December 2026 removal date.

I have left the class functional with a `FutureWarning` on instantiation so existing code doesn't break before the December 2026 deadline. After that we should remove the class entirely so we don't maintain unnecessary code.


alessandratrapani (approved):
LGTM

PR #574, opened 2026-04-16
Add channel name support for Bruker multi-channel data

h-mayorquin:
Bruker OME-TIFF files have real channel names in their OME-XML (`Channel/@Name="Ch1"`, `Channel/@Name="Ch2"`). This PR adds support to `get_available_channel_names()` which is the pattern that we follow so people can query the channel names and initialize multi-channel source data with an extractor (examples: [1](https://github.com/catalystneuro/roiextractors/blob/473b17ec378f6f06bca8dbf0395af2e5dbc533fc/src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py#L700), [2](https://github.com/catalystneuro/roiextractors/blob/473b17ec378f6f06bca8dbf0395af2e5dbc533fc/src/roiextractors/extractors/suite2p/suite2psegmentationextractor.py#L29), [3](https://github.com/catalystneuro/roiextractors/blob/473b17ec378f6f06bca8dbf0395af2e5dbc533fc/src/roiextractors/extractors/femtonicsimagingextractor/femtonicsimagingextractor.py#L524)).

Adding this required modifications at two levels. In `MultiTIFFMultiPageExtractor`, the channel resolution logic (previously a hardcoded `int(channel_name)`) was replaced with a `_get_channel_names()` method that returns the list of valid names. The base class returns numeric strings ("0", "1", ...) and resolves `channel_name` via `.index()`. Subclasses override `_get_channel_names()` to return format-specific names. In `OMETiffImagingExtractor`, `_parse_ome_metadata` now extracts `Channel/@Name` attributes from the OME-XML, and `_get_channel_names()` returns those names when present.

As a side benefit, this adds general `Channel/@Name` support to `OMETiffImagingExtractor` for any OME-TIFF file with named channels. Our standard OME-TIFF test stubs (bioformats-artificial, MitoCheck, tubhiswt) don't have `Channel/@Name`, so that path is tested via the Bruker multi-channel stub only. A static `get_available_channel_names(file_path)` method lets users discover channel names before constructing an extractor.


pauladkisson (approved):
Looks good to me!

PR #579, opened 2026-05-07
Refactor ThorTiffImagingExtractor to inherit from OMETiffImagingExtractor

h-mayorquin:
`ThorTiffImagingExtractor` now inherits from `OMETiffImagingExtractor`, so OME-XML parsing, multi-file discovery, and page iteration are shared with the rest of the OME-TIFF extractors. 

The subclass keeps only the ThorImage-specific work: reading the sidecar `Experiment.xml` for sampling frequency (`LSM/@frameRate`) and channel names (`Wavelengths/Wavelength/@name`), and overriding `_get_channel_names()` so the user-facing channel selector matches what ThorImage records. 

This PR looks large but is just droping all the code that is unecessary, fixing one bug in the ome sample frequency and modifying a test that was relying on a private attribute of the class.


pauladkisson (approved):
Looks good.

PR #580, opened 2026-05-07
Remove get_channel_names

h-mayorquin:
Time has come.


src/roiextractors/multisegmentationextractor.py
```diff
@@ -100,7 +100,8 @@ def __init__(self, segmentatation_extractors_list, plane_names=None):  # TODO: H
         self._sampling_frequency = self._segmentations[0].get_sampling_frequency()
         self._raw_movie_file_location = self._segmentations[0]._raw_movie_file_location
         self._channel_names = []
-        _ = [self._channel_names.extend(self._segmentations[i].get_channel_names()) for i in range(self._no_planes)]
```
pauladkisson, line 103: What's going on here?

pauladkisson: This code still depends on the concept of a channel names list for each segmentation extractor. I thought the point of this PR was to remove such said concepts since each extractor is single-channel.

pauladkisson (changes requested):
I think maybe there's a little more work to do here, removing the `_channel_names` attribute from segmentation extractors in addition to the public-facing method.

src/roiextractors/multisegmentationextractor.py
h-mayorquin: Good catch. This was an overly conservative change the LLM side.

Looking at it again, nothing reads `self._channel_names` on `MultiSegmentationExtractor`. `get_num_channels` is overridden directly on the multi class and sums across children, the base class's `get_num_channels` is already deprecated, and there are no external consumers of this class. More importantly though, see #588, I think we should deprecate and then remove `MultiSegmentationExtractor`.

I removed the lines.

pauladkisson (approved):
Looks good!

PR #584, opened 2026-05-08
Add `select_plane` to `ImagingExtractor`

h-mayorquin:
Add a method on the base `ImagingExtractor` that returns a lazy planar view of a single depth plane of a volumetric extractor. Naming matches `select_rois` (PR #547). The returned extractor is planaer (`is_volumetric=False`).


pauladkisson (approved):
Looks good

PR #589, opened 2026-05-08
Make `BrukerTiffImagingExtractor` read structural metadata from the Bruker XML

h-mayorquin:
Initially in PR #567 I made it inherit from `OMETiffImagingExtractor` because in the form it is, it is actually an ome tiff: it has the ome xml metadata embedded in the tiff and reading the structure from there works for most of our data. However, when profiling this in PR #592 I saw that it fails for some recordings. The reason is that Prairie View 5.7+ writes the ome xml in `<BinaryOnly>` form: the metadata embedded in the `.ome.tif` is only a stub and the real `<Pixels>` block (dimensions, channels, planes, file layout) is moved into a separate `.companion.ome` sidecar (#595). Our reader only looks at the embedded metadata and never at the companion, so on these files it finds no `<Pixels>` and raises `No Pixels element found in OME-XML metadata.

We could fix this by teaching our reader to follow the companion sidecar (issue #595). But even if we did that, I do not think relying on the ome metadata is the right call. Digging further in the [java prairie reader](https://github.com/ome/bioformats) (Bio-Formats's [`PrairieReader`](https://github.com/ome/bioformats/blob/4fe6b0d71ef4b201824ed2d5d90117e929420dfd/components/formats-gpl/src/loci/formats/in/PrairieReader.java)) I realize that the problem is bigger as bruker is not that good at respecting ome-tiff metadata (its ome xml sometimes reports the wrong dimension order, `XYZCT` when the bytes are `XYCZT`, and uses generic `Ch1`/`Ch2` names instead of the user's labels), so even with the companion resolved we would keep reading bad structural data. I confirmed the channel-name case directly: in one of our dual-channel recordings the ome xml labels the channels `Ch1`/`Ch2` while the bruker xml carries the user's actual `Red`/`Green` labels, so the user-facing channel names only survive if we read the bruker xml. As an example, `PrairieReader` itself ignores the ome metadata and hardcodes `dimensionOrder = "XYCZT"` [PrairieReader.java#L470](https://github.com/ome/bioformats/blob/4fe6b0d71ef4b201824ed2d5d90117e929420dfd/components/formats-gpl/src/loci/formats/in/PrairieReader.java#L470). In addition, the companion is a separate file that can go missing anyway. In fact, previous datasets of Bruker that we received lacked the companion file.

I think we should do the same, this PR fixes this by making bruker inheriting from the multi-tiff extractor bypassing the ome layer and using the Bruker xml as the ground truth for determining the structural parameters (num_planes, channel_names, etc). So, in summary, the authority for those parameters becomes the bruker xml instead of the ome one.


pauladkisson (approved):
Reasoning makes sense and code looks good to me. Sorry for the delay!

PR #592, opened 2026-05-19
Stop holding TIFF file handles in `MultiTIFFMultiPageExtractor` instance state

h-mayorquin:
The same Bruker recordings that exposed the `OverflowError` in PR #591 also exposed a second problem: `MultiTIFFMultiPageExtractor.__init__` opened every TIFF file in the dataset and held every handle for the lifetime of the extractor. On recordings with tens or hundreds of thousands of files, this exhausted the per-process file-descriptor limit on typical Linux systems (default `ulimit -n` of 1024 or 4096).

Construction now opens each file briefly to read its IFD count and closes it again; only the first file is read further to learn shape and dtype. `get_series` opens what it needs locally, closes the handles explicitly before returning, and never stores them on the instance. This is the same pattern we used before for Bruker for similar reasons ([deprecated `_BrukerTiffSinglePlaneImagingExtractor.get_series`](https://github.com/catalystneuro/roiextractors/blob/8901366acdf23ba9225510687f4df41c4ed33973/src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py#L733-L748)). 

Incidentally, now that no file descriptors persist between calls the extractor becomes picklable, which enables multi-processing workflows.


src/roiextractors/extractors/tiffimagingextractors/multitiffmultipageextractor.py
```diff
@@ -381,19 +378,32 @@ def get_series(self, start_sample: int | None = None, end_sample: int | None = N
...
-
-                tiff_handle = self._file_handles[file_index]
-                image_file_directory = tiff_handle.pages[ifd_index]
+                tiff_handle = opened_handles[table_row["file_index"]]
+                image_file_directory = tiff_handle.pages[table_row["IFD_index"]]
                 samples[return_index, :, :, depth_position] = image_file_directory.asarray()
 
+        # Explicit close: tifffile holds internal references so the refcount in the garbage collector
+        # won't promptly free these handles when the local dict goes out of scope.
+        for handle in opened_handles.values():
```
pauladkisson: This would still break if the user calls get_series(:). Maybe we should open the files one at a time?

h-mayorquin: I fixed this, it makes it slower ~30 % slower but let's have this functional and then we will improve its performance later.

pauladkisson (approved):
Looks good to me.

PR #593, opened 2026-05-19
Skip per-file TIFF opens at init for Bruker via a `_get_ifds_per_file()` override

h-mayorquin:
`MultiTIFFMultiPageExtractor` needs the page count of each file to compute `num_samples` and build the frame-to-page map in [`_create_frame_to_ifd_table`](https://github.com/catalystneuro/roiextractors/blob/85db418840fe90a102feab556059a3f0e293be43/src/roiextractors/extractors/tiffimagingextractors/multitiffmultipageextractor.py#L258). The default gets those counts by opening every TIFF, which is cheap for the handful of multi-page files most formats produce but costly for Bruker, which writes one single-page TIFF per frame: a recording is tens to hundreds of thousands of files, and the open-everything loop dominates construction and can re-trigger the file-descriptor exhaustion handled in #592.

This PR adds an overridable `_get_ifds_per_file()` hook to `MultiTIFFMultiPageExtractor` (the default still opens each file) and has `BrukerTiffImagingExtractor` override it to read counts from the Bruker configuration XML's `<File page="..."/>` attribute instead, opening zero images. The base still opens one file to probe shape and dtype, so construction now opens a single TIFF regardless of file count. 

The only spiky point is the testing. The way this fails is because the file system can't support opening that many files and that is not easy to test in a file. The test that I added patches the method that opens tiff files and in a block asserts that only one file is open. I think this does the work but something about it does not convince me fully. I think it should be fine to go until we come up with something better (if necessary) or if you have better suggestions I am all ears.


tests/test_brukertiffimagingextactor.py
```diff
@@ -307,6 +307,33 @@ def test_timestamps_and_sampling_frequency(self):
...
+        forward each call to the real opener and return its real result, so the shape/dtype probe
+        still gets a genuine file handle while the spy records the call. ``call_count`` is then the
+        number of files opened during construction, which must be exactly one.
+
+        This stub has 10 files; without the override, init would open 11 (one probe plus one per
+        file to count pages). Pinning the count to one catches a regression back to per-file opens.
+        """
+        from unittest.mock import patch
+
+        with patch("tifffile.TiffFile", wraps=tifffile.TiffFile) as mock_tiff_file:
```
pauladkisson, line 332: Normally I'm pretty skeptical about excessive patching, but in this case I think it's fine.

pauladkisson (approved):

tests/test_brukertiffimagingextactor.py
h-mayorquin: Yes, I don't like it either but I could not come up with something better as you can see by the long comment over there. Thanks for taking a look.

Issue #526, opened 2025-11-12
[Bug]: ValueError: Unknown series type: TSeries Brightness Over Time Element


julia-dziubek:
### What happened?

Hello,
I'm converting single-plane 2P data acquired with Bruker to NWB. 
My code breaks when creating the converter instance with `BrukerTiffSinglePlaneConverter` (see traceback)

I'm attaching the folder's XML file.
[TSeries-08162024-1854-001.xml](https://github.com/user-attachments/files/23505071/TSeries-08162024-1854-001.xml)
 

### Steps to Reproduce

```python
Here's my code: 

from pathlib import Path
from zoneinfo import ZoneInfo
from neuroconv.converters import BrukerTiffSinglePlaneConverter

# Define the data path
OPHYS_DATA_PATH = Path("D:/2P_data")
folder_path = OPHYS_DATA_PATH / <my data> / "TSeries-08162024-1854-001"

# Create the converter instance
converter = BrukerTiffSinglePlaneConverter(folder_path=folder_path)

I'm using an  environment with: 
Python 3.11.14
roiextractors 0.7.2
neuroconv 0.8.3
```

### Traceback

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[3], line 2
      1 #%% Create the converter instance
----> 2 converter = BrukerTiffSinglePlaneConverter(folder_path=folder_path)

File c:\Users\Julia\anaconda3\envs\nwb\Lib\site-packages\neuroconv\datainterfaces\ophys\brukertiff\brukertiffconverter.py:255, in BrukerTiffSinglePlaneConverter.__init__(self, folder_path, verbose)
    241 """
    242 Initializes the data interfaces for Bruker imaging data stream.
    243 
   (...)    249     Controls verbosity.
    250 """
    251 from roiextractors.extractors.tiffimagingextractors.brukertiffimagingextractor import (
    252     _determine_imaging_is_volumetric,
    253 )
--> 255 if _determine_imaging_is_volumetric(folder_path=folder_path):
    256     raise ValueError("For volumetric imaging data use BrukerTiffMultiPlaneConverter.")
    258 self.verbose = verbose

File c:\Users\Julia\anaconda3\envs\nwb\Lib\site-packages\roiextractors\extractors\tiffimagingextractors\brukertiffimagingextractor.py:106, in _determine_imaging_is_volumetric(folder_path)
    104             break
    105         else:
--> 106             raise ValueError(
[...]

PR #596, opened 2026-06-14
Handle multi-burst Bruker recordings that reset relativeTime per sequence

h-mayorquin:
Most Bruker recordings are sampled at a constant rate, but some are not. Acquisition modes like Brightness Over Time and multi-cycle timed series record in separate bursts with real, multi-second gaps between them, so no single sampling frequency can describe the recording. The extractor assumed every recording was uniformly sampled, and when it could not derive one rate for these it raised "Could not determine sampling frequency" instead of representing the genuine gapped timing. The mechanism is shown below: Prairie View resets the per-frame `relativeTime` to 0 at the start of each burst, so the old code's concatenate-and-find-one-rate approach saw a sawtooth and gave up.

```
stored relativeTime  (resets to 0 each burst — the sawtooth the old code choked on)

  burst 1               burst 2               burst 3
  0─0.07─…─0.27         0─0.07─…─0.27         0─0.07─…─0.27
  └─ concat across bursts → calculate_regular_series_rate() = None → raise


reconstructed timeline  (each burst offset by its absoluteTime start → monotonic, gapped)

  burst 1            gap ~10.3 s          burst 2            gap ~10.1 s         burst 3
  0──…──0.27  ───────────────────────►  10.6──…──10.9  ──────────────────────►  21.0──…──21.3
  frames 0–4                            frames 5–9                              frames 10–14

  get_timestamps() returns this vector, has_time_vector() = True
  sampling_frequency = within-burst rate (~15 Hz), reported as nominal
  warns: recording is not uniformly sampled
```

So in these cases the extractor becomes timestamps-first, like `MiniscopeImagingExtractor`: the real per-frame timestamps are the source of truth and the rate is only a derived summary, rather than the rate being primary and the timestamps reconstructed as `start + i / rate`. It rebuilds the true timeline by offsetting each burst by its `absoluteTime` start, exposes it through `get_timestamps()` (`has_time_vector()` is `True`), reports the within-burst rate as the nominal `sampling_frequency`, and warns that sampling isn't uniform. Single-sequence and volumetric recordings, whose timing is continuous, are unaffected. I also added a small multi-burst Brightness Over Time fixture (stubbed from data Julia shared on #526) whose test asserts both that `get_timestamps()` preserves the real inter-burst gaps and that `sampling_frequency` is the within-burst rate.


src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
```diff
@@ -190,10 +191,21 @@ def __init__(self, folder_path: PathType, channel_name: str | None = None):
...
-        sampling_frequency = calculate_regular_series_rate(timestamps)
+        sampling_frequency, num_segments = calculate_segmented_series_rate(timestamps)
         if sampling_frequency is None:
             raise ValueError("Could not determine sampling frequency from Bruker configuration XML.")
+        if num_segments > 1:
+            warnings.warn(
+                "This Bruker recording has multiple <Sequence> blocks (bursts/cycles) and is not "
+                "uniformly sampled. Reporting the within-burst frame rate as sampling_frequency; "
+                "use get_timestamps() for the true (gapped) per-frame timeline.",
+                stacklevel=2,
```
pauladkisson: Do we really need a warning here? It seems like a perfectly legitimate use case that our extractor should have handled automatically all along.

tests/test_brukertiffimagingextactor.py
```diff
@@ -463,6 +463,65 @@ def test_construct_and_select_channels(self):
...
+        assert extractor.get_sampling_frequency() == pytest.approx(15.02, rel=1e-3)
+
+        # true timeline: regular within each burst, real inter-burst gaps preserved
+        assert extractor.has_time_vector()
+        timestamps = extractor.get_timestamps()
+        assert len(timestamps) == 15
+        assert timestamps[0] == pytest.approx(0.0, abs=1e-6)
+        assert timestamps[1] - timestamps[0] == pytest.approx(0.0666, rel=1e-2)  # within burst
+        assert timestamps[5] - timestamps[4] == pytest.approx(10.34, rel=1e-2)  # burst 1 -> 2
+        assert timestamps[10] - timestamps[9] == pytest.approx(10.13, rel=1e-2)  # burst 2 -> 3
```
pauladkisson: With only 15 elements, we should be able to just assert on the full timestamps array, which is simpler and more complete.

tests/test_brukertiffimagingextactor.py
```diff
@@ -463,6 +463,65 @@ def test_construct_and_select_channels(self):
...
+        assert extractor.get_sampling_frequency() == pytest.approx(15.11, rel=1e-3)
+
+        assert extractor.has_time_vector()
+        timestamps = extractor.get_timestamps()
+        assert len(timestamps) == 9
+        assert timestamps[0] == pytest.approx(0.0, abs=1e-6)
+        assert timestamps[1] - timestamps[0] == pytest.approx(0.0662, rel=1e-2)  # within cycle
+        # the cycles are irregularly spaced in this recording (real, not uniform)
+        assert timestamps[3] - timestamps[2] == pytest.approx(25.72, rel=1e-2)  # cycle 0 -> 1
+        assert timestamps[6] - timestamps[5] == pytest.approx(7.70, rel=1e-2)  # cycle 1 -> 2
```
pauladkisson: With only 9 elements, we should be able to just assert on the full timestamps array, which is simpler and more complete.

pauladkisson (commented):
Looks good, small comments

tests/test_brukertiffimagingextactor.py
h-mayorquin: I agree. Thanks for taking a look to that.

tests/test_brukertiffimagingextactor.py
h-mayorquin: Changed.

src/roiextractors/extractors/tiffimagingextractors/brukertiffimagingextractor.py
h-mayorquin: I agree this could be better, and the warning isn't really actionable as it stands and I think it is more intrusive than I would like.

The problem is that our whole model assumes a constant sampling rate, so we need to inform the user that this is not the case here. A single sampling rate can't cover the gaps and the user should know this. I thought about alternatives like setting the rate to np.nan, but that's too much: at least in the samples we have, the within-burst segments are constant and the sampling rate is just the mean of those, so it is informative. On the other hand, putting the warning on the sampling frequency getter would be even more intrusive.

I think that eventually maybe we could have a way of indicating this, either with the SpikeInterface segment concept (which I don't like a lot) or with just a vector of offsets, but at the moment I am slightly in favor of the status quo.

Do you have a better idea? Or maybe you disagree that the user should know? let me know what you think.

pauladkisson: > The problem is that our whole model assumes a constant sampling rate, so we need to inform the user that this is not the case here.

Isn't that the point of `has_time_vector`? To indicate to users that they should be looking for a timestamps array instead of a simple sampling rate?

pauladkisson: > Do you have a better idea? Or maybe you disagree that the user should know? let me know what you think.

I guess I'm thinking that the existence of a sampling_rate field shouldn't (and doesn't?) indicate that the timestamps are perfectly uniform, and we shouldn't have to warn users about such a common situation.

h-mayorquin: I have always read it the other way: a reader should be able to assume the sampling rate is constant, and `has_time_vector()` was an internal handle we used for automated readers (the NWB conversion, for example) rather than something a human reader is expected to check. As an example, the repr always prints the sampling rate, both in SpikeInterface and here, which to me signals that homogeneous timestamps are the happy path and timestamps are sort of an oddity (I am pretty sure that's how Sam thinks about it on the SpikeInterface side).

That said, the fact that you interpreted it differently is evidence that I should not come too strong on this. I will remove the warning. Let's think about how to signal non-homogenous timestamps better.  I will raise it with the SpikeInterface team so maybe we can unify the convention on both sides.

pauladkisson: >  I will raise it with the SpikeInterface team so maybe we can unify the convention on both sides.

Sweet, would be good to get some consensus on this.

pauladkisson (approved):

PR #599, opened 2026-06-24
Memory-efficient `slice_field_of_view` for `ScanImageImagingExtractor`

h-mayorquin:
Overrides `slice_field_of_view` on `ScanImageImagingExtractor` so a cropped field of view reads only the requested region instead of materializing the full page, giving the multi-ROI mesoscope extractor in #597 a shared, memory-efficient cropping path to build on rather than rolling its own. This will benefit all the users of the scan image and not only the #597


src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
```diff
@@ -600,14 +612,86 @@ def get_series(self, start_sample: int | None = None, end_sample: int | None = N
...
+        absolute_column_start = self._column_slice.start + column_start
+
+        sliced = copy.copy(self)
+        sliced._row_slice = slice(absolute_row_start, absolute_row_start + (row_end - row_start))
+        sliced._column_slice = slice(absolute_column_start, absolute_column_start + (column_end - column_start))
+        sliced._num_rows = row_end - row_start
+        sliced._num_columns = column_end - column_start
+        # Keep the handle owner alive and mark this object as a borrower so __del__ does not close the
+        # shared file handles. The owner is always the originally constructed extractor.
+        sliced._fov_parent = self._fov_parent if self._fov_parent is not None else self
```
pauladkisson: Is there a reason why we want _fov_parent to always point to the original ancestor instead of the direct parent?

pauladkisson (approved):
Looks good to me. Just one curious question.

src/roiextractors/extractors/tiffimagingextractors/scanimagetiffimagingextractor.py
h-mayorquin: Ah, I actually did not realize that before when I reviewed the agent work. 

I am looking at it and I think that is so the `__del__` works well. The root is the only extractor that actually owns the shared file handles (every view is a shallow copy sharing the same readers), so it's the only object that needs to stay alive and the only one whose `__del__` closes them. A direct-parent chain would also keep the handles open, but it would needlessly pin every intermediate view alive instead of just the owner. 

I just renamed the variable `_fov_parent` -> `_fov_root` since the old name implied "direct parent" and was misleading.

